PR #1, opened 2019-04-24
Task

ivadimko:


config/backstop/backstopConfig.js
```diff
@@ -4,7 +4,7 @@
 const basicScenario = {
   label: 'test', // name of the test
   url: 'http://localhost:8080',
-  referenceUrl: '', // put here reference to github-pages with ready project
+  referenceUrl: 'https://mate-academy.github.io/layout_solutions/header/',
```
roman-mate: header is too generic. I'd put into under moyo_header. So that we can have more 'headers' in future.

readme.md
```diff
@@ -1,6 +1,32 @@
-# Boilerplate for layout tasks
+# Frontend practice with header
 
-Here is the description.
+## Main goal:
+Create HTML page with header. Understand `flexbox` usage. Create
+HTML layout from the mockup.
+
+Use this [mockup](#) for development.
```
roman-mate: Do we have the mock? it may be different from out reference implementation.

ivadimko: @roman-mate I'm waiting for mock from Alesya. She has done a general mockup and now we wait for splitting it to different tasks

roman-mate: asesome.

PR #2, opened 2019-04-26
Vitalii Pshychenko

VitaliiPshychenko:


src/style.css
```diff
@@ -0,0 +1,60 @@
+* {
+  font-family: "Roboto", sans-serif;
+  color: #000;
+  background-color: #fff;
+  margin: 0;
```
ivadimko: Suggestion: styles like `color`, `font-family` can be inherited. You can apply it directly to the `body`.
We need `margin: 0` and `box-sizing: border-box` with `*` to apply it for all elements at one time and to not repeat it every time.
`background-color` can also be applied to the `body`
So, the golden version will be: 
```css
* {
  margin: 0;
  box-sizing: border-box;
}
body {
  color: #000;
  background-color: #fff;
}
```

src/style.css
```diff
@@ -0,0 +1,60 @@
...
+
+.nav-item:last-child {
+  margin-right: 0;
+}
+
+.nav-item a:hover {
+  color: #00acdc;
+}
+
+.is-active {
```
ivadimko: Suggestion: in big projects, you will have many items which can be `active`. Better to define which element exatly you want to style with class `is-active`, because this class is like `utility class`.
```css
.nav-item a.is-active {...}
```

src/style.css
```diff
@@ -0,0 +1,60 @@
...
+  color: #00acdc;
+  position: relative;
+}
+
+.is-active:after {
```
ivadimko: the same
```css
.nav-item a.is-acitve::after {...}
```

src/index.html
```diff
@@ -11,9 +11,42 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Document</title>
   <link rel="stylesheet" href="./style.css">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet">
 </head>
 <body>
-
+  <header class="main-header">
+    <h1>
```
ivadimko: This is not an `h1`. Usually, the logo is a link to the homepage. So better to write:
```html
<a href="#" title="Go Home">
  <img ... >
</a>
```

PR #29, opened 2019-05-03
 rostyslav

RMasnyuk:


src/index.html
```diff
@@ -11,9 +11,22 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
```
gestap: You forgot about links to gh-page and test.

PR #30, opened 2019-05-03
Solution

KonstVolkovnitsky:


src/index.html
```diff
@@ -9,11 +9,26 @@
           maximum-scale=1.0,
           minimum-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link href = "https://fonts.googleapis.com/css?family=Roboto" rel = "stylesheet">
```
gestap: you need font with Cyrillic letter.

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  height: 60px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+nav {
+  display: flex;
+  align-items: center;
+}
+
+nav a {
```
gestap: better to add class to `a`

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+  position: relative;
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+}
+
+a:hover,
```
gestap: You add `a` class name `menu-item`. Why you write `a` here?

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+nav .is-active {
+  color: #00acdc;
+}
+
+nav .is-active:after {
```
gestap: no need to write `nav` here

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+}
+
+a:hover,
+nav .is-active {
```
gestap: also here

PR #37, opened 2019-05-09
Solution by Vlad Polets

vpolets:


src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  margin-right: 20px;
+}
+
+.nav__link_hovered:hover,
+.is-active {
+  color: var(--link-color);
+  position: relative;
+}
+
+.nav__link_active:after {
```
gestap: class `is-active` must have blue line.

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  content: "";
+  width: 37px;
```
gestap: you must use `width:100%`.

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  height: 4px;
+  background: var(--link-color);
+  position: absolute;
+  box-sizing: border-box;
+  left: 0;
+  bottom: 0;
+  border: 2px solid var(--link-color);
```
gestap: Use `border` or `height`, they do the same in this situation.

src/style.css
vpolets: @gestap 'link with blue color and line below is an active link. It should have class="is-active" and relevant styles'. Link has this class, and relevant class for this line. Or what do you mean when write 'must'?

vpolets: @gestap I fixed it. Now class `is-active` is only for tests

gestap: 1. class `is-active` must be in the project
2. elements with class `is-active` must have `color: #00acdc;` and underline.

if we want another item in the navigation was blue and with an underline, we need only add to him class `is-active`.

If you don't understand what I mean, write to me in slack.

PR #39, opened 2019-05-13
header

s0lm1r:


src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  display: flex;
+  justify-content: space-between;
+  flex-wrap: nowrap;
+  align-items: center;
+  height: 60px;
+}
+
+.header__nav {
+  align-items: center;
+  margin-top: -3px;
```
gestap: why you use -3px?

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  margin-left: 16px;
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  color: #000;
+  padding: 23px 0;
+  height: 15px;
```
gestap: why you need this?

PR #40, opened 2019-05-13
Vladimir's Solution

VladimirTi:
Tests doesn`t work (reference images didn`t download)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,46 @@
+@import url("https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&subset=cyrillic");
+
+body {
+  margin: 0;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 40px;
+  padding: 10px 50px;
+}
+
+.navigation {
+  display: flex;
+  justify-content: space-between;
+
+}
+
+.navigation > a {
+  position: relative;
+  margin-left: 20px;
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+}
+
+.navigation > a:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active:after {
+  position: absolute;
+  content: "";
+  width: 37px;
+  height: 4px;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  justify-content: space-between;
+  align-items: center;
+  height: 40px;
+  padding: 10px 50px;
+}
+
+.navigation {
+  display: flex;
+  justify-content: space-between;
+
```
YuriiHoliuk: remove

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+}
+
+.navigation > a:hover,
+.is-active {
```
YuriiHoliuk: is good to have simple selectors, but in the real world, you will have many is-active classes, better to use more specific as above

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  color: #00acdc;
+}
+
+.is-active:after {
+  position: absolute;
+  content: "";
+  width: 37px;
+  height: 4px;
```
YuriiHoliuk: ^

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  line-height: 14px;
+  text-transform: uppercase;
+}
+
+.navigation > a:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active:after {
```
YuriiHoliuk: better to do it with border, it will be simpler solution

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+
+.navigation {
+  display: flex;
+  justify-content: space-between;
+
+}
+
+.navigation > a {
+  position: relative;
+  margin-left: 20px;
```
YuriiHoliuk: Are you sure you need left margin in the first element?

src/style.css
VladimirTi: i think it impossible to do on this way of solution.

src/style.css
VladimirTi: No, but it requires another one css rule :)

PR #127, opened 2019-07-24
added module backstopConfig.

gestap:


gestap:
> can we move these files in node_modules ?

Maybe we can, but it would be not trivial. 
We can delete a few files.

PR #131, opened 2019-07-27
Solution to "Moyo header"

VladyslavSak:


src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  margin: 0;
+}
+
+ul,
+li {
+  margin: 0;
+  padding: 0;
+}
+
+header {
```
irenhh: create a class name and style it

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  display: flex;
+  height: 60px;
```
irenhh: you can delete height here

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  height: 40px;
+}
+
+.main-menu {
+  display: flex;
+  list-style: none;
+}
+
+.main-menu_item {
+  height: 60px;
```
irenhh: delete height here

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  display: flex;
+  align-items: center;
```
irenhh: `align-items` isn't needed here

PR #153, opened 2019-07-30
solution

youvovk:


src/style.css
```diff
@@ -0,0 +1,63 @@
+* {
```
irenhh: if you need to reset indentations, do it for each of the block you need

src/style.css
```diff
@@ -0,0 +1,63 @@
...
+
+body {
+  font-family: "Roboto", serif;
+}
+
+ul {
+  list-style: none;
+}
+
+a {
```
irenhh: what if you have another `<ul>` and `<a>` on the page and you don't want to reset this style for them?...

src/style.css
```diff
@@ -0,0 +1,63 @@
...
+  text-decoration: none;
+}
+
+.header--container {
```
irenhh: why do you use `--` here? this syntax has special meaning in BEM

src/index.html
```diff
@@ -11,8 +11,46 @@
...
+  <link href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap" rel="stylesheet"/>
 </head>
 <body>
+    <header>
+      <div class="header--container">
+        <a class="header-logo" href="#">
+          <img src="images/logo.png" alt="moyo-logo">
+        </a>
+
+
```
irenhh: why so many blank lines?

src/style.css
youvovk: done

src/style.css
youvovk: done

src/index.html
youvovk: I like distance))

src/style.css
```diff
@@ -0,0 +1,56 @@
+body {
+  font-family: "Roboto", serif;
+  margin: 0;
+  padding: 0;
```
irenhh: body doesn't have default paddings

PR #170, opened 2019-09-02
Solution

olesyamogil:


src/index.html
```diff
@@ -9,11 +9,24 @@
...
+  <title>MOYO</title>
+  <link href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap" rel="stylesheet">
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+  <header class="main-header">
+    <img src="images/logo.png" alt="Logo" class="logo">
+    <nav class="main-menu">
+        <a href="#" class="menu-item is-active">apple</a>
```
gestap: use 2 whitespaces instead of 4

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.menu-item {
+  position: relative;
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+  font-size: 12px;
+  line-height: 60px;
+  font-family: "Roboto", sans-serif;
```

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+.menu-item:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  background-color: #00acdc;
+  width: 37px;
```

image

PR #171, opened 2019-09-02
solution

karpyshynroman:


src/style.css
```diff
@@ -0,0 +1,54 @@
+body {
+  margin: 0;
+}
+
+.nav-link {
+  position: relative;
+  font-family: "Roboto", sans-serif;
```

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+  background-color: #00acdc;
+}
+
+.nav-list {
+  list-style: none;
+  display: flex;
+  flex-direction: row;
+}
+
+.main {
```
gestap:
```suggestion
.logo {
   width: 40px;
   height: 40px;
 }
```

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+  margin-top: 5px;
```
gestap:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+  text-decoration: none;
+  text-transform: uppercase;
+  font-weight: 500;
+  font-size: 12px;
+  margin-left: 20px;
+  line-height: 60px;
+  color: #000;
+}
+
+.is-active {
```

image

src/style.css
```diff
@@ -0,0 +1,60 @@
...
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: 4px;
+  border-radius: 8px;
+  background-color: #00acdc;
+}
+
+.nav-list {
+  /*list-style: none;*/
```
gestap: ?

PR #174, opened 2019-09-03
MOYO

Annetastar:


src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  padding: 0 50px;
+}
+
+.moyo_nav {
+  display: flex;
+}
+
+.link_nav {
+  position: relative;
+  font-family: "Roboto", sans-serif;
```

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 60px;
+  text-decoration: none;
+  text-transform: uppercase;
+  color: black;
+  margin-left: 20px;
+}
+
+.is-active {
```

image

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+
+.link_nav {
+  position: relative;
+  text-decoration: none;
+  text-transform: uppercase;
+  color: black;
+  margin-left: 20px;
+}
+
+.is-active:first-child {
```
gestap: No, need to add it to class where we add this margin
```suggestion
.link_nav:first-child {
```

Annetastar: O, thanks, sorry))

PR #175, opened 2019-09-03
Fs aug19 andrii p

Sijey:


src/style.css
```diff
@@ -0,0 +1,53 @@
...
+}
+
+.navbar {
+  margin-right: 50px;
+  display: flex;
+}
+
+.nav_link {
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
```

src/style.css
```diff
@@ -0,0 +1,53 @@
...
+  margin-left: 20px;
+}
+
+.nav_link:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active {
+  margin-left: 0;
```

gestap: image

gestap:
Good job!

PR #177, opened 2019-09-03
Fs aug19 kolya

chepa17:


src/style.css
```diff
@@ -0,0 +1,54 @@
...
+
+.big-nav {
+  display: flex;
+}
+
+.link-nav {
+  position: relative;
+  text-transform: uppercase;
+  color: #000;
+  font-family: "Roboto", sans-serif;
```
gestap: move all typographic style body

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+.link-nav:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  background-color: #00acdc;
+  width: 37px;
```
gestap: image

gestap: if we add this class to another element?

PR #186, opened 2019-09-04
solution

kostya-flern:


src/index.html
```diff
@@ -13,7 +13,30 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
+  <header class="header">
+    <a href="#" class="logo">
+      <img src="/images/logo.png" alt="logo">
+    </a>
+    <nav class="navbar">
+      <ul class="navbar-list">
+        <li><a class="nav_link is-active" href="#">Apple</a></li>
```
gestap: remove all unnecessary empty line

src/style.css
```diff
@@ -0,0 +1,57 @@
...
+.navbar-list {
+  margin: 0 50px;
+  display: flex;
+  text-decoration: none;
+  list-style: none;
+  height: 60px;
+}
+
+.nav_link {
+  text-decoration: none;
```
gestap: better to add write this rule like that
```
a {
  text-decoration: none;
}
```

gestap: you can have many links on your website

src/style.css
```diff
@@ -0,0 +1,61 @@
...
+  margin-left: 20px;
+}
+
+.nav_link:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active {
+  margin-left: 0;
```
gestap: `is-active` must only change color and add bottom line
```suggestion
```
image

src/style.css
```diff
@@ -0,0 +1,61 @@
...
+  position: relative;
```
gestap: why you need this rule?
```suggestion
```

PR #189, opened 2019-09-05
Sulution, 2 tests fail

lubremets:


src/index.html
```diff
@@ -10,10 +10,25 @@
...
+  <link href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap&subset=cyrillic" rel="stylesheet">
   <link rel="stylesheet" href="./style.css">
+
 </head>
 <body>
-
+  <header class="header">
+    <nav class="nav">
+      <ul class="nav_list">
+        <li class="nav_item"><a class="link is-active" href="#1">APPLE</a></li>
```
gestap: image

remove this margin

src/style.css
```diff
@@ -0,0 +1,88 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  margin: 0;
```
gestap: you already add it to `*`
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+
+a {
+  text-decoration: none;
+}
+
+.header {
+  box-sizing: border-box;
+  height: 60px;
+  position: relative;
+  margin: 0;
```
gestap:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  display: flex;
+  justify-content: flex-end;
+}
+
+.nav_list {
+  display: flex;
+  flex-direction: row;
```
gestap: it's default value
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  justify-content: flex-end;
```
gestap: in your case this rule does nothing
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  align-items: center;
```
gestap: this too
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  margin: 0;
```
gestap:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  list-style-type: none;
```
gestap: better to write
```suggestion
ul {
  list-style-type: none;
}
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  position: absolute;
+  display: block;
+  width: 100%;
+  border-bottom: 4px solid #00acdc;
+  border-radius: 4px;
+  bottom: 0;
+}
+
+.is-active {
+  color: #00acdc !important;
```
gestap: NO! don't use important!

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  width: 100%;
+  border-bottom: 4px solid #00acdc;
+  border-radius: 4px;
+  bottom: 0;
+}
+
+.nav_item a {
+  color: black;
+  position: relative;
+  display: block;
```
gestap:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  margin: 0;
+  list-style-type: none;
+}
+
+.nav_list:last-child {
+  margin-right: 50px;
+}
+
+.nav_item {
+  line-height: 60px;
```
gestap:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  position: absolute;
+  display: block;
+  width: 100%;
+  border-bottom: 4px solid #00acdc;
+  border-radius: 4px;
+  bottom: 0;
+}
+
+.nav_item a {
+  color: black;
```
gestap:
```suggestion
  color: black;
  line-height: 60px;
```

src/style.css
lubremets: Why is this better?

src/style.css
lubremets: I use !important only for tests 'class="is-active"'

src/style.css
lubremets: I use it, because I want click on whole block, not just click on the text

src/style.css
gestap: you can have many `ul` on your page, better to write this rule once.

src/style.css
gestap: why you need `important` for this test?

gestap: rewrite style without important

src/style.css
lubremets: I fix this.

src/style.css
lubremets: Because, I need more specificity for this class.
I rewrite this style.

PR #262, opened 2020-02-27
Solution

SeiAlek:
- [DEMO LINK](https://SeiAlek.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://SeiAlek.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,75 @@
+html,
+body {
+  margin: 0;
+  background-color: #e5e5e5;
+}
+
+.header {
+  display: flex;
+  height: 60px;
+  min-width: 1024px;
+  margin: 0 auto;
+  padding: 0 50px;
+  background-color: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  box-sizing: border-box;
+  justify-content: space-between;
+}
+
+.header-logo {
+  display: inline-block;
+  height: 40px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,75 @@
...
+
+.header-menu {
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.header-menu > li {
+  list-style-type: none;
+}
```
vkryvytskyy: remove this and write this rule inside `.header-menu--item`

vkryvytskyy: Try to avoid increasing the specificity of the selector. One class is enough to identify the elements on the page. No need to write something like .block.colored-block. Use .block and .block--colored instead, and apply each of them to your element

src/style.css
```diff
@@ -0,0 +1,75 @@
...
+
+.header-menu--link:hover {
+  color: #00acdc;
+}
+
+.is-active.header-menu--link {
+  color: #00acdc;
+}
+
+.is-active.header-menu--link::after {
```

src/style.css
```diff
@@ -0,0 +1,75 @@
...
+  text-transform: uppercase;
+  text-decoration: none;
+  color: #000;
+}
+
+.header-menu--link:hover {
+  color: #00acdc;
+}
+
+.is-active.header-menu--link {
```

PR #268, opened 2020-02-27
Solution

yurez91:
- [DEMO LINK](https://yurez91.github.io/layout_moyo-header/)
   - [TEST REPORT LINK](https://yurez91.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,25 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Moyo header</title>
   <link rel="stylesheet" href="./style.css">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap" rel="stylesheet">
 </head>
 <body>
-  <h1>Moyo header</h1>
+  <header class="header">
+    <div class="logo">
+      <a href="#"><img src="./images/logo.png" alt="logo"></a>
+    </div>
+    <nav>
+      <ul class="navbar">
+        <li class="navbar-item"><a href="#" class="navbar-link is-active">Apple</a></li>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -7,8 +7,25 @@
...
 </head>
 <body>
-  <h1>Moyo header</h1>
+  <header class="header">
+    <div class="logo">
+      <a href="#"><img src="./images/logo.png" alt="logo"></a>
+    </div>
+    <nav>
+      <ul class="navbar">
+        <li class="navbar-item"><a href="#" class="navbar-link is-active">Apple</a></li>
```
vkryvytskyy: One tag - one line. Use new lines for nested tags. Don't write nested tags in one line.

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.navbar-item {
+  list-style: none;
+  margin-right: 20px;
+}
+
+.navbar .navbar-item:last-child {
```

src/index.html
```diff
@@ -7,8 +7,41 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Moyo header</title>
   <link rel="stylesheet" href="./style.css">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap" rel="stylesheet">
 </head>
 <body>
-  <h1>Moyo header</h1>
+  <header class="header">
+    <div class="logo">
+      <a href="#"><img src="./images/logo.png" alt="logo"></a>
```
vkryvytskyy: One tag - one line. Use new lines for nested tags. Don't write nested tags in one line.

PR #272, opened 2020-02-28
Solution

romarios1987:
- [DEMO LINK](https://romarios1987.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://romarios1987.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,45 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Moyo header</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Moyo header</h1>
-</body>
+  <head>
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Moyo header</title>
+
+    <link
+      href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap"
+      rel="stylesheet"
+    />
+    <link rel="stylesheet" href="./style.css" />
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  justify-content: space-between;
+  padding: 0 50px;
+  background-color: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  height: 60px;
+}
+
+.header__logo img {
+  margin-top: 2px;
+}
```
vkryvytskyy: Try to avoid stylizing tags. Always use classes for this. Tags stylizing is OK when you reset default browser styles, like margin: 0 for BODY tag, which has 8px margins. Setting width and height for all DIV tags is bad practice

vkryvytskyy: Try to avoid increasing the specificity of the selector. One class is enough to identify the elements on the page. No need to write something like .block.colored-block. Use .block and .block--colored instead, and apply each of them to your element

romarios1987: ubdated

src/index.html
```diff
@@ -1,14 +1,45 @@
...
+          <li class="nav__item"><a href="#" class="nav__link">Samsung</a></li>
+          <li class="nav__item">
+            <a href="#" class="nav__link">Смартфоны и телефоны</a>
+          </li>
+          <li class="nav__item">
+            <a href="#" class="nav__link" data-qa="hover"
+              >Ноутбуки и компьютеры</a
+            >
+          </li>
+          <li class="nav__item"><a href="#" class="nav__link">Гаджеты</a></li>
```
vkryvytskyy: One tag - one line. Use new lines for nested tags. Don't write nested tags in one line.

src/index.html
```diff
@@ -1,14 +1,45 @@
...
+    <link
+      href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap"
+      rel="stylesheet"
+    />
+    <link rel="stylesheet" href="./style.css" />
+  </head>
+  <body>
+    <header class="header">
+      <a class="header__logo" href="/"
+        ><img class="header__logo__img" src="./images/logo.png" alt="Moyo"
```
vkryvytskyy: this looks strange

vkryvytskyy:
```suggestion
      <a class="header__logo" href="/">
        <img class="header__logo__img" src="./images/logo.png" alt="Moyo"/>
```

romarios1987: it's okay now ?

src/index.html
romarios1987: edited

PR #316, opened 2020-03-09
solution

coastin:
- [DEMO LINK](https://coastin.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://coastin.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,7 +13,25 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+  <header class="menu">
+    <section class="navigation">
+        <section class="logo">
+            <a href="#">
+              <img src="images/logo.png" alt="Логотип MoYo">
+            </a>
+        </section>
+          <nav class="navbar">
+            <a href="#" class="nav-link is-active">Apple</a>
+            <a href="#" class="nav-link">Samsung</a>
+            <a href="#" class="nav-link">Смарфоны и телефоны</a>
+            <a href="#" class="nav-link" data-qa="hover">Ноутбуки и компьютеры</a>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -13,7 +13,25 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+  <header class="menu">
+    <section class="navigation">
+        <section class="logo">
+            <a href="#">
+              <img src="images/logo.png" alt="Логотип MoYo">
+            </a>
```
vkryvytskyy: Follow the indentation rules. Add an additional space of 2 spaces for nested tags/text. 4 - is too much.

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.navigation {
+  display: flex;
+
+}
+
+.logo a {
```

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  padding: 10px 0;
+}
+
+.navbar {
+  display: flex;
+  margin-left: auto;
+  align-items: center;
+}
+
+.navbar .nav-link {
```

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+.navbar .navLink:hover {
+  color: #00acdc;
+}
+
+.navbar .is-active {
+  color: #00acdc;
+  position: relative;
+}
+
+.navbar .is-active::after {
```

vkryvytskyy:
fix previous comments
on hover your links doesn't change color

PR #320, opened 2020-04-24
header solution denkondratiev

denkondratiev:
- [DEMO LINK](https://denkondratiev.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://denkondratiev.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,77 @@
+@import url("https://fonts.googleapis.com/css?family=Roboto:500&display=swap");
+
+body {
+  margin: 0;
+  background-color: #000;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.container,
+.menu-list {
+  display: flex;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,77 @@
...
+  background-color: #000;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.container,
+.menu-list {
+  display: flex;
```
irenhh, line 14: display styles better to be written for each element separately

src/style.css
```diff
@@ -0,0 +1,77 @@
...
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.menu-list {
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
```
irenhh, line 34: aren't padding 0 already?

PR #353, opened 2020-07-02
Develop

ArtemiiKab:
- [DEMO LINK](https://ArtemiiKab.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ArtemiiKab.github.io/layout_moyo-header/report/html_report/) 

4 bottom tests are not passed. Discussed with Dennis on QA session. 
 Since bottom tests take links with space between, it would take unnecessary additional code lines with paddings added to my nav_links just to pass the tests. So it was decided not to put them.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,5 +19,5 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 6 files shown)


ArtemiiKab:
Realized that test is right, and links should have space around them equal to header hight in order to make the clickable area bigger.

PR #366, opened 2020-07-04
Solution

furude-rika:
- [DEMO LINK](https://furude-rika.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://furude-rika.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+}
+
+.nav-header {
+  display: flex;
+  justify-content: space-between;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.first-item {
+  position: relative;
+}
+
+.underline-link::after {
+  content: "";
+  position: absolute;
+  width: 37px;
```
vsm2k17, line 66:
```suggestion
  width: 100%;
```

vsm2k17: if you change the text inside, the width will be different

PR #367, opened 2020-07-04
Solution

uni-8:
- [DEMO LINK](https://uni-8.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://uni-8.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,60 @@
+html {
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 60px;
+  padding: 0 50px;
+  background-color: #fff;
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.menu__list {
+  list-style: none;
+  padding: 0;
+  margin: 0;
+  display: flex;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,62 @@
...
+  padding: 0;
+  margin: 0;
+  display: flex;
+}
+
+.menu__link {
+  text-decoration: none;
+  display: block;
+  position: relative;
+  padding: 23px 0;
```
vsm2k17, line 35:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+  margin-left: 20px;
+  font-size: 12px;
+  line-height: 14px;
```
vsm2k17, line 38:
```suggestion
  line-height: 60px;
```

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+  font-weight: 500;
+  text-transform: uppercase;
+  color: #000;
+}
+
+.first_link {
+  margin: 0;
+}
```
vsm2k17, line 46: you can use pseudo-class :first-child here, instead of creating an extra class

uni-8: if i do that all the links lose their margin cuz they are all first children in <li> tags

PR #382, opened 2020-07-06
flex box

zarva6596:
[DEMO LINK](https://zarva6596.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://zarva6596.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -1,14 +1,37 @@
...
+      rel="stylesheet"
+    >
+  </head>
+  <body>
+    <header class="header">
+      <img
+        src="images/logo.png"
+        alt="logo"
+        class="logo"
+      >
```
vsm2k17, line 20: moyo logo should be a link that leads to home page

src/style.css
```diff
@@ -0,0 +1,60 @@
...
+  top: 10px;
+}
+
+.nav {
+  display: flex;
+  justify-content: flex-end;
+}
+
+.link {
+  padding: 22.5px 0;
```
vsm2k17, line 29: try to avoid float numbers, you can use line-height equal height of your container in this case

PR #387, opened 2020-07-06
solution

zkhomer:
- [DEMO LINK](https://zkhomer.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://zkhomer.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,76 @@
+
+html {
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+ul {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,75 @@
...
+  color: #000;
+}
+
+.main-nav-link:hover {
+  color: #00acdc;
+}
+
+.first-link {
+  margin-left: 0;
+}
```
vsm2k17, line 60: use pseudo-class :first-child instead of this class

PR #485, opened 2020-11-14
MOYO-header solution

Roman-Matsuk:
- [DEMO LINK](https://Roman-Matsuk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Roman-Matsuk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,75 @@
+body {
+  font-size: 12px;
+  font-family: "Roboto", sans-serif;
+}
+
+* {
+  margin: 0;
+  box-sizing: content-box;
+}
+
+.header-nav {
+  display: flex;
+  justify-content: space-between;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,76 @@
...
+.header-nav {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+.nav {
+  display: flex;
+  background-color: #fff;
```
aionova, line 20: White is default background-color

aionova (approved):
Well done! :fire:

src/style.css
Roman-Matsuk: oops :)

PR #506, opened 2020-11-16
Develop

SofiiaAsieieva:
[DEMO LINK](https://SofiiaAsieieva.github.io/layout_moyo-header/)<br>
[TEST REPORT LINK](https://SofiiaAsieieva.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -9,66 +9,47 @@ html {
 
 body {
   margin: 0;
+  /*background-color: black;*/
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
+  align-content: flex-start;
+  position: relative;
+  right: 0;
+  left: 0;
+  height: 60px;
+  /*background-color: rgba(255, 255, 255, 1);*/
   background-color: black;
 }
 
-.nav {
-  position: relative;
-  background-color: rgba(255, 255, 255, 1);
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,79 @@
...
+
+body {
+  margin: 0;
+  background-color: black;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+}
+
+header {
+  height: 60px;
```
aionova, line 13: It's better not to set height in px, but to adjust it with paddings

src/style.css
```diff
@@ -0,0 +1,79 @@
...
+}
+
+.is-active::after {
+  content: "";
+  width: 37px;
+  height: 4px;
+  background-color: rgba(0, 172, 220, 1);
+  border-radius: 8px;
+  position: absolute;
+  top: calc(100% - 4px);
```
aionova, line 78: height may change on different screens, so better to set position from bottom, not top

aionova (changes requested):

src/style.css
aionova: Sorry, I was wrong with this

aionova (approved):
Great!)

PR #508, opened 2020-11-16
header is created

HaidaiAndrii:
- [DEMO LINK](https://HaidaiAndrii.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://HaidaiAndrii.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  width: 40px;
+  height: 40px;
+}
+
+.items {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  list-style-type: none;
+  height: 60px;
```
aionova, line 32: what happens if you remove height here?

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  padding: 0;
+}
+
+.nav_item {
+  font-family: inherit;
```
aionova, line 37: font-family is inherited by default

aionova (changes requested):
good start!

src/style.css
HaidaiAndrii: nothing, header still 60px in height

src/style.css
HaidaiAndrii: removed

aionova (approved):
Awesome!

PR #589, opened 2021-03-26
Moyo-header

Andriykooo:
- [DEMO LINK](https://Andriykooo.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Andriykooo.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -7,8 +7,6 @@ body {
 }
 
 header {
-  max-width: 1200px;
-  
   display: flex;
   justify-content: space-between;
   align-items: center;
--- a/src/style.css
+++ b/src/style.css
@@ -1,13 +1,14 @@
 * {
   box-sizing: border-box;
+}
+
+body {
   margin: 0;
-  padding: 0;
 }
 
 header {
   max-width: 1200px;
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,60 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
vsm2k17, line 4: too heave for `*` selector

src/style.css
```diff
@@ -0,0 +1,60 @@
...
+}
+
+header {
+  max-width: 1200px;
+  min-width: 1024px;
```
vsm2k17, line 9: bad idea, remove min-width. the horizontal scroll appears on your page on small screens.

vsm2k17 (changes requested):
almost done

vsm2k17 (changes requested):
image
that's how it looks on the big screen, let's make it flexible so on big screens navigation gonna stick to the right side of the screen.

vsm2k17 (approved):
good job

PR #598, opened 2021-03-28
Solution

herkoss:
https://herkoss.github.io/layout_moyo-header/
https://herkoss.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,38 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Moyo header</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Moyo header</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,14 +1,38 @@
...
+    <link rel="stylesheet" href="style.css">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet">
+  </head>
+  <body>
+    <header class="header">
+      <div class="container">
+        <img class="moyo__logo" src="images/logo.png"
```
vsm2k17, line 18: should be clickable link

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.product__link:hover {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  width: 100%;
```
vsm2k17, line 54: redundant if you use left 0 and right 0

vsm2k17 (changes requested):
almost done

src/style.css
```diff
@@ -0,0 +1,70 @@
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  display: flex;
+}
+
+.container {
```
vsm2k17, line 10: the container is redundant just style header, and use paddings instead of margins

vsm2k17 (approved):
great

PR #604, opened 2021-03-29
solution

IvanKolodiazhnyi:
- [DEMO LINK](https://IvanKolodiazhnyi.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://IvanKolodiazhnyi.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,60 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Moyo header</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Moyo header</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Moyo header</title>
+    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,14 +1,60 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Moyo header</title>
+    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
+  </head>
+  <body>
+    <header class="header">
+      <div>
+        <img
```
vsm2k17, line 18: should be clickable link

src/style.css
```diff
@@ -0,0 +1,68 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
```
vsm2k17, line 4: use only box-size property on  * selector , other properties are too heavy

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  --color_text: #00acdc;
+}
+
+.header {
+  display: flex;
+  padding: 0 50px;
+  max-width: 1200px;
+  min-width: 1024px;
```
vsm2k17, line 12: bad idea don't set width explicitly

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.moyo__logo {
+  position: absolute;
+  top: 10px;
+}
+
+.nac {
+  /* Я не знаю чи варто щось сюда писати, або щось перенести */
```
vsm2k17, line 24: you don't have class `nac`, so remove this

src/style.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: 4px;
+  width: 37px;
```
vsm2k17, line 65: you already wrote left: 0 and right: 0, no need to set width

vsm2k17 (changes requested):
almost done

PR #626, opened 2021-03-29
solution

Allrightman:
[DEMO LINK](https://Allrightman.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://Allrightman.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,59 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Moyo header</title>
-  <link rel="stylesheet" href="./style.css">
-</head>
-<body>
-  <h1>Moyo header</h1>
-</body>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,14 +1,59 @@
...
+    <title>Moyo header</title>
+    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet">
+  </head>
+  <body>
+    <header class="header">
+      <img
```
vsm2k17, line 17: should be clickable link

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+.is-active,
+.nav__link:hover {
+  color: var(--color-blue);
+}
+
+.is-active::after {
+  position: absolute;
+  content: "";
+  bottom: 0;
+  width: 37px;
```
vsm2k17, line 65:
```suggestion
  width: 100%;
```
in that way it will work for every link does not matter what length it would be

vsm2k17 (approved):
good job, don't forget to fix comments

src/index.html
Allrightman: Thanks a lot!

PR #628, opened 2021-03-30
pending remarks

amithos:
I`m sorry about the links. Deploy doesn't work, but I'll try to fix it later, on Q&A

- [DEMO LINK](https://amithos.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://amithos.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,67 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+* {
+  padding: 0;
+  margin: 0;
+}
+
+body {
+  --logo-size: 40px;
+  --brand-color: #00acdc;
+  --header-height: 60px;
+  --header-content-margin: 50px;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.header {
+  display: flex;
+  height: var(--header-height);
+  min-width: 1024px;
+  justify-content: space-between;
+  align-items: center;
+}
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,67 @@
...
+  --header-content-margin: 50px;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.header {
+  display: flex;
+  height: var(--header-height);
+  min-width: 1024px;
```
vsm2k17, line 21: remove, that's a bad idea, your code should be flexible

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+.links:hover {
+  color: var(--brand-color);
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  right: 10px;
+  left: 10px;
```
vsm2k17, line 63: avoid this, problem is your link has a margin, better to add them on your nav__link
```suggestion
  right: 0;
  left: 0;
```

vsm2k17 (changes requested):
image
shouldn't be a margin on the left side of it 
image
shouldn't be a margin on the right side of it

vsm2k17 (changes requested):
image
if you remove margin and padding from ul tests will pass

vsm2k17 (approved):
image
remove left margin from the first link and right margin from the last one

PR #631, opened 2021-03-30
Solution

vitaliy-buchinchik:
- [DEMO LINK](https://vitaliy-buchinchik.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://vitaliy-buchinchik.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,70 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  --header-height: 60px;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  height: var(--header-height);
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,70 @@
...
+
+.nav {
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.nav__list {
+  margin: 0;
+  padding: 0;
```
vsm2k17, line 35: if you want to remove default values like margin and padding remove them for ALL tags, use tag for this instead of a specific class

src/style.css
```diff
@@ -0,0 +1,70 @@
...
+  list-style: none;
+}
+
+.nav__item {
+  display: flex;
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+  margin-left: 0;
```
vsm2k17, line 46: remove

vsm2k17 (approved):
good job

PR #774, opened 2021-07-07
header done

nastyyvolk:
- [DEMO LINK](https://nastyyvolk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://nastyyvolk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,65 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.header {
+  position: sticky;
+  height: 60px;
+  top: 0;
+  display: flex;
+  padding: 10px 50px 10px 22.5px ;
+  justify-content: space-between;
+  align-items: center;
+
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+
+  box-shadow: 5px 0 3px 1px lightgray;
+  color: black;
+}
+
+.image__container {
+  width: 40;
+  height: 40;
+}
+
+.nav {
+  display: flex;
+}
+
+.nav__list {
+  list-style-type: none;
+  display: flex;
+}
+
+.nav__item {
+  margin-left: 20px;
+}
+
+.nav__link {
+  position: relative;
+  display: block;
+  height: 60px;
+
+  line-height: 60px;
+  color: black;
+  text-decoration: 0;
+}
+
+.nav__link:hover,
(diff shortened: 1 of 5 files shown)


danheim:
не соответствует дизайну

src/style.css
```diff
@@ -0,0 +1,65 @@
+body {
+  margin: 0;
+  padding: 0;
```
danheim, line 3: remove padding, body doesn't have it

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+.header {
+  position: sticky;
+  height: 60px;
+  top: 0;
+  display: flex;
+  padding: 10px 50px 10px 22.5px ;
+  justify-content: space-between;
+  align-items: center;
+
+  font-family: Roboto, sans-serif;
```
danheim, line 15: move font family to html

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  font-weight: 500;
+  font-size: 12px;
+
+  box-shadow: 5px 0 3px 1px lightgray;
+  color: black;
```
danheim, line 20: use hex

```suggestion
  color: #000;
```

danheim (changes requested):
Great start

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.nav {
+  display: flex;
+}
+
+.nav__list {
+  list-style-type: none;
+  display: flex;
+  margin-right: 50px;
```
varseniuk, line 36: You could use `padding: 0 50px` for your header instead of giving margins for logo and nav

varseniuk (changes requested):
Let's fix all comments that @danheim and I left and it's done

src/style.css
nastyyvolk: sorry, i commited the wrong file

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  height: 3px;
```
varseniuk, line 58:
```suggestion
  height: 4px;
```
It should be 4px according to the design

varseniuk (approved):

PR #795, opened 2021-07-07
Moyo header

kristof22122:
- [DEMO LINK](https://kristof22122.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://kristof22122.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,66 @@
+body {
+  margin: 0;
+}
+
+.header {
+  height: 60px;
+}
+
+.wrapper {
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo-img {
+  width: 40px;
+  height: 40px;
+  margin-left: 50px;
+}
+
+.nav-list {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  list-style: none;
+}
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,66 @@
+body {
+  margin: 0;
+}
+
+.header {
+  height: 60px;
+}
+
+.wrapper {
+  font-family: "Roboto", sans-serif;
```
varseniuk, line 10:
```suggestion
  font-family: "Roboto", sans-serif;
```
Move font family to `html` selector

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  align-items: center;
+  list-style: none;
+}
+
+.nav-item {
+  margin-right: 20px;
+}
+
+.nav-list > .nav-item:last-child {
+  margin-right: 50px;
```
varseniuk, line 42: You can use padding:0 50px for your wrapper. It would be more logical
Here. for the last-child you can use margin-right: 0, because last item should not have it

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.nav-link {
+  text-decoration: none;
+  color: #000;
+  line-height: 60px;
```
varseniuk, line 48: You could center the text inside it using `display:flex` and `align-items:center`

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  background-color: #00acdc;
+  border-radius: 8px;
+  width: 100%;
+  height: 2px;
```
varseniuk, line 64:
```suggestion
  height: 4px;
```
It should be 4px by design

varseniuk (changes requested):
Almost done. Let's make some changes to styles

varseniuk (approved):

PR #802, opened 2021-07-07
done moyo

Andrii-Kalaida:
https://Andrii-Kalaida.github.io/layout_moyo-header/
https://Andrii-Kalaida.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -57,7 +57,7 @@ html {
 .is-active:after {
   width: 100%;
   height: 4px;
-  content: '';
+  content: "";
   background-color: #00acdc;
   left: 0;
   bottom: 0;
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,65 @@
+html {
+  font-family: "Roboto", "Arial", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  width: 100%;
+  min-height: 60px;
+  background-color: #fff;
+}
+
+.logo {
+  position: absolute;
+  left: 50px;
+}
+
+.nav {
+  margin-right: 40px;
+  text-align: center;
+}
+
+.item {
+  height: 100%;
+  display: block;
+  margin: 0 10px;
+}
+
+.list {
+  display: flex;
+  align-items: stretch;
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
+.link {
+  display: block;
+  padding: 22px 0;
+  text-decoration: none;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,65 @@
+html {
+  font-family: "Roboto", "Arial", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: flex-end;
```
dima-tkachenko, line 7: try to use justify-content: space-between;;

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  align-items: center;
+  width: 100%;
+  min-height: 60px;
```
dima-tkachenko, line 10: bad idia use height to header

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  background-color: #fff;
+}
+
+.logo {
+  position: absolute;
+  left: 50px;
+}
+
+.nav {
+  margin-right: 40px;
```
dima-tkachenko, line 20: try add padding to header

dima-tkachenko (changes requested):

src/style.css
Andrii-Kalaida: agree. it is better to use here padding

src/style.css
Andrii-Kalaida: space-beetween breaks my code

src/style.css
Andrii-Kalaida: I think It's a good idea to use min-height for the header if we use flexbox to align items, min-height means that container will stretch if needed

src/style.css
danheim: flex-end looks weird

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+  justify-content: flex-end;
+  align-items: center;
+  width: 100%;
+  min-height: 60px;
+  background-color: #fff;
+}
+
+.logo {
+  position: absolute;
+  left: 50px;
```
danheim, line 16: use margin, remove absolute

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.nav {
+  padding-right: 40px;
+  text-align: center;
+}
+
+.item {
+  height: 100%;
+  display: block;
```
danheim, line 26: inline block would be better

danheim (changes requested):

Denchepornyuk (approved):

PR #925, opened 2021-09-08
header was made

podlesnyi-pavel:
- [DEMO LINK](https://podlesnyi-pavel.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://podlesnyi-pavel.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,79 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-size: 12px;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  --main-color: #00acdc;
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+  line-height: 14px;
+}
+
+.header__logo {
+  height: 40px;
+  width: 40px;
+  display: flex;
+  margin: auto 0;
+}
+
+.nav {
+  text-transform: uppercase;
+}
+
+.nav__list {
+  display: flex;
+  /* gap: 0 20px; */
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
+.nav__item + .nav__item {
+  margin: 0 0 0 20px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,79 @@
...
+html {
+  font-size: 12px;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  --main-color: #00acdc;
```
VinnieJ-2k20, line 14: Better to add it in root element
https://developer.mozilla.org/en-US/docs/Web/CSS/:root

src/style.css
```diff
@@ -1 +1,79 @@
...
+  margin: auto 0;
+}
+
+.nav {
+  text-transform: uppercase;
+}
+
+.nav__list {
+  display: flex;
+  /* gap: 0 20px; */
```
VinnieJ-2k20, line 37: Don't leave commented code.

src/style.css
```diff
@@ -1 +1,79 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.nav__item + .nav__item {
+  margin: 0 0 0 20px;
+}
+
+.nav__link {
+  /* --pw-link: 10px; */
```
VinnieJ-2k20, line 48: Same.

src/style.css
```diff
@@ -1 +1,79 @@
...
+
+.is-active::after {
+  content: "";
+  width: 100%;
+  /* width: calc(100% - var(--pw-link) * 2); */
+  height: 4px;
+  position: absolute;
+  bottom: 0;
+  left: 50%;
+  transform: translateX(-50%);
```
VinnieJ-2k20, line 72: Better to use `left: 0` and `right: 0` instead of width and transform.

VinnieJ-2k20 (changes requested):
Solution is good, but let's make your code even better.

src/style.css
```diff
@@ -1 +1,79 @@
...
+  transition: all 0.2s;
+}
+
+.is-active {
+  color: var(--main-color);
+}
+
+.is-active::after {
+  content: "";
+  width: 100%;
```
VinnieJ-2k20, line 67: Useless with `left` and `right` coordinates present.

VinnieJ-2k20 (approved):
Good job 👍

PR #987, opened 2021-10-13
mayo header

DmytroKashenok:
- [DEMO LINK](https://DmytroKashenok.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://DmytroKashenok.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,67 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+}
```
illiarerun, line 4: Don't reset your styles with * selector, move it to reset.css and reset only for specific tags

src/style.css
```diff
@@ -1 +1,67 @@
...
+
+html {
+  box-sizing: border-box;
```
illiarerun, line 7: This is what you can move to * selector

PR #991, opened 2021-10-13
Moyo header completed

Vitaliy-0:
Demo
https://Vitaliy-0.github.io/layout_moyo-header/
Test report
https://Vitaliy-0.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,6 +1,9 @@
 * {
-  font-family: "Roboto", sans-serif;
   box-sizing: border-box;
+}
+
+html {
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -1 +1,54 @@
-/* Styles go here */
+* {
+  font-family: "Roboto", sans-serif;
```
AlinaMatuschak, line 2: Don't add fonts for all elements with * selector. Use html selector

PR #1000, opened 2021-10-13
Develop

IrynaHerasymchuk:
[DEMO LINK](https://IrynaHerasymchuk.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://IrynaHerasymchuk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,6 +1,4 @@
 * {
-  margin: 0;
-  padding: 0;
   box-sizing: border-box;
 }
 
@@ -31,7 +29,7 @@ html {
   text-decoration: none;
   text-transform: uppercase;
 
-  display: inline-block;
+  display: block;
   padding: 23px 0;
 }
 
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,56 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
```
illiarerun, line 3: Don't reset your margins and paddings with * selector, better move it to reset.css and reset them for specific tags

src/style.css
```diff
@@ -1 +1,56 @@
...
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  color: #000;
+  text-decoration: none;
+  text-transform: uppercase;
+
+  display: inline-block;
```
illiarerun, line 34: No need of using inline-block here, block would be enough
```suggestion
  display: block;
```

illiarerun (approved):
LGTM 🔥

PR #1001, opened 2021-10-13
moyo header

MagicBox15:
- [DEMO LINK](https://MagicBox15.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://MagicBox15.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -21,7 +21,7 @@ html {
 
 .nav__item {
   list-style: none;
-  display: inline-block;
+  display: block;
   text-transform: uppercase;
 }
 
@@ -30,12 +30,11 @@ html {
 }
 
 .nav__link {
-  display: inline-block;
+  display: block;
   text-decoration: none;
   color: #000;
   font-size: 12px;
   font-weight: 500;
-  /* margin: 0 10px; */
   padding: 23px 0;
 }
 


src/style.css
```diff
@@ -1 +1,60 @@
...
+
+.logo {
+  padding: 10px 0;
+  height: 40px;
+  width: 40px;
+}
+
+.nav__item {
+  list-style: none;
+  display: inline-block;
```
illiarerun, line 24: No need to use inline-block here, block would be enough, u have the flex wrapper

src/style.css
```diff
@@ -1 +1,60 @@
...
+  text-transform: uppercase;
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  display: inline-block;
```
illiarerun, line 33: Same here

src/style.css
```diff
@@ -1 +1,60 @@
...
+  text-decoration: none;
+  color: #000;
+  font-size: 12px;
+  font-weight: 500;
+  /* margin: 0 10px; */
```
illiarerun, line 38: Don't leave commented code like that
```suggestion
```

illiarerun (changes requested):
Great job 👏 Fix these and it will be perfect

src/style.css
MagicBox15: ok, cleared

PR #1016, opened 2021-10-17
moyo header trial 1

dlugash:
- [DEMO LINK](https://dlugash.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://dlugash.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,81 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home"
+        class="logo"
+      >
+        <img
+          src="./images/logo.png"
+          alt="logo"
+        >
+      </a>
+      <nav>
+        <ul class="nav">
+          <li class="nav__item">
+            <a href="#apple"
+              class="is-active nav__link"
+            >
+              Apple
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung"
+              class="nav__link"
+            >
+              Samsung
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#смартфоны_и_телефоны"
+              class="nav__link"
+            >
+              Смартфоны и телефоны
+            </a>
+          </li>
+          <li class="nav__item"
+            data-qa="hover"
+          >
+            <a href="#ноутбуки_и_компьютеры"
+              class="nav__link"
+            >
+              Ноутбуки и компьютеры
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#гаджеты"
+              class="nav__link"
+            >
+              Гаджеты
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#планшеты"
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -6,9 +6,81 @@
...
     <title>Moyo header</title>
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home"
+        class="logo"
+      >
```
illiarerun, line 16: If you have less than 3 attibutes, there is no need to place them on a newline, you may fix it everywhere
```suggestion
      <a href="#home" class="logo">
```

src/style.css
```diff
@@ -1 +1,75 @@
 /* Styles go here */
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.header {
+  width: 1200px;
```
illiarerun, line 8: There is no need for that, your header should be responsive
```suggestion
```

src/style.css
```diff
@@ -1 +1,75 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  box-sizing: border-box;
```
illiarerun, line 12: Move box-sizing to * selector, it should be the same for all elements

src/style.css
```diff
@@ -1 +1,75 @@
...
+  margin: 0 auto;
+  padding: 0 50px;
+  background-color: #fff;
+}
+
+@media (max-width: 1024px) {
+  .header {
+    width: 1024px;
+  }
+}
```
illiarerun, line 22: Unnecessary, again
```suggestion
```

src/style.css
```diff
@@ -1 +1,75 @@
...
+
+.logo {
+  display: inline-block;
```
illiarerun, line 25: You have a flex wrapper around, you don't need inline-block here, you can just make it block
```suggestion
  display: block;
```

src/style.css
```diff
@@ -1 +1,72 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin: 0 auto;
+  padding: 0 50px;
+  background-color: #fff;
+}
+
+.logo {
+  display: block;
```
ihor-jpeg, line 22: if the parent has `display: flex;` then all its direct childrens' `display` prop. will be set to `block` automatically. no need for this

src/style.css
```diff
@@ -1 +1,72 @@
...
+}
+
+.is-active {
+  position: relative;
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  width: 100%;
```
ihor-jpeg, line 64: if you set
left: 0;
right: 0;

then the width will be 100% automatically. no need for this

src/style.css
```diff
@@ -1 +1,72 @@
 /* Styles go here */
+* {
+  max-width: 1200px;
```
ihor-jpeg, line 3: you definitely don't want the `max-width` prop. to be applied to everything. HTML-wise, usually you create a `div` container inside your `main`, which contains all the code and which only purpose is to limit the maximum width.
but for now you can just move it to the `.header`

src/style.css
```diff
@@ -1 +1,72 @@
...
+  margin: 0 auto;
+  padding: 0 50px;
+  background-color: #fff;
+}
+
+.logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+  padding: 10px 0;
```
illiarerun, line 25: Also you can remove these paddings, as you have align-items center

illiarerun (approved):
Good job 🔥 But make sure to fix my and Ihor's comments

PR #1062, opened 2021-11-18
Solution

Valeria-Valuyskaya:
- [DEMO LINK](https://Valeria-Valuyskaya.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Valeria-Valuyskaya.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,60 @@
 /* Styles go here */
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 16px;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  display: flex;
+  padding: 0 50px;
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  justify-content: space-between;
+  align-items: center;
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.nav__item {
+  display: inline-block;
+  padding-right: 20px;
+}
+
+.nav__link {
+  font-style: normal;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1 +1,60 @@
 /* Styles go here */
+html {
```
Elena-brv, line 2:
```suggestion
html {
```

Elena-brv (changes requested):
- this indent shouldn't be here
image

Elena-brv (changes requested):
- indent decreased but still here 
image

src/style.css
```diff
@@ -1 +1,61 @@
...
+  width: 40px;
+  height: 40px;
+}
+
+.nav {
+  display: inline-block;
+}
+
+.nav__item {
+  display: inline-block;
```
Elena-brv, line 28: no need, better add display: flex; for .nav__list

src/style.css
```diff
@@ -1 +1,61 @@
...
+  padding-right: 20px;
```
Elena-brv, line 29:
```suggestion
  margin-right: 20px;
```

but 
image

image

Elena-brv (changes requested):

src/style.css
```diff
@@ -1 +1,69 @@
...
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.nav {
+  display: flex;
+}
```
Elena-brv, line 25: no need

Elena-brv (approved):

PR #1092, opened 2021-11-23
moyo navigation bar

stanislav-bzhezinskyi:
https://stanislav-bzhezinskyi.github.io/layout_moyo-header/
https://stanislav-bzhezinskyi.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,74 @@
-/* Styles go here */
+.body {
+  margin: 0;
+  background-color: cadetblue;
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,74 @@
...
+  height: 60px;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
+
+.is-active,
+.nav__link {
+  flex-direction: column;
```
vkryvytskyy, line 50: А зачем тут делать flex-direction column?

vkryvytskyy (approved):

PR #1119, opened 2021-12-21
Moyo navbar added v.1

RomaBoreiko:
- [DEMO LINK](https://RomaBoreiko.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://RomaBoreiko.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,77 @@
 <!doctype html>
-<html lang="en">
+<html lang="ru">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <!-- <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    > -->
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -1,14 +1,77 @@
...
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <!-- <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    > -->
```
Anna-Tkachenko, line 14: Uncomment this code for import needed styles

src/style.css
```diff
@@ -1 +1,71 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
```
Anna-Tkachenko, line 2: It's redundant here

Anna-Tkachenko (changes requested):
Good job! 👍 
Just try to fix all tests
Check indents of the list on 1024 px screen (if you delete it - it should fix some tests)
image

src/style.css
```diff
@@ -1 +1,73 @@
...
+.is-active::after {
+  content: "";
+  position: absolute;
+  top: 56px;
+  left: 10px;
+  width: 36px;
+  height: 4px;
+  background: #00acdc;
+  border-radius: 8px;
+}
```
Anna-Tkachenko, line 73: This styles should work correct for each link, if you add class is-active for other link - it won't be correct
Use width: 100% and try to set position using position: relative and bottom properties (without top, left)

Anna-Tkachenko (changes requested):
Fix styles for ::after

Anna-Tkachenko (approved):
👍

PR #1122, opened 2021-12-21
first commit

tt-denis:
- [DEMO LINK](https://tt-denis.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://tt-denis.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,88 @@
 /* Styles go here */
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+}
+
+body {
+  box-sizing: border-box;
+  margin: 0;
+}
+
+.header {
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1 +1,88 @@
 /* Styles go here */
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+}
```
Anna-Tkachenko, line 5: Check font-styles from figma (you should use google fonts to import correct font)

Anna-Tkachenko (changes requested):
Good job! 👍 
Just check font-styles in figma and import correct font using google fonts

Anna-Tkachenko (approved):

PR #1166, opened 2022-01-10
Develop

Hovera-Liubomyr:
https://Hovera-Liubomyr.github.io/layout_moyo-header/
https://Hovera-Liubomyr.github.io/layout_moyo-header/report/html_report/


aliceblunt8 (changes requested):
the mock should be adaptive, the 2 images in the mockup represent how the header should look on different screen sizes and your final solution should consist of one header adapted to fit on 1024px and 1200px the same as on the mock

see the example 
https://volodymyrshymanskyi.github.io/layout_moyo-header/

PR #1196, opened 2022-01-19
Moyo Home work

Ostafii4uk:
- [DEMO LINK](https://Ostafii4uk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Ostafii4uk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,66 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+a {
+  display: block;
+}
+
+.header {
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.img__logo {
+  margin: 10px 0;
+  display: block;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,66 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+a {
```
solaryasha, line 5: Avoid styling tags

src/style.css
```diff
@@ -1 +1,66 @@
...
+.nav__item:last-child {
+  margin-right: 0;
+}
+
+.nav__link {
+  padding: 23px 0;
+  position: relative;
+  text-decoration: none;
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
```
solaryasha, line 42: Better to define font's in `html` tag and inherit in links

solaryasha (approved):

src/style.css
Ostafii4uk: Only selectors, I understood)

Ostafii4uk:
Aproved

PR #1213, opened 2022-01-26
MOYO-HEADER

oleg-zolkiewskyi:


aliceblunt8 (changes requested):
header should take 100% of the screen's width
image

PR #1326, opened 2022-03-28
moyo header is done

ZUBOV-ILLIA:
- [DEMO LINK](https://ZUBOV-ILLIA.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ZUBOV-ILLIA.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,11 +1,10 @@
-body {
-  margin: 0;
+html {
   font-family: Roboto, sans-serif;
-  background: #000;
 }
 
-ul {
+body {
   margin: 0;
+  background: #000;
 }
 
 .header__inner {
@@ -29,6 +28,7 @@ ul {
 }
 
 .nav__inner {
+  margin: 0;
   display: flex;
   justify-content: space-between;
   align-items: center;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1 +1,79 @@
-/* Styles go here */
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
```
Denchepornyuk, line 3: add font-family for html tag

src/style.css
```diff
@@ -1 +1,79 @@
...
+  background: #000;
+}
+
+ul {
```
Denchepornyuk, line 7: don't use tags in css, you have class `nav__inner`

src/style.css
```diff
@@ -1 +1,79 @@
...
+  list-style: none;
+}
+
+.nav__btn {
+  margin: 0 10px;
+}
+
+.nav__btn_first {
+  margin-left: 0;
+}
```
Denchepornyuk, line 46: don't create a separate class for the first element, use CSS pseudo-classes

src/style.css
```diff
@@ -1 +1,79 @@
...
+
+.nav__btn_last {
```
Denchepornyuk, line 48: the same, remove `nav__btn_last` and use CSS pseudo-classes

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1347, opened 2022-03-31
MOYO header is created

enrikadej:


src/index.html
```diff
@@ -7,8 +7,63 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img src="./images/logo.png" alt="logo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+
+          <li class="nav__item item-is-active">
```
Denchepornyuk, line 23: Do you need `item-is-active` class?
You can make underline with `is-active` class

src/style.css
```diff
@@ -1 +1,59 @@
...
+  text-transform: uppercase;
+  transition: color 0.3s;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.item-is-active {
+  position: relative;
```
Denchepornyuk, line 44: add `position: relative;` to is-active and remove `item-is-active` class

Denchepornyuk (changes requested):
Screenshot 2022-03-31 at 18 36 15

add demo link to pull request description

Denchepornyuk (approved):

PR #1354, opened 2022-04-03
First_commit_MOYO

igoree88:
I couldn't  pass the tests. I spent a lot of time trying to solve this problem but unfortunately I still don't understand my mistake (

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,12 +3,51 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1 +1,87 @@
...
+html {
+  font-family: Roboto, sans-serif;
+}
+
+.header {
+  position: sticky;
+  top: 0;
+  right: 0;
+  left: 0;
+  background-color: #fff;
```
Anna-Tkachenko, line 15: It's redundant style, header has white background by default

src/index.html
```diff
@@ -3,12 +3,51 @@
...
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
+        <a class="logo__link" href="#home">logo</a>
```
Anna-Tkachenko, line 17: Will be better to add logo as image and add correct alt attribute to it

src/style.css
```diff
@@ -1 +1,87 @@
...
+  position: absolute;
+  left: 50px;
+  top: 10px;
+  background-image: url(./images/logo.png);
+  background-repeat: no-repeat;
+}
+
+.nav__list {
+  position: relative;
+  left: 377px;
```
Anna-Tkachenko, line 37: What is 377px? Will be better to use display: flex for header and set position of navigation using justify-content property

Anna-Tkachenko (approved):

PR #1421, opened 2022-05-05
Created a page based on the layout

Cherednychenko:
- [DEMO LINK](https://Cherednychenko.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Cherednychenko.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,50 +17,86 @@
     <header>
       <div class="logo">
         <a href="#homePage">
-          <img src="./images/logo.png" alt="website logo">
+          <img
+            src="./images/logo.png"
+            alt="website logo"
+          >
         </a>
       </div>
 
       <nav class="nav">
         <ul class="nav__list">
           <li class="nav__item">
-            <a href="#" class="nav__link
-               is-active"
+            <a
+              href="#"
+              class="nav__link is-active"
             >
               apple
             </a>
           </li>
 
           <li class="nav__item">
-            <a href="#" class="nav__link">samsung</a>
+            <a
+              href="#"
+              class="nav__link"
+            >
+              samsung
+            </a>
           </li>
 
           <li class="nav__item">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -1,14 +1,69 @@
...
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">смартфоны и телефоны</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link"
+               data-qa="hover"
+            >
```
Olexandr6, line 45: [Checklist](https://github.com/mate-academy/layout_html-form/blob/master/checklist.md) 3rd item

src/style.css
```diff
@@ -1 +1,60 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  text-transform: uppercase;
```
Olexandr6, line 4: Don't set this `text-transform` for the whole page; you should set it for a particular element.

src/style.css
```diff
@@ -1 +1,60 @@
...
+  padding-left: 0;
+}
+
+.nav__item {
+  margin-right: 20px;
+}
+
+.nav__item:last-child {
+  margin-right: 0;
+}
```
Olexandr6, line 45: Let's combine this logic and use only one selector? Try to figure out how you can do this)

Olexandr6 (changes requested):
Good solution🔥, but it can be improved👇

src/style.css
Cherednychenko: done
thanks for the idea)

src/style.css
Cherednychenko: done

src/index.html
Cherednychenko: done

Olexandr6 (approved):
Looks good🔥

PR #1461, opened 2022-05-09
header

vova-dzhuryk:
- [DEMO LINK](https://vova-dzhuryk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://vova-dzhuryk.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,75 @@
...
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  height: 60px;
+  position: relative;
```
jstmpelowycz, line 16: Try to avoid using `position: relative;` if not needed

jstmpelowycz: Actually, the only place where you need to put it  – `is-active`, by this allowing `::after` to be properly implemented

src/style.css
```diff
@@ -1 +1,75 @@
...
+
+.logo {
+  position: absolute;
+  left: 50px;
+  top: 10px;
+}
+
+.nav {
+  position: absolute;
+  right: 50px;
```
jstmpelowycz, line 28: You don't need to use `position: absolute;` here, implement it differently

jstmpelowycz: Just use the usual flow of these elements, supplementing them with paddings/margins

src/style.css
jstmpelowycz, line 16: Avoid using `position: relative;` if not needed

jstmpelowycz: Actually, the only place where you should put is `is-active`, by doing so we allow `::after` to be implemented properly

src/style.css
```diff
@@ -1 +1,75 @@
...
+}
```
jstmpelowycz, line 29: You don't need to use `position: absolute;`, try to implement `nav` and `logo` differently

jstmpelowycz: Just use the standard flow of these elements, complementing them with paddings/margins

src/style.css
```diff
@@ -1 +1,75 @@
...
+.link__container {
+  margin-left: 20px;
+}
+
+.link__container:first-child {
+  margin: 0;
+}
+
+.link {
+  display: inline-block;
```
jstmpelowycz, line 49: Link is `inline-block` by default

src/style.css
```diff
@@ -1 +1,75 @@
...
+  position: relative;
```
jstmpelowycz, line 50: The same issue with `position: relative;` here

jstmpelowycz (changes requested):
A few meaningful issues to fix, and you're done!

jstmpelowycz (approved):
Good job 🔥

jstmpelowycz (approved):
Re-approve

PR #1529, opened 2022-06-06
moyo v.01

IvanG6:
https://ivang6.github.io/layout_moyo-header/

https://ivang6.github.io/layout_moyo-header/report/html_report/


src/style.css
```diff
@@ -1 +1,66 @@
...
+
+.nav__link {
+  position: relative;
+  display: inline-block;
+  margin-left: 20px;
+  padding-top: 23px;
+  padding-bottom: 23px;
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
```
marina-tilniak, line 21: it's default value, don't set it

src/style.css
```diff
@@ -1 +1,66 @@
...
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.nav__link.is-active {
+  color: #00acdc;
+}
+
+.nav__link.is-active::after {
```
marina-tilniak, line 37:
```suggestion
.is-active::after {
```

src/style.css
```diff
@@ -1 +1,66 @@
...
+  content: "";
+  position: absolute;
+  border-bottom: 4px solid #00acdc;
+  border-radius: 8px;
+  display: flex;
+  right: 0;
+  margin-top: 19px;
+  left: 0;
+  transition: all 1s;
```
marina-tilniak, line 46: don't use all, use specify animated property instead

src/style.css
marina-tilniak: don't increase specificity, if don't need it

src/style.css
```diff
@@ -1 +1,66 @@
...
+}
+
+.container {
+  display: flex;
+  max-width: calc(100% - 100px);
```
marina-tilniak, line 51: just set paddings, I guess it will fix tests

marina-tilniak (changes requested):

src/style.css
IvanG6: I can't figure out where to fix it.

src/style.css
```diff
@@ -1 +1,68 @@
...
+  border-radius: 8px;
+  display: flex;
+  right: 0;
+  margin-top: 19px;
+  left: 0;
+}
+
+.container {
+  display: flex;
+  max-width: calc(100% - 100px);
```
marina-tilniak, line 53:
```suggestion
.container {
  display: flex;
  max-width: 100% ;
  padding: 0 50px;
```

marina-tilniak (changes requested):

src/style.css
IvanG6: corrected

PR #1535, opened 2022-06-07
Initial commit

Vasyliy22:


vkryvytskyy (changes requested):
Header has a box-shadow in the design, I see no box-shadow here

ArturErmolenko (approved):

PR #1537, opened 2022-06-07
MOYO header

Kras4iy:
- [DEMO LINK](https://Kras4iy.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Kras4iy.github.io/layout_moyo-header/report/html_report/)

Some test (2 of 8) of pixel perfect didnt pass. I cant see difference to understand where my mistake but i think that avg user won't notice any changes :)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,73 @@
 /* Styles go here */
+* {
+  margin: 0;
+}
+
+html {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+}
+
+.header {
+  display: flex;
+  align-items: center;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,74 @@
 /* Styles go here */
+body {
+  margin: 0;
+}
+
+html {
+  margin: 0;
```
VETALEK, line 7:
```suggestion
```
you don't need to reset `margin` on `html` 😉

VETALEK (approved):
Like 👍 

You may add `padding` to header instead of doing separate `margin`s 😉

image

PR #1557, opened 2022-06-08
Develop

PavloMarkov:
[DEMO LINK](https://PavloMarkov.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://PavloMarkov.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,80 @@
...
+.__link {
+  display: block;
+  margin: 0;
+  height: 40px;
+}
+
+.nav__list {
+  margin: 0 5px;
+  padding: 0 5px;
+}
```
VETALEK, line 35:
```suggestion
.nav__item {
  margin: 0 5px;
  padding: 0 5px;
}
```
it is an item, not a list 😉
and your 1st and last elements don't need margins (left and right respectively) 🙃

VETALEK (changes requested):
Good start 😁
commented 👍🏽

PR #1626, opened 2022-07-03
Develop

Nataliia95254:


src/style.css
```diff
@@ -1 +1,65 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+ul {
```
ArturErmolenko, line 5: add styles for classes not tags

src/style.css
```diff
@@ -1 +1,65 @@
...
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
+a {
```
ArturErmolenko, line 11: same

src/style.css
```diff
@@ -1 +1,65 @@
...
+}
+
+.header_logo {
+  padding: 10px 50px;
+
+}
+
+.nav__list {
+  display: flex;
+  height: 60px;
```
ArturErmolenko, line 30:
```suggestion
```
don't hardcode height

src/style.css
```diff
@@ -1 +1,62 @@
...
+  padding-right: 50px;
+}
+
+.header_logo {
+  padding: 10px 50px;
+
+}
+
+.nav__item {
+  font-family: "Roboto", sans-serif;
```
alexandra-protyanova, line 32: move to `html` selector

alexandra-protyanova (approved):

PR #1697, opened 2022-07-14
solution

yevhenii-podplietnii:
[DEMO LINK](https://yevhenii-podplietnii.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://yevhenii-podplietnii.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,74 @@
-/* Styles go here */
+body,
+li {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+ul {
+  margin: 0;
+  padding: 0;
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 50px;
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo {
+  height: 40px;
+}
+
+.nav__list {
+  height: 60px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,74 @@
...
+.logo {
+  height: 40px;
+}
+
+.nav__list {
+  height: 60px;
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  text-align: center;
```
AndriiRut, line 30: Do you need both of them?

src/style.css
```diff
@@ -1 +1,74 @@
...
+  text-transform: uppercase;
+  color: #000;
+}
+
+.list__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  display: block;
```
AndriiRut, line 51: Why do you need to make it block?

AndriiRut (changes requested):
Good job!

src/style.css
yevhenii-podplietnii: Because without block, my :after element cling to bottom of text "Apple"

AndriiRut: Doesn't it work?

AndriiRut (changes requested):

AndriiRut (approved):

PR #1711, opened 2022-07-14
moyoHeader

Alej4ndro1:
- [DEMO LINK](https://Alej4ndro1.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Alej4ndro1.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,32 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
+        <img
+          src="./images/logo.png"
+          alt="moyo"
+        >
+      </div>
+      <nav class="nav">
+        <div class="menu">
+          <ul class="nav__list">
+            <li><a href="#" class="is-active">apple</a></li>
+            <li><a href="#">samsung</a></li>
+            <li><a href="#">смартфоны и телефоны</a></li>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -7,8 +7,32 @@
...
+      <div class="logo">
+        <img
+          src="./images/logo.png"
+          alt="moyo"
+        >
+      </div>
+      <nav class="nav">
+        <div class="menu">
+          <ul class="nav__list">
+            <li><a href="#" class="is-active">apple</a></li>
```
AndriiRut, line 25:
```suggestion
            <li>
              <a href="#" class="is-active">apple</a>
            </li>
```
https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md

src/style.css
```diff
@@ -1 +1,72 @@
...
+
+.header {
+  position: relative;
+  display: flex;
+}
+
+.nav {
+  display: flex;
+  align-items: center;
+  position: absolute;
```
AndriiRut, line 13: why do you need `absolute` here?

src/style.css
```diff
@@ -1 +1,72 @@
...
+.nav__list {
+  display: flex;
+  align-items: center;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+  height: 60px;
+}
+
+.nav__list li {
```
AndriiRut, line 35: don't use tags to implement styles! By doing this you are hang the specificity.

Check the cheklist

src/style.css
```diff
@@ -1 +1,72 @@
-/* Styles go here */
+body {
```
AndriiRut, line 1: reed about [universal selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Universal_selectors)
using this selector you can delete all margin: 0 below it

src/style.css
AndriiRut: `padding: 20 0; `will be enough to center links and make header of right hight

AndriiRut (changes requested):

src/style.css
Alej4ndro1: to position my nav
padding: 20 0; is not working

AndriiRut (approved):

PR #1716, opened 2022-07-14
v.1

SvyatoslavYatsenko:
- [DEMO LINK](https://svyatoslavyatsenko.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://svyatoslavyatsenko.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -7,8 +7,92 @@
...
+              href="#samsung"
+            >
+              samsung
+            </a>
+          </li>
+          <li class="nav__item">
+            <a
+              class="nav__link"
+              href="#mobiles"
+            >
```
AndriiRut, line 48:
```suggestion
            <a class="nav__link" href="#mobiles">
```
DOn't need to start each attribute from the new line if tag has only two attributes.

src/style.css
```diff
@@ -1 +1,66 @@
...
+  margin-left: 0;
+}
+
+.nav__link:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.nav__link:hover:not([data-qa="hover"]):after,
+.is-active:after {
```
AndriiRut, line 57: You don't need to use pseudo-element `after` to all links.
```suggestion
.is-active:after {
```

AndriiRut (changes requested):

src/index.html
SvyatoslavYatsenko: it's much more comfortable for me to read code in this style. Plus, if we add new attributes it will be unnecessary to move the code on a large number of lines. I think so

src/style.css
SvyatoslavYatsenko: thanks

AndriiRut (approved):

PR #1769, opened 2022-07-31
fixed some issues

AlexKinda:
MOYO header

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,40 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a class="logo" href="#home">
+        <img src="./images/logo.png" alt="logo">
+      </a>
+      <nav>
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="is-active" href="#">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a class="link" href="#">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a class="link" href="#">Смартфоны и телефоны</a>
+          </li>
+          <li class="nav__item">
+            <a class="link" href="#" 
+            data-qa="hover"
+            >Ноутбуки и компьютеры</a>
+          </li>
+          <li class="nav__item">
+            <a class="link" href="#">Гаджеты</a>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -9,6 +9,40 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a class="logo" href="#home">
+        <img src="./images/logo.png" alt="logo">
+      </a>
+      <nav>
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="is-active" href="#">Apple</a>
```
VinnieJ-2k20, line 19: Each link should be an anchor link with corresponding address, not just `#`
```suggestion
            <a class="is-active" href="#apple">Apple</a>
```

src/style.css
```diff
@@ -1 +1,76 @@
 /* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@100;400;700&display=swap");
+
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
```
VinnieJ-2k20, line 8: See checklist item 9

src/style.css
```diff
@@ -1 +1,76 @@
...
+
+body {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+}
+
+.header {
+  min-width: 1024px;
+  max-width: 1200px;
```
VinnieJ-2k20, line 17: Header should have full page width on higher resolutions, don't hardcode these values.

src/style.css
```diff
@@ -1 +1,76 @@
...
+
+.is-active {
+  color: #00acdc;
+  text-decoration: none;
+  font-weight: 500px;
+}
+
+.is-active::after {
+  position: relative;
+  top: 18.5px;
```
VinnieJ-2k20, line 51: Don't use decimal px values. Also, don't hardcode such distances, because it will break if header height changes.
Try to find different reference points, for example, bottom side of the element.

VinnieJ-2k20 (changes requested):
You are on the right way, but this task needs some improvements.

1. Fix the width of the header.
2. Font weight on the deployed project is different from design.
3. Blue underline is misplaced
4. Other small comments below.

PR #1801, opened 2022-08-10
Moyo header

artem-musii:
[DEMO LINK](https://artem-musii.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://artem-musii.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,12 +3,56 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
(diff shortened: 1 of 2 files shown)


Denchepornyuk (changes requested):
good job 👍

Screenshot 2022-08-16 at 14 03 56

add transition for hover effect

Denchepornyuk (approved):
Screenshot 2022-08-16 at 16 56 44

don't write a commit message like this
Reading commit message developer should be able to understand main point from commit without checking code

PR #1805, opened 2022-08-11
initial commit

igor-1987:
- [DEMO LINK](https://igor-1987.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://igor-1987.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,75 @@
...
+:root {
+  --primary-font-color: #000;
+  --secondary-font-color: #00acdc;
+  --primary-bg-color: #fff;
+  --secondary-bg-color: #e9e9e9;
+}
+
+* {
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 12: Better to set this to elements that need it, why do we need to set it for every element even if it have no margins or paddings?

src/style.css
```diff
@@ -1 +1,75 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+:root {
+  --primary-font-color: #000;
+  --secondary-font-color: #00acdc;
+  --primary-bg-color: #fff;
+  --secondary-bg-color: #e9e9e9;
+}
```
OleksiiBlezniuk, line 8: 🔥

OleksiiBlezniuk (approved):
Good job!

src/style.css
igor-1987: I guess this guy is guilty: https://www.youtube.com/c/TraversyMedia. 
He did it so frequently in his videos, so I've repeated that as standart setting.
But yes, are right, its time to think much more about meaning and necessity of each selector.

src/style.css
igor-1987: Thank you for review! It was simple but interesting and useful task.

PR #1810, opened 2022-08-12
moyo-header was created

artem-nasadchuk:


src/style.css
```diff
@@ -1 +1,70 @@
 /* Styles go here */
+
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+body {
+  margin: 0;
+  background: #e5e5e5;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
```
Denchepornyuk, line 9: move font styles to html tag

Denchepornyuk (changes requested):
Screenshot 2022-08-16 at 14 03 56

add transition for hover effect

Denchepornyuk (approved):

PR #1829, opened 2022-08-16
created html file with nav header

ArtemHulkoDev:
- [DEMO LINK](https://ArtemHulkoDev.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ArtemHulkoDev.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,59 @@
-/* Styles go here */
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav {
+  padding: 11px 0;
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  padding-left: 0;
+}
+
+.nav__list-item + .nav__list-item {
+  margin-left: 20px;
+}
+
+.nav__list-link {
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+  padding: 23px 0;
+}
+
+.is-active {
+  color: #00acdc;
+  position: relative;
+}
+
+.is-active::before {
+  content: "";
+  position: absolute;
+  height: 4px;
+  width: 37px;
+  background-color: #00acdc;
+  bottom: 0;
+  border-radius: 8px;
+}
+
+.nav__list-link:hover {
+  color: #00acdc;
+}
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Moyo header
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_moyo-header/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_moyo-header/report/html_report/)
+- [DEMO LINK](https://ArtemHulkoDev.github.io/layout_moyo-header/)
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -1 +1,61 @@
...
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 10: I don't thins that body should have this property

src/style.css
```diff
@@ -1 +1,61 @@
-/* Styles go here */
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 9: Body don't have any paddings by default

src/style.css
```diff
@@ -1 +1,61 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav {
+  padding: 11px 0;
+}
```
OleksiiBlezniuk, line 23: It's strange, I don't see any need in it

src/style.css
```diff
@@ -1 +1,61 @@
...
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  padding-left: 0;
+}
```
OleksiiBlezniuk, line 29: What about default margins?

src/style.css
```diff
@@ -1 +1,61 @@
...
+
+.nav__list-item + .nav__list-item {
+  margin-left: 20px;
+}
```
OleksiiBlezniuk, line 33: `:not` and `:first-child` is more readable and understandable

src/style.css
```diff
@@ -1 +1,61 @@
...
+
+.nav__list-link {
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+  padding: 23px 0;
```
OleksiiBlezniuk, line 39: Paddings for inline elements don't works the way we expect, make it block or flex

src/style.css
```diff
@@ -1 +1,61 @@
...
+  transition: all ease 0.3s;
```
OleksiiBlezniuk, line 40: Transition is cool 🔥  But set `all` to it is a bad idea, use only that property that will change

src/style.css
```diff
@@ -1 +1,61 @@
...
+  background-color: #00acdc;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  border-radius: 8px;
+}
+
+.nav__list-link:hover {
+  color: #00acdc;
+}
```
OleksiiBlezniuk, line 61: Would be nice to group your styles, if you have styles for hover it should be near to element class, that will be more readable

src/style.css
OleksiiBlezniuk: I got it, it was `kostyl` because of default ul margins 🙂  Fix it please

OleksiiBlezniuk (changes requested):

src/style.css
ArtemHulkoDev: @OleksiiBlezniuk I don't understand why padding for inline elements doesn't work for this one. Could you explain, please?

OleksiiBlezniuk: It's just a specifics of blocks model, they work different way. For blocks and inline blocks paddings increases element size and increase clickable zone, but for inline elements paddings just increases clickable zone, and can be collapsed

OleksiiBlezniuk (approved):
GJ!

src/style.css
ArtemHulkoDev: @OleksiiBlezniuk understood, thx :)

PR #1832, opened 2022-08-16
moyo must be happy

ihorandrianov:
- [DEMO LINK](https://ihorandrianov.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ihorandrianov.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,25 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header" id="home">
+      <a href="#home" class="logo__link"><img src="./images/logo.png" alt="moyo logo" class="logo__img"></a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__li"><a class="nav__link is-active" href="/">Apple</a></li>
+          <li class="nav__li"><a class="nav__link" href="/">Samsung</a></li>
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,90 @@
...
+
+}
+
+.logo__img {
+  position: absolute;
+  width: 40px;
+  height: 40px;
+  left: 50px;
+  top: 10px;
+}
```
OleksiiBlezniuk, line 85: Well, now I understood why header is relative
As for me it's a strange solution, much easier to use flexbox

src/style.css
```diff
@@ -1 +1,90 @@
...
+
+.nav__li {
+  position: relative;
+  margin: 0 10px;
+}
+
+.nav__li:last-child {
+  margin-right: 0;
+}
```
OleksiiBlezniuk, line 61: And what about first element?

OleksiiBlezniuk (changes requested):

src/index.html
```diff
@@ -7,8 +7,91 @@
...
+              href="/"
+            >
+              Apple
+            </a>
+          </li>
+          <li class="nav__li">
+            <a
+              class="nav__link"
+              href="/"
+            >
```
OleksiiBlezniuk, line 40: You can leave it on one line since there a less the three attributes

src/style.css
```diff
@@ -1 +1,78 @@
...
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  margin: 0;
```
OleksiiBlezniuk, line 13: html don't have margins

src/style.css
```diff
@@ -1 +1,78 @@
...
+.nav {
+  display: flex;
+  flex-direction: row;
+  margin: 0 50px 0 auto;
+}
+
+.logo__img {
+  width: 40px;
+  height: 40px;
+  margin: 10px 0 0 50px;
```
OleksiiBlezniuk, line 77: `align-tems: center` is much better option, what if we will decide to make header higher, or just make links higher? It's better to avoid values hardcoding

OleksiiBlezniuk (approved):

PR #1835, opened 2022-08-16
Solution

anton-iskryk:
[https://anton-iskryk.github.io/layout_moyo-header/](url)
[https://anton-iskryk.github.io/layout_moyo-header/report/html_report/](url)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,68 @@
 /* Styles go here */
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+}
+
+.header {
+  display: flex;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,68 @@
 /* Styles go here */
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
```
Denchepornyuk, line 5: add font style to html instead of body

Denchepornyuk (changes requested):
Screenshot 2022-08-16 at 16 00 41

something wrong with links from PR description

Denchepornyuk (approved):

PR #1845, opened 2022-08-16
added the header

vladimir-dublya:
- [DEMO LINK](https://vladimir-dublya.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://vladimir-dublya.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,118 @@
-/* Styles go here */
+@font-face {
+  font-family: Roboto;
+  font-weight: 500;
+  src: url(fonts/Roboto-Medium.ttf);
+}
```
OleksiiBlezniuk, line 5: You can just import fonts from google-fonts, no need to use @font-face and upload font files to PR

src/style.css
```diff
@@ -1 +1,118 @@
...
+
+.html {
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 8: Are you sure that html should have this property?

src/style.css
```diff
@@ -1 +1,118 @@
...
+}
+
+.body {
+  margin: 0;
+}
+
+.link-reset {
+  color: inherit;
+  text-decoration: none;
+}
```
OleksiiBlezniuk, line 24: Where have you used this class?

src/style.css
```diff
@@ -1 +1,118 @@
...
+
+.lires {
+  margin: 0;
+  padding: 0;
+
+  list-style-type: none;
+  line-height: 1;
+}
```
OleksiiBlezniuk, line 32: No need to create separate class to reset default styles, just set this styles to elements class or apply it to tag
(you can use tags for default styles reset)

src/style.css
```diff
@@ -1 +1,118 @@
...
+
+.header {
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.header__container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
```
OleksiiBlezniuk, line 49: BEM is cool, but you can simplify your styles.
This two classes can be combined, also paddings can be set here, and container can be moved outside header

src/style.css
```diff
@@ -1 +1,118 @@
...
+  text-transform: uppercase;
+  text-align: center;
+  text-decoration: none;
+  transition: color 0.3s;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+
+  transition: color 0.3s;
```
OleksiiBlezniuk, line 90: Transition is cool, but there is no need to set it here, since you've set it to `nav__link` class

src/style.css
```diff
@@ -1 +1,118 @@
...
+}
+
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  opacity: 1;
+
+  transition: opacity 0.3s ease;
```
OleksiiBlezniuk, line 117: Same here with transition

src/style.css
```diff
@@ -1 +1,118 @@
...
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  position: relative;
+
+  display: block;
+  align-items: center;
+  justify-content: center;
```
OleksiiBlezniuk, line 73: Can this rules be applied if your element is just block?

src/style.css
```diff
@@ -1 +1,118 @@
...
+
+  height: 100%;
+  padding: 24px 0;
+
+  color: #000;
+  font-weight: 500;
+  font-size: 12px;
```
OleksiiBlezniuk, line 80: It's a font related styles, they can be set in `html` too

src/style.css
```diff
@@ -1 +1,118 @@
...
+  text-transform: uppercase;
+  text-align: center;
```
OleksiiBlezniuk, line 82: Your link always have same width as it's content, no need to center text

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
GJ, so you just forgot to deploy last changes?

PR #1847, opened 2022-08-16
Added MOYO menu

jojnajbude:
[DEMO LINK](https://jojnajbude.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://jojnajbude.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,49 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img
+        src="images/logo.png"
+        alt="Moyo"
+        class="logo"
+      >
+      <nav class="nav">
+        <ul class="menu">
+          <li class="menu-option">
+            <a href="#" class="menu-option-link  is-active">APPLE</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">SAMSUNG</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">SMARTPHONES</a>
+          </li>
+          <li class="menu-option">
+            <a
+              href="#"
+              class="menu-option-link"
+              data-qa="hover"
+            >LAPTOPS & COMPUTERS</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">GADGETS</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">TABLETS</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">PHOTO</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">VIDEO</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,67 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+.header {
+  width: 100%;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+  margin-left: 50px;
+}
+
+.nav {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  margin-right: 40px;
+}
+
+.menu {
+  display: flex;
+  list-style: none;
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,55 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="index.html" class="image-link">
+        <img
+        src="images/logo.png"
+        alt="Moyo"
+        class="logo"
+      >
```
OleksiiBlezniuk, line 25:
```suggestion
        <img
          src="images/logo.png"
          alt="Moyo"
          class="logo"
        >
```

src/index.html
```diff
@@ -7,8 +7,55 @@
...
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">SMARTPHONES</a>
+          </li>
+          <li class="menu-option">
+            <a
+              href="#"
+              class="menu-option-link"
+              data-qa="hover"
+            >LAPTOPS & COMPUTERS</a>
```
OleksiiBlezniuk, line 43:
```suggestion
            <a
              href="#"
              class="menu-option-link"
              data-qa="hover"
            >
              LAPTOPS & COMPUTERS
            </a>
```

src/index.html
```diff
@@ -7,8 +7,55 @@
...
+      <nav class="nav">
+        <ul class="menu">
+          <li class="menu-option">
+            <a href="#" class="menu-option-link  is-active">APPLE</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">SAMSUNG</a>
+          </li>
+          <li class="menu-option">
+            <a href="#" class="menu-option-link">SMARTPHONES</a>
```
OleksiiBlezniuk, line 36: use text-transform for that purpose

src/style.css
```diff
@@ -1 +1,73 @@
...
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.image-link {
+  display: block;
+  height: 40px;
+  padding: 10px 0;
```
OleksiiBlezniuk, line 20: There is no need for this changes, you've already set align-items to container, so it will align them

OleksiiBlezniuk: Use flex here to avoid this extra space at the bottom of link

src/style.css
```diff
@@ -1 +1,73 @@
...
+  margin-left: 50px;
```
OleksiiBlezniuk, line 21: Strange solution, you can just set horizontal paddings to parent container instead of margins for both childs

src/style.css
```diff
@@ -1 +1,73 @@
...
+}
+
+.nav {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
```
OleksiiBlezniuk, line 26: Use html tag for font styles

src/style.css
```diff
@@ -1 +1,73 @@
...
+  margin-right: 40px;
+}
+
+.menu {
+  display: flex;
+  list-style: none;
+  padding-inline-start: 0;
```
OleksiiBlezniuk, line 33: I think `padding: 0` is enough

src/style.css
```diff
@@ -1 +1,73 @@
...
+  justify-content: flex-start;
```
OleksiiBlezniuk, line 34: I don't think you need this rule

src/style.css
```diff
@@ -1 +1,73 @@
...
+  margin: 0;
+}
+
+.menu-option {
+  display: flex;
+  align-items: center;
+  text-align: center;
+  padding: 0 10px;
```
OleksiiBlezniuk, line 42: Padding to link container will not increase link clickable zone, so I don't see any reason to use padding instead of margin but ok.
But setting padding for both sides of element is strange, especially when you have to calculate parent's margin to include this value.

Set padding/margin to one side and disable it for first/last element, don't make it overcomplicated

src/style.css
```diff
@@ -1 +1,73 @@
...
+}
+
+.menu-option-link {
+  display: block;
+  text-decoration: none;
+  text-transform: uppercase;
```
OleksiiBlezniuk, line 48: okay, in this case your text shouldn't be uppercased in html

src/style.css
```diff
@@ -1 +1,73 @@
...
+  color: black;
+  padding: 23px 0;
+}
+
+.menu-option-link:hover {
+  color: #00acdc;
+
+}
```
OleksiiBlezniuk, line 56:
```suggestion
.menu-option-link:hover {
  color: #00acdc;
}
```

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (changes requested):
Please fix previous comments

src/style.css
```diff
@@ -1 +1,75 @@
...
+  padding: 0;
+  justify-content: center;
+  margin: 0;
+}
+
+.menu-option {
+  display: flex;
+  align-items: center;
+  text-align: center;
+  margin: 0 10px;
```
OleksiiBlezniuk, line 45: Again, this thing break tests, set margins to one side only and disable them for last element
If your container wasn't flexbox, you also could face margin collapse

OleksiiBlezniuk (approved):

PR #1853, opened 2022-08-16
nav was created

varkovroma:


src/style.css
```diff
@@ -1 +1,53 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
```
Andr1yK, line 5: move to html

src/style.css
```diff
@@ -1 +1,53 @@
...
+.container {
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+  background-color: #fff;
+}
+
+.logo {
+  margin-top: 10px;
+}
```
Andr1yK, line 18: IMO, this should not be the case here

src/style.css
```diff
@@ -1 +1,53 @@
...
+
+.nav__list {
+  display: flex;
+  margin: 0;
+}
+
+.nav__item {
+  list-style: none;
```
Andr1yK, line 26: move to nav__list

Andr1yK (changes requested):

Andr1yK (changes requested):
In addition, add hover effect

Andr1yK (changes requested):
Hover doesn't work

src/style.css
Denchepornyuk: test

src/index.html
```diff
@@ -6,9 +6,54 @@
...
+          </a>
+        </div>
+        <nav class="nav">
+          <ul class="nav__list">
+            <li class="nav__item">
+              <a href="#" class="nav__link is-active">apple</a>
+            </li>
+            <li class="nav__item">
+              <a href="#" class="nav__link">samsung</a>
+            </li>
```
Anna-Tkachenko, line 29: Let's specify correct names for links instead of simple # sign
```suggestion
              <a href="#apple" class="nav__link is-active">
                apple
              </a>
            </li>
            <li class="nav__item">
              <a href="#samsung" class="nav__link">
                samsung
              </a>
            </li>
```

src/style.css
```diff
@@ -1 +1,59 @@
...
+  background-color: #e5e5e5;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+  font-style: normal;
+  font-size: 12px;
+  text-transform: uppercase;
+  box-sizing: border-box;
```
Anna-Tkachenko, line 15: Will be better to specify box-sizing property for all elements using * selector, not only for html tag

Anna-Tkachenko (changes requested):
If you'll reset default padding for ul - test should passed
Fix it on the next pr, where you'll need to use Moyo Header
image

Anna-Tkachenko (approved):

PR #1855, opened 2022-08-16
Layout was created

OleksandrGlulhovskiy:
- [DEMO LINK](https://OleksandrGlulhovskiy.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://OleksandrGlulhovskiy.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,67 @@
 /* Styles go here */
+.header {
+  padding: 0 50px;
+}
+
+.container {
+  max-width: 1200px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin: 0 auto;
+}
+
+.logo_pic {
+  width: 40px;
+  height: 40px;
+}
+
+.nav {
+  display: flex;
+  align-items: center;
+  height: 60px;
+}
+
+.nav_list {
+  display: flex;
+}
+
+.nav_items {
+  display: flex;
+  position: relative;
+  align-items: center;
+  font-size: 12px;
+  line-height: 15px;
+  text-transform: uppercase;
+}
+
+.nav_link {
+  color: #000;
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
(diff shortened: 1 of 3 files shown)


OleksiiBlezniuk (changes requested):
It's a third homework were you've just ignored tests initially. You can skip some tests, that is hard to detect where problem is (just to avoid playing with pixels) But in you case there are issues that you can detect, so it's not that case when you can just skip test.

Please, don't ignore tests

src/style.css
```diff
@@ -1 +1,71 @@
...
+.header {
+  padding: 0 50px;
+}
+
+.container {
+  max-width: 1200px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
```
OleksiiBlezniuk, line 15: Now your header is not in center, but it should be

src/style.css
```diff
@@ -1 +1,71 @@
...
+  align-items: center;
+  font-size: 12px;
+  line-height: 15px;
+  text-transform: uppercase;
+}
+
+.nav_link {
+  color: #000;
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
```
OleksiiBlezniuk, line 45: Set font styles to html tag

OleksiiBlezniuk (changes requested):
GJ, but fix UI bug, center your header, it should'n be on the left side of the page

Screenshot 2022-08-17 at 18 28 17

src/style.css
```diff
@@ -1 +1,77 @@
...
+
+.logo_pic {
+  width: 40px;
+  height: 40px;
+}
+
+.nav {
+  display: flex;
+  align-items: center;
+  height: 60px;
```
OleksiiBlezniuk, line 33: Instead of height hardcoding better to reset default ul margins

OleksiiBlezniuk (approved):

PR #1874, opened 2022-08-16
added header

NickBrodyuk:
- [DEMO LINK](https://NickBrodyuk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://NickBrodyuk.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,70 @@
...
+  list-style-type: none;
+  display: flex;
+  justify-content: flex-end;
+  margin: 0;
+}
+
+.nav__link {
+  display: inline-block;
+  margin-right: 20px;
+  padding: 24px 0;
```
OleksiiBlezniuk, line 41: You can set a bit smaller padding here, 23px with right font and font-size should work well here 🙂

src/style.css
```diff
@@ -1 +1,70 @@
...
+
+  text-decoration: none;
+  color: rgb(20, 18, 18);
+  font-weight: bold;
+  text-transform: uppercase;
+}
+
+.is-active {
+  position: relative;
+  color: rgb(105, 210, 210);
```
OleksiiBlezniuk, line 51: Why not `#00ACDC` as it was in mockup?

OleksiiBlezniuk: fix everywhere please

src/style.css
```diff
@@ -1 +1,70 @@
...
+  right: 0;
+
+  height: 4px;
+  background-color: rgb(105, 210, 210);
+  width: 100%;
+}
+
+.nav__link:hover {
+  color: rgb(105, 210, 210);
+  transition: 0.3s;
```
OleksiiBlezniuk, line 69: Transition, great 🔥 
But we usually set transition to element class, not for the state. Now it will not work when you remove cursor from element

src/index.html
```diff
@@ -1,14 +1,89 @@
...
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <!-- я знаю, що не попадаю по пікселям,
+      але вже година ночі - пощади
+     -->
```
OleksiiBlezniuk, line 14: Sorry 😢

OleksiiBlezniuk: Problem is not just in pixels, there were few other issues that need to bi fixed and I believe that this will help to fix some of tests. It's okay if some of tests will be broken because of pixel-perfect, but other issues should be fixed

OleksiiBlezniuk (changes requested):

MaksimusCRS (approved):
GJ

PR #1878, opened 2022-08-17
Develop

La-La-Op:
I have some problem with tests you can see them on the screenshot. Header works as it must according to the task and it is looks as it should.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,99 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class ="header">
+      <a href="#" class ="logo_link">
+        <img
+          src="/images/logo.png"
+          class="logo"
+          alt="logo"
+         >
+      </a>
+      <nav class="link_box">
+        <ul class="link_list">
+          <li
+            class="list_item"
+          >
+            <a
+              href="#"
+              class="link is-active"
+            >
+              apple
+            </a>
+          </li>
+          <li
+            class="list_item"
+          >
+            <a
+              href="#"
+              class="link"
+            >
+              samsung
+            </a>
+          </li>
+          <li
+             class="list_item"
+          >
+            <a
+              href="#"
+              class="link"
+            >
+              smartphones
+            </a>
+          </li>
+          <li
+            class="list_item"
+          >
+            <a
+              href="#"
+              class="link"
+              data-qa="hover"
+            >
+              laptops & computers
+            </a>
+          </li>
+          <li
+            class="list_item"
+          >
+            <a
+              href="#"
+              class="link"
+            >
+              gadgets
+            </a>
+          </li>
+          <li
+            class="list_item"
+          >
+            <a
+              href="#"
+              class="link"
+            >
+              tablets
+            </a>
+          </li>
+          <li
+             class="list_item"
+          >
+            <a
+              href="#"
+              class="link"
+            >
+              photo
+            </a>
+          </li>
+          <li
+            class="list_item"
+          >
+            <a
+              href="#"
+              class="link"
+            >
+              video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -9,6 +9,100 @@
...
+          src="/images/logo.png"
+          class="logo"
+          alt="logo"
+         >
+      </a>
+      <nav class="link_box">
+        <ul class="link_list">
+          <li
+            class="list_item"
+          >
```
OleksiiBlezniuk, line 24: It's overkill, you can leave attribute on one line there were less then 3

src/style.css
```diff
@@ -1 +1,77 @@
...
+  display: flex;
+  align-items: center;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.list_item {
+  margin-left: 20px;
+}
```
OleksiiBlezniuk, line 40: And what about the first element? This thing can break tests

src/style.css
```diff
@@ -1 +1,77 @@
...
+
+.link {
+  text-decoration: none;
+  padding: 25px 0;
+  margin-top: 10px;
```
OleksiiBlezniuk, line 45: Do inline elements have margins? 🙂

src/style.css
```diff
@@ -1 +1,77 @@
...
+  color: black;
+}
+
+.link:hover {
+  transition: 0.5s;
```
OleksiiBlezniuk, line 50: 🔥

OleksiiBlezniuk: But it's better to set transition to element class, not to the hover state, because now it's not working when you remove your cursor from element

src/style.css
```diff
@@ -1 +1,77 @@
...
+  color: #00acdc;
+  position: relative;
+}
+
+.is-active::after {
+  display: block;
+  position: absolute;
+  content: "";
+
+  top: 58px;
```
OleksiiBlezniuk, line 64: But you just need to place it at the bottom, why do you need to calculate this value? Just set bottom: 0 🙂

src/style.css
```diff
@@ -1 +1,77 @@
...
+  height: 4px;
+
+  background: #00acdc;
+  border-radius: 8px;
+}
+
+.logo_link {
+  margin-bottom: 7px;
+  margin-top: 10px;
+}
```
OleksiiBlezniuk, line 77: Looks like `kostyl`

OleksiiBlezniuk: You can just set align-items rule to parent and it will center it horizontally

src/style.css
OleksiiBlezniuk: Also paddings not works as we expect for inline items

OleksiiBlezniuk: Lets make blocks from them, and remove this margin, you don't need it

OleksiiBlezniuk (changes requested):
GJ, but there is some points that can be improved
Also some of them should fix some tests

src/index.html
La-La-Op: done

src/style.css
La-La-Op: Done, and yes, fixing that mistake helped me to pass a few more test without difficulties

src/style.css
La-La-Op: done

src/style.css
La-La-Op: done

src/style.css
La-La-Op: done

src/style.css
La-La-Op: done

src/style.css
```diff
@@ -1 +1,82 @@
...
+  padding: 0;
+}
+
+.list_item {
+  margin-left: 20px;
+}
+
+.first_item_margin {
+  margin-left: 0;
+}
```
OleksiiBlezniuk, line 45: It's overkill, no need to create separate class to disable margins for last element, use pseudoclasses

OleksiiBlezniuk (approved):
Good job! Try to use pseudoclasses instead of additional classes next time 🙂

OleksiiBlezniuk (approved):

PR #1880, opened 2022-08-17
HEADER

VovaDzhuryk:
- [DEMO LINK](https://VovaDzhuryk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://VovaDzhuryk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,13 +2,52 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,57 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+}
+
+body,
+.nav__menu {
+  margin: 0;
+}
```
OleksiiBlezniuk, line 9: You can use tag for styles reset

src/index.html
```diff
@@ -2,13 +2,52 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="logo">
+        <img src="./images/logo.png" alt="logo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__menu">
+          <li class="nav__container">
```
OleksiiBlezniuk, line 21:
```suggestion
          <li class="nav__item">
```

OleksiiBlezniuk: container - it's an element that wraps element, so if it's a nav container - it should wrap nav element. In this case - it's an item

OleksiiBlezniuk (approved):

PR #1888, opened 2022-08-17
Added header moyo

AniaHna:
[DEMO LINK](https://AniaHna.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://AniaHna.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,61 @@
-/* Styles go here */
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 50px;
+  /* height: 60px; */
+  box-sizing: border-box;
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+  position: relative;
+}
+
+.nav__list {
+  text-decoration: none;
+  margin-left: 20px;
+  text-transform: uppercase;
+  font-size: 12px;
+  font-weight: 500;
+  color: #000;
+  font-style: normal;
+  font-family: "Roboto", sans-serif;
+  padding-top: 23px;
+  padding-bottom: 23px;
+  line-height: 60px;
+}
+
+.nav__list:hover {
+  color: rgba(0, 172, 220, 1);
+}
+
+.nav__ul {
+  display: flex;
+  list-style-type: none;
+  justify-content: space-between;
+}
+
+.is-active {
+  position: relative;
+  color: rgba(0, 172, 220, 1);
+}
+
+.is-active::after {
+  content: " ";
+  width: 37px;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,61 @@
...
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 50px;
+  /* height: 60px; */
```
OleksiiBlezniuk, line 11: Remove commented code please

src/style.css
```diff
@@ -1 +1,61 @@
...
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 12: This block don't have any borders, so i think this rule can be removed

src/style.css
```diff
@@ -1 +1,61 @@
...
+
+.nav__list {
+  text-decoration: none;
+  margin-left: 20px;
+  text-transform: uppercase;
+  font-size: 12px;
+  font-weight: 500;
+  color: #000;
+  font-style: normal;
+  font-family: "Roboto", sans-serif;
```
OleksiiBlezniuk, line 25: Set font styles to `html` tag

src/style.css
```diff
@@ -1 +1,61 @@
...
+  padding-top: 23px;
+  padding-bottom: 23px;
```
OleksiiBlezniuk, line 27: Combine paddings to one rule

src/style.css
```diff
@@ -1 +1,61 @@
...
+}
+
+.is-active {
+  position: relative;
+  color: rgba(0, 172, 220, 1);
+}
+
+.is-active::after {
+  content: " ";
+  width: 37px;
```
OleksiiBlezniuk, line 48: And what if other link will be active?

OleksiiBlezniuk: Screenshot 2022-08-17 at 18 33 48

OleksiiBlezniuk (changes requested):

src/style.css
AniaHna: font-size: 12px;
font-weight: 500;

Is this too?

OleksiiBlezniuk: Yep

src/style.css
```diff
@@ -1 +1,60 @@
...
+  background: rgba(0, 172, 220, 1);
+  bottom: 0;
+  position: absolute;
+  display: block;
+  left: 0;
+}
+
+.logo {
+  margin: 10px 0;
+}
```
OleksiiBlezniuk, line 60: I don't think that you really need this margins since you've aligned elements
Also `display: flex` will fix additional bottom space of the link

OleksiiBlezniuk (approved):

PR #1891, opened 2022-08-17
moyo-header was added

Ihor-1993:
- [DEMO LINK](https://Ihor-1993.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Ihor-1993.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,91 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Moyo header</h1>
+
+    <header class="header">
+
+      <div class="shell">
+
+        <div class="header__shell">
+
+          <a href="#" class="logo">
+
+            <img
+              src="images/logo.png"
+              alt="Moyo logo"
+              >
+          </a>
+
+          <nav class="header__nav">
+
+            <ul class="nav">
+
+              <li class="nav__item">
+                <a href="#" class="nav__link is-active">
+                  Apple
+                </a>
+              </li>
+
+              <li class="nav__item">
+                <a href="#" class="nav__link">
+                  Samsung
+                </a>
+              </li>
+
+              <li class="nav__item">
+                <a href="#" class="nav__link">
+                  Smartphones
+                </a>
+              </li>
+
+              <li class="nav__item">
+                <a
+                  href="#"
+                  class="nav__link"
+                  data-qa="hover"
+                >
+                  Laptops &amp; Computers
+                </a>
+              </li>
+
+              <li class="nav__item">
+                <a href="#" class="nav__link">
+                  Gadgets
+                </a>
+              </li>
+
+              <li class="nav__item">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -6,9 +6,91 @@
...
+              src="images/logo.png"
+              alt="Moyo logo"
+              >
+          </a>
+
+          <nav class="header__nav">
+
+            <ul class="nav">
+
+              <li class="nav__item">
```
OleksiiBlezniuk, line 33: There are too much blank lines
It's great to visually separate some big code blocks with blank lines, it's increase readability, but in this case it works opposite 🥲

src/index.html
```diff
@@ -6,9 +6,91 @@
...
+                <a
+                  href="#"
+                  class="nav__link"
+                  data-qa="hover"
+                >
+                  Laptops &amp; Computers
+                </a>
+              </li>
```
OleksiiBlezniuk, line 59: But this one is good 👍

src/index.html
```diff
@@ -6,9 +6,91 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Moyo header</h1>
+
+    <header class="header">
+
+      <div class="shell">
```
OleksiiBlezniuk, line 17: shell it's a strange name, usually we name such div's as `container` or `wrapper` but ok
what is not ok - it's more logic if header container will be outside of header, not inside 🙂

src/style.css
```diff
@@ -1 +1,77 @@
...
+  width: 100%;
+  margin: 0 auto;
+  padding: 0 50px;
+}
+
+.header__shell {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
```
OleksiiBlezniuk, line 30: You can apply this styles to header, no need to create additional divs for such purpose

src/index.html
```diff
@@ -6,9 +6,91 @@
...
+
+          <a href="#" class="logo">
+
+            <img
+              src="images/logo.png"
+              alt="Moyo logo"
+              >
+          </a>
+
+          <nav class="header__nav">
```
OleksiiBlezniuk, line 29: Have you used this class?

src/style.css
```diff
@@ -1 +1,77 @@
...
+
+.logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+}
+
+.nav {
```
OleksiiBlezniuk, line 38:
```suggestion
.nav__list {
```

OleksiiBlezniuk: Better to make names more clear, at first I thought to write comments that navs don't have ant default margins or paddings 🙂

src/style.css
```diff
@@ -1 +1,77 @@
...
+  color: inherit;
+  text-decoration: none;
+  text-transform: uppercase;
+  transition: color 0.5s;
+}
+
+.nav__link.is-active,
+.nav__link:hover {
+  color: #00acdc;
+}
```
OleksiiBlezniuk, line 66: 👍

OleksiiBlezniuk (changes requested):
Good job! There are few comments to fix

src/index.html
Ihor-1993: No, class was delete

src/index.html
Ihor-1993: I tried to wrap header in a container. But then the tests didn't work(

OleksiiBlezniuk (approved):
GJ!

PR #1893, opened 2022-08-18
add header with logo and links, add effects

SerhiiRomaniukWD:
- [DEMO LINK](https://SerhiiRomaniukWD>.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://SerhiiRomaniukWD.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -8,7 +8,70 @@
...
+          >
+        </a>
+      </div>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link is-active"
+            >APPLE</a>
```
Andr1yK, line 28:
```suggestion
            <a
              href="#"
              class="nav__link is-active"
            >
              APPLE
            </a>
```

src/style.css
```diff
@@ -1 +1,90 @@
-/* Styles go here */
+/* add fonts */
```
Andr1yK, line 1:
```suggestion
/* fonts */
```

Andr1yK (changes requested):
Do u need all these files?

image

src/index.html
Anna-Tkachenko, line 28: Fix indents for all links like this:
```suggestion
            <a
              href="#"
              class="nav__link is-active"
            >
              APPLE
            </a>
```

src/style.css
```diff
@@ -1 +1,90 @@
...
+  color: #00acdc;
+  transition: 0.3s;
+}
+
+.is-active::after {
+  display: block;
+  position: absolute;
+
+  content: "";
+  width: 37px;
```
Anna-Tkachenko, line 85: It should works correct for all links, you shouldn't hardcode width value, set 100%

src/style.css
```diff
@@ -1 +1,90 @@
...
+
+/* logo container */
+.logo__container {
+  display: flex;
+}
+
+.logo__link {
+  display: flex;
+
+  padding: 10px 0 10px 50px;
```
Anna-Tkachenko, line 33: Will be better to specify correct right and left paddings for the whole header (they're the same, 50px) instead of adding padding to the logo and to the navigation block

Anna-Tkachenko (changes requested):

Anna-Tkachenko (changes requested):
I don't see any commits after my comments

PR #1904, opened 2022-08-21
added header

Inn0o:
https://Inn0o.github.io/layout_moyo-header/
https://Inn0o.github.io/layout_moyo-header/report/html_report/


src/style.css
```diff
@@ -1 +1,59 @@
-/* Styles go here */
+.header {
+  margin-top: 10px;
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
```
liliia-livandovska, line 6: it's better to move fonts to html {..}

src/style.css
```diff
@@ -1 +1,59 @@
...
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+a {
+  position: relative;
+  color: #000;
+  text-decoration: none;
+  display: inline-block;
```
liliia-livandovska, line 46: its default, remove it

PR #1953, opened 2022-09-09
created moyo header

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/layout_moyo-header/)
[TEST](https://iamgranovskaya.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,71 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+}
```
Virass, line 5:
```suggestion
* {
	box-sizing: border-box;
}

html {
  margin: 0;
  padding: 0;
  font-family: "Roboto", sans-serif;
}
```

Virass (changes requested):

src/style.css
iamgranovskaya: if i move margin and padding into html, it will work only for html tag, it means that body, ul will have defaul indents

src/style.css
```diff
@@ -1 +1,75 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
SanyaBratashchuk, line 4: do not apply any rules to the `*` selector except box-sizing

SanyaBratashchuk (changes requested):

marina-tilniak (approved):

PR #1964, opened 2022-09-14
added header

oirevlis1:
https://oirevlis1.github.io/layout_moyo-header/
https://oirevlis1.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,60 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    >
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home">
+        <img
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,61 @@
...
+  white-space: nowrap;
+  list-style: none;
+}
+
+.nav-item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav-link {
+  font-family: Roboto, sans-serif;
```
TarasHoliuk, line 37: move it to `html`
add one widespread sans-serif font: `font-family: Roboto, Arial, sans-serif;`

src/index.html
```diff
@@ -7,8 +7,60 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home">
+        <img
+        class="header-img"
+        src="images/logo.png"
+        alt="logo MOYO"
+      >
```
TarasHoliuk, line 28:
```suggestion
        <img
          class="header-img"
          src="images/logo.png"
          alt="logo MOYO"
        >
```
better to format in this way :arrow_up:

TarasHoliuk (changes requested):
Very good! Only small changes are needed

src/style.css
oirevlis1: did so, but now the tests fail

TarasHoliuk: a bit weird, but the main thing is that works properly (font-family applied): 
image

of course, it's important to have tests passed but taking into account the above - approved :+1: 
P.S. Just wanted you to know that we should define font-family on the highest level

TarasHoliuk (approved):

PR #1974, opened 2022-09-17
first version

Vladimir-Zadorozhnyi:
https://Vladimir-Zadorozhnyi.github.io/layout_moyo-header/
https://Vladimir-Zadorozhnyi.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,61 @@
-/* Styles go here */
+* {
+  margin: 0;
+}
+
+body {
+  font-family: Roboto, Helvetica, sans-serif;
+  font-weight: 500;
+}
+
+.nav__list {
+  display: flex;
+  width: 100%;
+  background-color: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  list-style: none;
+  justify-content: flex-end;
+  align-items: center;
+  padding: 23px 0;
+}
+
+.nav__link {
+  padding: 0 10px;
+  font-size: 12px;
+}
+
+.link {
+  padding: 23px 0;
+  color: #000;
+  text-decoration: none;
+}
+
+.link:hover {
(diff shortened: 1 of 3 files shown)


SanyaBratashchuk (changes requested):
Header should be flex with justify-content: space-between that's why you can't pass tests

Vladimir-Zadorozhnyi:
> Header should be flex with justify-content: space-between that's why you can't pass tests
I added styles for header, but i have problem with space on the right. I don't know like it delete.

src/style.css
```diff
@@ -1 +1,69 @@
...
+
+.is-active::after {
+  position: absolute;
+  display: block;
+  height: 4px;
+  width: 37px;
+  border-radius: 4px;
+  background-color: #00acdc;
+  top: 56px;
+  left: 0;
```
Bohdan-Maksymiuk, line 57: instead of calculating top try to use bottom: 0

src/style.css
```diff
@@ -1 +1,69 @@
...
+.is-active {
+  position: relative;
+  color: #00acdc;
+}
+
+.is-active::after {
+  position: absolute;
+  display: block;
+  height: 4px;
+  width: 37px;
```
Bohdan-Maksymiuk, line 53: What if it will be link with larger word? Will you calculate it for everyone?
```suggestion
  width: 100%;

```

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-09-20 at 10 29 07

you have box-shadow on ul, but it whould be on the whole header

src/style.css
Vladimir-Zadorozhnyi: I don't know like it work, but it work!

Bohdan-Maksymiuk (approved):

PR #2036, opened 2022-09-29
Completed task

vadymbaranov:
- [DEMO LINK](https://Largoxoxo.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Largoxoxo.github.io/layout_moyo-header/report/html_report/)

HEADS UP
I'm puzzled as to how you set margin-right for the 'VIDEO' element without explicitly using it. Is there a way to center all items inside flexbox with set margins on left and right?


danylo-onishchuk (changes requested):
нет ссылки на демо и тесты

danylo-onishchuk:
Я не согласен

Ссылки тут вовсе не нужны

Largoxoxo:
Links added.

src/style.css
```diff
@@ -1 +1,78 @@
...
+  font-size: 12px;
+  line-height: 14px;
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
```
denys-danyliuk, line 14: don't do it in two places

src/index.html
```diff
@@ -7,8 +7,52 @@
...
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
```
denys-danyliuk, line 30:
```suggestion
          <li class="nav__item">
            <a href="#apple" class="nav__link is-active">Apple</a>
          </li>
          
          <li class="nav__item">
            <a href="#samsung" class="nav__link">Samsung</a>
          </li>
```

src/style.css
```diff
@@ -1 +1,78 @@
...
+.nav__item:last-child {
+  margin: 0;
+}
+
+.nav__link {
+  display: flex;
+  text-decoration: none;
+  color: #000;
+  align-items: center;
+  height: 60px;
```
denys-danyliuk, line 57: don't hardcode height, use paddings and display: inline-block;

denys-danyliuk (approved):

src/style.css
Largoxoxo: Hardcoded height due to my using of display: flex. Using flex prevented from correctly centering items with padding. When paddings where removed and hard height was added (according to figma design) the items correctly centered themselves.
This change was suggested by Illia on my QnA session yesterday. 
Is it better to use display: inline-block with paddings instead of display: flex with hard height in future situations with navigation (similar to this task)?

PR #2053, opened 2022-09-30
Develop

Bohdan-Kovalenko1992:
- [DEMO LINK](https://Bohdan-Kovalenko1992.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Bohdan-Kovalenko1992.github.io/layout_moyo-header/report/html_report/)


mykola-koval (changes requested):
seems like all the tests have failed, let's fix all the margins/paddings that fail the test
🌻

Olexandr6 (approved):
I will approve, but in the catalog task the header should be fixed

PR #2107, opened 2022-10-19
Moyo header was created

Oleksandr-Filo:
- [DEMO LINK](https://Oleksandr-Filo.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Oleksandr-Filo.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -6,9 +6,72 @@
...
+        <img src="images/logo.png" alt="Moyo-logo">
+      </a>
+
+      <nav>
+        <ul class="header__nav">
+          <li class="header__nav-item">
+            <a href="#apple" class="header__nav-link is-active">
+              Apple
+            </a>
+          </li>
```
valerii-derkach, line 26:
```suggestion
        <ul class="header__nav nav">
          <li class="nav__item">
            <a href="#apple" class="nav__link is-active">
              Apple
            </a>
          </li>
```

the navigation can be not only in the header and can be the separate independent block

it is better to use more readable names here

valerii-derkach (approved):

src/index.html
Oleksandr-Filo: Thank you for advice, done.

PR #2215, opened 2022-11-04
start

TetKopot:
[DEMO LINK](https://TetKopot.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://TetKopot.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,70 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+
+}
+
+body {
+  margin: 0;
+}
+
+img {
+  height: 40px;
+  width: 40px;
+}
+
+.nav_item {
+  display: flex;
+  list-style-type: none;
+  margin-right: 20px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,72 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav_list {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  margin: 0;
+  padding-inline-start: 0;
+}
```
valerii-derkach, line 72: styles are applied and overwritten in the order in which they are written, so if we are resetting - we should put nav__list first then nav__item then nav__link
the same with the fonts in nav__link - first the family (in your case you inherit) and then the weight
try to change the order
if you have any questions, feel free and ask in slack

valerii-derkach (changes requested):
order of styles

valerii-derkach (approved):

PR #2247, opened 2022-11-07
header was added

danulyk05:


src/style.css
```diff
@@ -1 +1,59 @@
-/* Styles go here */
+body {
+  font-family: Roboto, sans-serif;
```
stanislav-topikha, line 2: move it to `html` tag

stanislav-topikha: add fallback font

src/style.css
```diff
@@ -1 +1,59 @@
...
+  margin: 0;
+  box-sizing: border-box;
```
stanislav-topikha, line 4: move it to `*` tag

src/style.css
```diff
@@ -1 +1,59 @@
...
+  position: relative;
+  font-family: inherit;
+  color: #000;
+  text-transform: uppercase;
+  padding: 23px 0;
+  margin-left: 20px;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
stanislav-topikha, line 42: use var for repeated values

stanislav-topikha (changes requested):

src/style.css
danulyk05: [CODE KNOWLEDGE] - Don't use * selector for zeroing out your margins or paddings. It's still inefficient for browser to read your web document

stanislav-topikha: u can use * selector only with box-sizing, so lets put box-sizing there

src/style.css
```diff
@@ -1 +1,66 @@
...
+* {
+  box-sizing: border-box;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+html {
+  font-family: Roboto, sans-serif;
+  --color-mockup-blue: #00acdc;
+}
```
stanislav-topikha, line 10:
```suggestion
* {
  box-sizing: border-box;
}

html {
  font-family: Roboto, sans-serif;
  font-size: 12px;
  font-weight: 500;
  
  --color-mockup-blue: #00acdc;
}
```

stanislav-topikha (changes requested):

src/style.css
danulyk05: okey i'm understand thank you

stanislav-topikha (approved):

PR #2259, opened 2022-11-14
header added

RexarFrontEnd:


polosanya (changes requested):
Please add demo and test links to your pull request

src/style.css
```diff
@@ -1 +1,69 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+* {
+  margin: 0;
+  padding: 0;
```
polosanya, line 5: It is better not to use `*` selector(except box-sizing property). If you want to reset default paddings and margins, better to specify for which tags you want to reset them

src/style.css
```diff
@@ -1 +1,69 @@
...
+}
+
+.is-active::after {
+  position: absolute;
+  content: "";
+  height: 4px;
+  width: 100%;
+  background-color: #00acdc;
+  left: 0;
+  top: 56px;
```
polosanya, line 67: Please pay attention to your CSS properties order
Please fix all cases

polosanya (changes requested):
Almost perfect! 
Please check the comments

src/style.css
```diff
@@ -1 +1,71 @@
...
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  background-color: #fff;
+  margin-bottom: 74px;
```
loralevitska, line 24: for what?

image

src/style.css
```diff
@@ -1 +1,71 @@
...
+body {
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  background: #e5e5e5;
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
```
loralevitska, line 19: it's redundant default value

loralevitska (changes requested):
almost done! need some improvements

loralevitska (approved):
great!

PR #2306, opened 2022-11-25
Develop

MayerLansky:
[DEMO LINK](https://MayerLansky.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://MayerLansky.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,22 +17,18 @@
           alt="moyo"
         >
       </a>
-      <nav class="nav_list">
-        <label>
-          <ul>
-            <li class="links"><a href="#" class="is-active">APPLE</a></li>
-            <li class="links"><a href="#">SAMSUNG</a></li>
-            <li class="links"><a href="#">SMARTPHONES</a></li>
-            <li class="links" data-qa="hover"><a href="#">LAPTOPS & COMPUTERS</a></li>   <!-- data hover, is it right?-->
-            <li class="links"><a href="#">GADGETS</a></li>
-            <li class="links"><a href="#">TABLETS</a></li>
-            <li class="links"><a href="#">PHOTO</a></li>
-            <li class="links"><a href="#">VIDEO</a></li>
-          </ul>
-        </label>
+      <nav>
+        <ul class="nav_list">
+          <li class="links"><a href="#apple" class="link_a is-active">APPLE</a></li>
+          <li class="links"><a href="#samsung" class="link_a">SAMSUNG</a></li>
+          <li class="links"><a href="#smartphones" class="link_a">SMARTPHONES</a></li>
+          <li class="links" data-qa="hover"><a href="#laptops" class="link_a">LAPTOPS & COMPUTERS</a></li>
+          <li class="links"><a href="#gadgets" class="link_a">GADGETS</a></li>
+          <li class="links"><a href="#tablets" class="link_a">TABLETS</a></li>
+          <li class="links"><a href="#photo" class="link_a">PHOTO</a></li>
+          <li class="links"><a href="#video" class="link_a">VIDEO</a></li>
+        </ul>
       </nav>   
-      
     </header>
-    
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -2,13 +2,37 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <a href="#home">
+        <img 
+          src="./images/logo.png" 
+          class="logo" 
+          alt="moyo"
+        >
+      </a>
+      <nav class="nav_list">
+        <label>
+          <ul>
+            <li class="links"><a href="#" class="is-active">APPLE</a></li>
+            <li class="links"><a href="#">SAMSUNG</a></li>
+            <li class="links"><a href="#">SMARTPHONES</a></li>
+            <li class="links" data-qa="hover"><a href="#">LAPTOPS & COMPUTERS</a></li>   <!-- data hover, is it right?-->
+            <li class="links"><a href="#">GADGETS</a></li>
+            <li class="links"><a href="#">TABLETS</a></li>
+            <li class="links"><a href="#">PHOTO</a></li>
+            <li class="links"><a href="#">VIDEO</a></li>
+          </ul>
+        </label>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -2,13 +2,33 @@
...
+      <nav>
+        <ul class="nav_list">
+          <li class="links"><a href="#apple" class="link_a is-active">APPLE</a></li>
+          <li class="links"><a href="#samsung" class="link_a">SAMSUNG</a></li>
+          <li class="links"><a href="#smartphones" class="link_a">SMARTPHONES</a></li>
+          <li class="links" data-qa="hover"><a href="#laptops" class="link_a">LAPTOPS & COMPUTERS</a></li>
+          <li class="links"><a href="#gadgets" class="link_a">GADGETS</a></li>
+          <li class="links"><a href="#tablets" class="link_a">TABLETS</a></li>
+          <li class="links"><a href="#photo" class="link_a">PHOTO</a></li>
+          <li class="links"><a href="#video" class="link_a">VIDEO</a></li>
```
TarasHoliuk, line 29: You've set `uppercase` - that's the correct way to do it but in this case, no need to do it in HTML
```suggestion
          <li class="links"><a href="#video" class="link_a">Video</a></li>
```

src/style.css
```diff
@@ -1 +1,90 @@
...
+.links:last-child {
+  margin-right: 0;
+}
+
+.link_a {
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+  cursor: pointer;
+  font-family: inherit;
```
TarasHoliuk, line 54: it inherits by default - redundant

src/style.css
```diff
@@ -1 +1,90 @@
...
+  display: block; /*??????????*/
```
TarasHoliuk, line 55: redundant

src/style.css
```diff
@@ -1 +1,90 @@
...
+
+.links:last-child {
+  margin-right: 0;
+}
+
+.link_a {
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+  cursor: pointer;
```
TarasHoliuk, line 53: default style for all links - redundant

src/style.css
```diff
@@ -1 +1,90 @@
...
+.link_a:active::after {
+  content: "";
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  height: 4px;
+  width: 100%;
+  background-color: #00acdc;
+  /*border-radius: 8px;*/
+}
```
TarasHoliuk, line 90: I understand the reason you've added it, but you could delete it here since we can't select some tab permanently for now

src/style.css
```diff
@@ -1 +1,90 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+}
+
+/*test shows 6-2. please help to realize my mistakes*/
```
TarasHoliuk, line 5: I don't see the reason to test fails, try to fix everything I've suggested

It looks as expected in the demo, so if after your changes test still fails - it's not a big problem in this case

TarasHoliuk (changes requested):

src/index.html
```diff
@@ -2,13 +2,33 @@
...
+      <a href="#home">
+        <img 
+          src="./images/logo.png" 
+          class="logo" 
+          alt="moyo"
+        >
+      </a>
+      <nav>
+        <ul class="nav_list">
+          <li class="links"><a href="#apple" class="link_a is-active">Apple</a></li>
```
SanyaBratashchuk, line 22:
```suggestion
          <li class="links">
            <a href="#apple" class="link_a is-active">
              Apple
            </a>
          </li>
```
Fix it everywhere

src/style.css
```diff
@@ -1 +1,87 @@
...
+.nav_list {
+  display: flex;
+  text-align: center;
+  margin: 0;
+  padding: 0;
+}
+
+.links {
+  margin-right: 20px;
+  padding: 23px 0;
```
SanyaBratashchuk, line 36: Set this paddings to the link instead of `li`

src/style.css
```diff
@@ -1 +1,87 @@
...
+}
+
+.links:last-child {
+  margin-right: 0;
+}
+
+.link_a {
+  text-decoration: none;
+  text-transform: uppercase;
+  cursor: pointer;
```
SanyaBratashchuk, line 50:
```suggestion
```

src/style.css
```diff
@@ -1 +1,87 @@
...
+.link_a:active::after {
+  content: "";
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  height: 4px;
+  width: 100%;
+  background-color: #00acdc;
+  border-radius: 8px;
+}
```
SanyaBratashchuk, line 87: combine these selectors

SanyaBratashchuk (changes requested):

SanyaBratashchuk (changes requested):
There is no changes since last review. Push your changes

src/style.css
MayerLansky: the thing is that after combining these two elements, the first link Apple goes out from the nav, namely absolute position is getting wrong.

src/style.css
MayerLansky: I did it earlier, however, in this case all the paddings worked wrongly, they did not affect the links at all

src/style.css
```diff
@@ -1 +1,86 @@
...
+
+.links:last-child {
+  margin-right: 0;
+}
+
+.link_a {
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+  font-family: inherit;
```
SanyaBratashchuk, line 51:
```suggestion
```

src/style.css
```diff
@@ -1 +1,86 @@
...
+  padding: 0;
+}
+
+.links {
+  margin-right: 20px;
+  position: relative;
+  list-style: none;
+  justify-content: center;
+  text-align: center;
+  padding: 23px 0;
```
SanyaBratashchuk, line 40: set this paddings to the `a`

SanyaBratashchuk (changes requested):

src/style.css
MayerLansky: as i mentioned previously, when i set it to the a , all paddings had been worked wrongly

src/index.html
```diff
@@ -2,13 +2,49 @@
...
+          class="logo" 
+          alt="moyo"
+        >
+      </a>
+      <nav>
+        <ul class="nav_list">
+          <li class="links">
+            <a href="#apple" class="link_a is-active">Apple</a>
+          </li>
+          <li class="links">
```
DarkMistyRoom, line 25: Add empty lines between multiline sibling blocks of HTML

src/style.css
DarkMistyRoom: why? they should work correctly, if you do everything in the right way

DarkMistyRoom: and if you have questions, ask in the chat

DarkMistyRoom (changes requested):
Try to set the paddings to `a`, not to `li `to pass the tests and to satisfy this condition
image

src/style.css
```diff
@@ -1 +1,85 @@
...
+  position: absolute;
+  height: 4px;
+  left: 0;
+  bottom: 0;
+  width: 100%;
+  background: #00acdc;
+  border-radius: 8px;
+}
+
+.link_a:active::after {
```
AlisaVlasova, line 76: redundant

AlisaVlasova (approved):

PR #2309, opened 2022-11-26
added moyo- header

alla-savaryn:


polosanya (changes requested):
Please add demo and test links to your Pull request description

alla-savaryn:
[DEMO LINK](https://alla-savaryn.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://alla-savaryn.github.io/layout_moyo-header/report/html_report/)

polosanya (changes requested):
Please try to fix the problems to pass all the tests. Compare the reference and test screens.
Feel free to ask for help in fe_chat if you have any difficulties)
image

src/style.css
```diff
@@ -1 +1,74 @@
...
+  margin-right: 0;
+}
+
+.nav__links:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+  height: 15px;
```
loralevitska, line 62:
```suggestion
  height: 15px;
  position: relative
```

src/style.css
```diff
@@ -1 +1,74 @@
...
+}
+
+.is-active::after {
+  display: block;
+  content: "";
+  width: 37px;
+  height: 4px;
+  position: relative;
```
loralevitska, line 70:
```suggestion
  position: absolute;
```

loralevitska (changes requested):
almost done!

1. don't use paddings, use flex instead. Don't use left and right side margins, instead use only left or right (except first or last child)
something like that 
image
I guess it helps you to fix 2 failed tests

src/style.css
```diff
@@ -1 +1,68 @@
...
+}
+
+.logo {
+  display: flex;
+}
+
+.nav__list {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
```
TarasHoliuk, line 25: redundant

src/style.css
```diff
@@ -1 +1,68 @@
...
+  box-shadow: 1px 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo {
+  display: flex;
+}
+
+.nav__list {
+  display: flex;
+  flex-direction: row;
```
TarasHoliuk, line 24: it's a default value, no need to define it

TarasHoliuk (approved):
GJ! Approved :heavy_check_mark: 
I've added 2 small comments just FYI. So no need to make additional changes - the task is approved.

PR #2311, opened 2022-11-26
Develop

GUSILLUS:
removed vertical padding and used align-items:centre to element


OlEzhA0:
Hey! Where can I find links on gh-pages and tests report?

src/index.html
```diff
@@ -7,8 +7,88 @@
...
+          src="images\logo.png" 
+          alt="logo"
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a 
+              href="#apple" 
+              class="nav__item__link is-active"
```
OlEzhA0, line 28: would be better to use `is-active` class in `li`

OlEzhA0, line 28: use just `nav__link`
here is example how name classed with BEM convention

`block-name__elem-name_mod-name_mod-val`

src/style.css
```diff
@@ -1 +1,84 @@
...
+body,
+header,
+a,
+img,
+div,
+nav,
+ul,
+li {
+  margin: 0;
+}
```
OlEzhA0, line 10: it's redundant, why we need to reset margins for `header` or `a` for example? not all this tags have margins by default

src/style.css
```diff
@@ -1 +1,84 @@
...
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  text-transform: uppercase;
```
OlEzhA0, line 16: there is no necessary to use styles like this globally
use it directly to item

src/style.css
```diff
@@ -1 +1,84 @@
...
+  line-height: 14px;
+}
+
+.header {
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  box-sizing: border-box;
```
OlEzhA0, line 26: let's reset this style globally

src/style.css
```diff
@@ -1 +1,84 @@
...
+  align-items: center;
+
+}
+
+.nav {
+  display: flex;
+  justify-content: space-between;
```
OlEzhA0, line 33: why do we need `space-between` here?

src/style.css
```diff
@@ -1 +1,84 @@
...
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  padding: 0;
+  justify-content: flex-ends;
```
OlEzhA0, line 40: there is no value in css like `flex-ends`

src/style.css
```diff
@@ -1 +1,84 @@
...
+}
+
+.is-active {
+  color: #00acdc;
+  position: relative;
+}
+
+.nav__item .is-active {
+  color: #00acdc;
+}
```
OlEzhA0, line 74: why do we need this? you added color above

src/style.css
```diff
@@ -1 +1,84 @@
...
+.nav__list {
+  display: flex;
+  list-style: none;
+  padding: 0;
+  justify-content: flex-ends;
+}
+
+.nav__item__link {
+  align-items: center;
+  position: relative;
```
OlEzhA0, line 45: why do we need `position: relative` here?

src/style.css
```diff
@@ -1 +1,84 @@
...
+}
+
+.nav__item .is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  width: 37px;
```
OlEzhA0, line 78: don't define width in px, because when you move `is-active` class to another link, you'll see incorrect result
Screenshot 2022-11-28 at 12 42 01 PM

OlEzhA0 (changes requested):
Hey! please check my comments and ping me in fe_chat if you have any questions

src/style.css
GUSILLUS: image
So what i need to do?

src/index.html
GUSILLUS: this is the test condition

GUSILLUS:
- [DEMO LINK](https://GUSILLUS.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://GUSILLUS.github.io/layout_moyo-header/report/html_report/)

witflash:
You have some unresolved questions. Please check them and add changes or give answers.

Virass (changes requested):
Agree, recheck previous comments

src/style.css
```diff
@@ -1 +1,69 @@
...
+
+.logo {
+  display: inline-flex;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.nav__item:last-of-type .nav__link {
```
witflash, line 51: You may try to find a better solution here.
For example: `.nav__link .nav__link--last`

witflash (approved):
Cool!

PR #2312, opened 2022-11-26
Header was added

anna-inozemets:
https://anna-inozemets.github.io/layout_moyo-header/
https://anna-inozemets.github.io/layout_moyo-header/report/html_report/


ArturErmolenko (changes requested):
Good job 👍 
add box-shadow as on mockup

PR #2317, opened 2022-11-28
header was added

Nikita211112:
https://Nikita211112.github.io/layout_moyo-header/

https://Nikita211112.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,67 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,67 @@
...
+  margin: 0;
+}
+
+.header__item {
+  font-size: 12px;
+  line-height: 1;
+  margin: 24px 0;
+}
+
+.header__item + .header__item {
```
Bohdan-Maksymiuk, line 41: do you need to increase specificity here?

Bohdan-Maksymiuk (approved):

src/style.css
Nikita211112: > do you need to increase specificity here?

I don't need him here, you're right

Nikita211112: > do you need to increase specificity here?

No, but using this method, I indented to the left.Because of this way, the first element ignores the indentation

PR #2352, opened 2022-12-06
finished the task with header

andrii-kurash:
https://andrii-kurash.github.io/layout_moyo-header/

:)


ArturErmolenko (changes requested):
Good job 👍 
add box-shadow as on mockup

PR #2361, opened 2022-12-06
add moyo-header

Dashaverz:
https://Dashaverz.github.io/layout_moyo-header/
https://Dashaverz.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,68 @@
 /* Styles go here */
+* {
+  box-sizing: border-box;
+}
+
+body {
+  color: #000;
+  margin: 0;
+  font-size: 12px;
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo_link {
+  display: flex;
+}
+
+.nav__link {
+  color: inherit;
+  text-decoration: none;
+  text-transform: uppercase;
+  transition: 0.3s color;
+  padding: inherit;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,68 @@
 /* Styles go here */
+* {
+  box-sizing: border-box;
+}
+
+body {
```
Anna-Tkachenko, line 6: will be better to set these styles using separate class instead of using tag selector

src/style.css
```diff
@@ -1 +1,68 @@
...
+  color: #000;
+  margin: 0;
+  font-size: 12px;
+  font-family: "Roboto", sans-serif;
```
Anna-Tkachenko, line 10: move font styles to html tag

src/style.css
```diff
@@ -1 +1,68 @@
...
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.nav__list li {
```
Anna-Tkachenko, line 45: don't use tag selector for setting styles

Anna-Tkachenko (changes requested):

src/style.css
```diff
@@ -1 +1,72 @@
...
+html {
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+}
+
+.body__style {
+  color: #000;
+  margin: 0;
+  font-size: 12px;
+
```
Anna-Tkachenko, line 15: delete redundant empty line

src/style.css
```diff
@@ -1 +1,72 @@
...
+.nav__li {
+  padding: 23px 0;
+  position: relative;
+  margin-right: 20px;
+}
+
+.nav__list li:last-child {
+  margin-right: 0;
+}
```
Anna-Tkachenko, line 57: don't use tag selectors for setting classes

Anna-Tkachenko (approved):
Good job! 🔥

PR #2375, opened 2022-12-07
finaly finish task

alSemenyuk:
- [DEMO LINK](https://alSemenyuk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://alSemenyuk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,7 +1,10 @@
+html {
+  font-family: Roboto, sans-serif;
+}
+
 body {
   margin: 0;
   padding: 0;
-  font-family: Roboto, sans-serif;
 }
 
 .header {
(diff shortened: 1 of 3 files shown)


Anna-Tkachenko (changes requested):
the image should has 40*40 px size and 10px padding to the top and the bottom of the header
try to set it and re-run tests
image

src/style.css
```diff
@@ -1 +1,72 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
Anna-Tkachenko, line 7: do you need to reset paddings for the body tag?

Anna-Tkachenko (approved):
Good job! 🔥

PR #2389, opened 2022-12-07
add moyo header

artem-sobolevskyi:
- [DEMO LINK](https://artem-sobolevskyi.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://artem-sobolevskyi.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,69 @@
 /* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  background: #fff;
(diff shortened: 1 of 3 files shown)


Anna-Tkachenko (changes requested):
reset margin for the first link and try to pass tests one more time
image

src/style.css
```diff
@@ -1 +1,73 @@
 /* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: "Roboto", sans-serif;
```
Anna-Tkachenko, line 8: will be better to move font styles to html tag

src/style.css
```diff
@@ -1 +1,73 @@
 /* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
+body {
+  margin: 0;
+  padding: 0;
```
Anna-Tkachenko, line 6: do you need to reset paddings for the body?

Anna-Tkachenko (approved):
Good job! 🔥

PR #2391, opened 2022-12-08
moyo header

Tsvylenyeva:
- [DEMO LINK](https://Tsvylenyeva.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Tsvylenyeva.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,80 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img
+        src="/images/logo.png"
+        alt="logo"
+        class="image"
+      >
+      <nav class="nav">
+        <ul class="ul">
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            apple
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            samsung
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            smartfones
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            laptops & computers
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            gadgets
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+          tablets
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            photo
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            video
+            </a>
+          </li>
+        </ul>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -9,6 +9,80 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img
```
Anna-Tkachenko, line 13: logo should be the link, check point 8 in checklist

src/index.html
```diff
@@ -9,6 +9,80 @@
...
+        class="image"
+      >
+      <nav class="nav">
+        <ul class="ul">
+          <li class="li">
+            <a
+              href="#"
+              class="a"
+            >
+            apple
```
Anna-Tkachenko, line 25: check indents for text inside all links

src/index.html
```diff
@@ -9,6 +9,80 @@
...
+            </a>
+          </li>
+          <li class="li">
+            <a
+              href="#"
```
Anna-Tkachenko, line 30: add corresponding values for href attribute for all links
```suggestion
              href="#samsung"
```

src/style.css
```diff
@@ -1 +1,46 @@
 /* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap");
+
+.header {
+  width: 100%;
+  height: 60px;
```
Anna-Tkachenko, line 6: don't set exact height for header, it should be calculated according to the content
just set correct width and height for the logo, and set correct top and bottom paddings for the header, so the full height will be also 60px

src/style.css
```diff
@@ -1 +1,46 @@
...
+  height: 100%;
+  display: block;
+  font-style: normal;
+  font-size: 12px;
+  text-transform: uppercase;
+  font-weight: 500px;
+  padding-left: 20px;
+}
+
+.a {
```
Anna-Tkachenko, line 34: check class names, you shouldn't use tag names as class names, it should be meaningful and understandable
also, I don't see styles for active link, check the task description one more time

Anna-Tkachenko (changes requested):
delete `backstop data` folder and try to run tests one more time

src/style.css
Tsvylenyeva: I'll check everything and make it better, thanks for commenting :)

src/index.html
```diff
@@ -6,9 +6,92 @@
...
+            gadgets
+            </a>
+          </li>
+          <li class="nav__link">
+            <a
+              href="#tablets"
+              class="nav__link"
+            >
+          tablets
+            </a>
```
Denchepornyuk, line 75: you didn't fix indents

src/style.css
```diff
@@ -1 +1,45 @@
 /* Styles go here */
+
+body {
+  margin: 0;
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
```
Denchepornyuk, line 5: it's better to add `font-family` to html tag

Denchepornyuk (approved):

PR #2395, opened 2022-12-08
MOYO-HEADER

Kardash-Oleksandr:
- [DEMO LINK](https://Kardash-Oleksandr.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Kardash-Oleksandr.github.io/layout_moyo-header/report/html_report/)


Anna-Tkachenko:
Demo link doesn't work, please, fix it 🧚‍♀️

src/index.html
```diff
@@ -7,8 +7,57 @@
...
+        <img
+          src="./images/logo.png"
+          alt="moyo_logo"
+          class="logo__icon"
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#" class="nav__link is-active">APPLE</a>
```
Anna-Tkachenko, line 30: Add corresponding values for href attributes for each link
```suggestion
            <a href="#apple" class="nav__link is-active">APPLE</a>
```

src/index.html
```diff
@@ -7,8 +7,57 @@
...
+          </li>
+          <li class="nav__item">
+            <a href="#" class="nav__link">SAMSUNG</a>
```
Anna-Tkachenko, line 33: You can change register via css rule

src/style.css
```diff
@@ -1 +1,67 @@
-/* Styles go here */
+body {
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", Arial, sans-serif;
```
Anna-Tkachenko, line 4: move font styles to html tag

src/style.css
```diff
@@ -1 +1,67 @@
-/* Styles go here */
+body {
+  margin: 0;
+  padding: 0;
```
Anna-Tkachenko, line 3: why did you reset paddings for body? looks like body doesn't have the default paddings
image

src/style.css
```diff
@@ -1 +1,67 @@
...
+  font-family: "Roboto", Arial, sans-serif;
+}
+
+.header {
+  display: flex;
+  flex-wrap: wrap;
```
Anna-Tkachenko, line 9: this property is redundant, delete it
for what reason did you add this property? probably you don't need it

src/style.css
```diff
@@ -1 +1,67 @@
...
+
+.link__logo {
+  height: 40px;
+  width: 40px;
+}
+
+.logo__icon {
+  height: 100%;
+  width: 100%;
+}
```
Anna-Tkachenko, line 25: looks like you added height and width for inline elements, check it

src/style.css
```diff
@@ -1 +1,67 @@
...
+  align-items: center;
+  padding: 23px 0;
+  position: relative;
+}
+
+.nav__link.is-active {
+  color: #00acdc;
+}
+
+.nav__link.is-active::after {
```
Anna-Tkachenko, line 54: increasing selector specificity is redundant here, `.nav__link` is redundant here
```suggestion
.is-active {
  color: #00acdc;
}

.is-active::after {
```

Anna-Tkachenko (changes requested):
Good job! Just left you some comments! ⭐️🔥

src/index.html
Anna-Tkachenko: I think this comment is not valid, because we don't have blocks with corresponding ids right now

Anna-Tkachenko: @Anna-Tkachenko hello

src/style.css
Kardash-Oleksandr: @Anna-Tkachenko Thanks

ArturErmolenko (approved):

PR #2408, opened 2022-12-12
Develop

vladis9:


src/style.css
```diff
@@ -1 +1,61 @@
-/* Styles go here */
+* {
```
AlisaVlasova, line 1: use * only for box-sizing, move font props to html tag and spacing you can reset for required tags directly

src/style.css
```diff
@@ -1 +1,61 @@
...
+  align-items: center;
+}
+
+.nav__list:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav__link {
+  display: flex;
+  height: 60px;
```
AlisaVlasova, line 36: it`s better to use vertical paddings instead of height, it will prevent content overflow

AlisaVlasova (changes requested):

src/style.css
vladis9: understand)

src/style.css
vladis9: check

src/style.css
```diff
@@ -1 +1,70 @@
...
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  font-style: normal;
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
tetiana-ivanchenko, line 10: body hasn't default paddings

src/style.css
```diff
@@ -1 +1,70 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
```
tetiana-ivanchenko, line 2: add fallback font

src/style.css
```diff
@@ -1 +1,70 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.nav__items {
+  margin: 0;
```
tetiana-ivanchenko, line 33: you do it twice: for tag and for class

src/style.css
```diff
@@ -1 +1,70 @@
...
+  padding: 23px 0;
+  flex-direction: column;
+  justify-content: center;
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
tetiana-ivanchenko, line 54: you can move repeatable colors into variable

tetiana-ivanchenko (changes requested):

src/style.css
vladis9: what that mean? in figma only one font

vlad-demchuk (approved):

PR #2425, opened 2022-12-15
moyo-header var inserted

postrygan:
https://postrygan.github.io/layout_moyo-header/
https://postrygan.github.io/layout_moyo-header/report/html_report/


AlisaVlasova (changes requested):
good job!)

PR #2441, opened 2022-12-20
Flexbox / MOYO Header 

valntyn:
[DEMO LINK](https://valntyn.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://valntyn.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,66 @@
-/* Styles go here */
+html {
+  font-family: Roboto, Arial, sans-serif;
+  --default-color: #00acdc;
+}
+
+body {
+  margin: 0;
+  background-color: #e5e5e5;
+}
+
+.header {
+  display: flex;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,66 @@
...
+
+.nav_list {
+  list-style: none;
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.nav_item {
+  display: flex;
```
witflash, line 31: Actually, you don't have a need to use `flex` here.
You may use `display: block`

witflash (approved):
Well done!
Look at my comment. It's not critical but will be good for your future work.

src/style.css
valntyn: thank you for the correction
I will keep it in mind

PR #2444, opened 2022-12-21
commit1

edkido:
[DEMO LINK](https://edkido.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://edkido.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,7 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+}
+
 html {
   font-family: Roboto, Arial, Helvetica, sans-serif;
 }
@@ -15,10 +18,8 @@ body {
 
 .header {
   display: flex;
-  box-sizing: border-box;
   justify-content: space-between;
   align-items: center;
-  height: 60px;
   line-height: 14px;
   padding: 0 50px;
   background: #fff;
@@ -45,10 +46,7 @@ body {
   color: var(--regular-font);
   cursor: pointer;
   padding: 23px 0;
-  flex-direction: column;
-  justify-content: center;
   text-transform: uppercase;
-  text-align: center;
 }
 
 .nav__link:hover {


src/style.css
```diff
@@ -1 +1,71 @@
 /* Styles go here */
```
OlEzhA0, line 1: remove commented code

src/style.css
```diff
@@ -1 +1,71 @@
...
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.header {
+  display: flex;
+  box-sizing: border-box;
```
OlEzhA0, line 18: move `box-sizing: border-box;` to `*` selector

src/style.css
```diff
@@ -1 +1,71 @@
...
+  justify-content: space-between;
+  align-items: center;
+  height: 60px;
```
OlEzhA0, line 21: don't hardcore height here

src/style.css
```diff
@@ -1 +1,71 @@
...
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  display: flex;
+  text-decoration: none;
+  color: var(--regular-font);
+  cursor: pointer;
```
OlEzhA0, line 46: it's default value for `a`

src/style.css
```diff
@@ -1 +1,71 @@
...
+  padding: 23px 0;
+  flex-direction: column;
+  justify-content: center;
```
OlEzhA0, line 49: why do we need this?

src/style.css
```diff
@@ -1 +1,71 @@
...
+  text-transform: uppercase;
+  text-align: center;
```
OlEzhA0, line 51: the same
Why do we need text align here?

OlEzhA0 (changes requested):
Hey! Great job
please pay attention to small comments

src/style.css
edkido: I see, thanks for your review. I had to add it because wasn't sure if it's default in all browsers. But now i checked info about compatibility and i know there is no need in writing default value.

src/style.css
edkido: was trying different ways to solve, and forgot to remove the rest

src/style.css
edkido: on my journey to solve this task - forgot to remove useless attributes

src/style.css
```diff
@@ -1 +1,69 @@
...
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Roboto, Arial, Helvetica, sans-serif;
+}
+
+:root {
+  --regular-font: #000;
+  --hover-font: #00acdc;
```
witflash, line 11: It's better to name variables with a typed prefix at the start:
```
--font-regular;
--font-hover;
```

src/style.css
```diff
@@ -1 +1,69 @@
...
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  line-height: 14px;
```
witflash, line 23: `line-height` is better to set with relative value or in percentage.
For example:
```
line-height: 1;
line-height: 1.5;
line-height: 150%;
```
I like the relative values better.

witflash (approved):
Good job!
Pay attention to my last comments for your future work.

PR #2467, opened 2022-12-31
attempt1

Roman-Kulchytskyi:
https://roman-kulchytskyi.github.io/layout_moyo-header/
https://roman-kulchytskyi.github.io/layout_moyo-header/report/html_report/


src/style.css
```diff
@@ -1 +1,64 @@
...
+}
+
+.header {
+  display: flex;
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  margin: 0;
```
SanyaBratashchuk, line 16:
```suggestion
```
Header doesn't have default margins

src/style.css
```diff
@@ -1 +1,64 @@
...
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
```
SanyaBratashchuk, line 28:
```suggestion
  list-style: none;
  margin: 0;
  padding: 0;
```
(Optional) It is better to move this properies under the `ul` selector

src/style.css
```diff
@@ -1 +1,64 @@
...
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  text-transform: uppercase;
+  color: #000;
+}
+
+.is-active {
+  position: relative;
+  color: #00acdc;
```
SanyaBratashchuk, line 48: It is good practice to move all colors to the variables

SanyaBratashchuk (changes requested):

tetiana-ivanchenko (approved):

PR #2478, opened 2023-01-04
work done

kirill-vypirovskyi:
- [DEMO LINK] (https://kirill-vypirovskyi.github.io/layout_moyo-header/)
- [TEST REPORT LINK] (https://kirill-vypirovskyi.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,41 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="logo__link">
+        <img
+          src="./images/logo.png"
+          alt="moyo logo"
+          class="logo__image">
+      </a>
+      <nav class="navigation">
+        <ul class="menu">
+          <li class="menu__item">
+            <a href="#" class="menu__link is-active">Apple</a>
+          </li>
+          <li class="menu__item">
+            <a href="#" class="menu__link">Samsung</a>
+          </li>
+          <li class="menu__item">
+            <a href="#" class="menu__link">Smartphones</a>
+          </li>
+          <li class="menu__item" data-qa="hover">
+            <a href="#" class="menu__link">Laptops & Computers</a>
+          </li>
+          <li class="menu__item">
+            <a href="#" class="menu__link">Gadgets</a>
+          </li>
+          <li class="menu__item">
+            <a href="#" class="menu__link">Tablets</a>
+          </li>
+          <li class="menu__item">
+            <a href="#" class="menu__link">Photo</a>
+          </li>
+          <li class="menu__item">
+            <a href="#" class="menu__link">Video</a>
+          </li>
+        </ul>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -6,9 +6,61 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="logo__link">
+        <img
+          src="./images/logo.png"
+          alt="moyo logo"
+          class="logo__image">
```
OleksiiBlezniuk, line 26:
```suggestion
        <img
          src="./images/logo.png"
          alt="moyo logo"
          class="logo__image"
        >
```

src/style.css
```diff
@@ -1 +1,66 @@
-/* Styles go here */
+html,
+body,
+ul {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 6: This property is not inherited and you didn't set any paddings or borders to this tags, so it's unnecessary here

src/style.css
```diff
@@ -1 +1,66 @@
...
+  display: flex;
+  justify-content: space-between;
+  background-color: #fff;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo__link {
+  height: 40px;
+  margin: auto 0;
```
OleksiiBlezniuk, line 24: Ok, but as for me it's more logical to set `align-items` to parent instead (usually aligning by cross axis applied to all elements)

src/style.css
```diff
@@ -1 +1,66 @@
...
+  display: flex;
+  justify-content: end;
+  list-style: none;
+}
+
+.menu__item:not(:last-child) {
+  position: relative;
+  display: flex;
+  margin-right: 20px;
+}
```
OleksiiBlezniuk, line 37: I guess only margins shouldn't be applied to last child 🙂

src/style.css
```diff
@@ -1 +1,66 @@
...
+.menu__link.is-active {
+  color: #00acdc;
+}
+
+.menu__link.is-active::after {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  content: "";
+  width: 100%;
```
OleksiiBlezniuk, line 48: If you set width: 100% - it means that it will take all width, and no need to set `left: 0`

src/style.css
```diff
@@ -1 +1,66 @@
...
+  height: 4px;
+  border-radius: 8px;
+  background-color: #00acdc;
+}
+
+.menu__link {
+  --height: 60px;
```
OleksiiBlezniuk, line 55: Nice, but usually variables are created in root

OleksiiBlezniuk (approved):
Avoid applying unnecessary rules and keep in mind that `box-sizing` is not inherited rule

PR #2481, opened 2023-01-05
header added

daryna-hnidash:


mykhalenych:
Where are links with demo and test?

src/index.html
```diff
@@ -2,13 +2,79 @@
...
+        <img 
+          src="images/logo.png" 
+          alt="MOYO_logo"
+          class="logo"
+          width="40px"
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__list__item">
```
witflash, line 32: It's better to name complex element such `nav__list-item`. But not critical

src/style.css
```diff
@@ -1 +1,68 @@
...
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  background-color: var(--accent-blue);
+  width: 37px;
+  height: 4px;
+  border-radius: 8px;
+  margin-top: 18.5px;
```
witflash, line 66: Don't use not integer values in px. It will cause different problems in different browsers.

witflash (changes requested):
Good job!
Please do small improvements.

PR #2512, opened 2023-01-13
created and styled header

KseniaSkr:
- [DEMO LINK](https://KseniaSkr.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://KseniaSkr.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,83 @@
 /* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+}
+
+body {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+}
+
+.nav {
+  display: flex;
+  margin-right: 50px;
+}
+
+.nav_menu {
+  display: flex;
+  justify-content: flex-end;
+  margin: 0;
+  padding: 0;
+}
+
+.nav_item {
+
+  list-style: none;
+  margin: 0 10px;
+}
+
+.nav_item:first-child {
+  margin-left: 0;
+}
+
+.nav_item:last-child {
+  margin-right: 0;
+}
+
+.nav_link {
+  position: relative;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -1 +1,83 @@
...
+
+.nav_menu {
+  display: flex;
+  justify-content: flex-end;
+  margin: 0;
+  padding: 0;
+}
+
+.nav_item {
+
```
witflash, line 30: Redundant empty line

src/style.css
```diff
@@ -1 +1,83 @@
...
+  list-style: none;
+  margin: 0 10px;
```
witflash, line 32: you can use just `margin-right: 20px;`. It will be the same, but more correct

src/style.css
```diff
@@ -1 +1,83 @@
...
+}
+
+.nav_item:first-child {
+  margin-left: 0;
```
witflash, line 36: If you use the rule from prev comment, you will not need to use this selector.

src/style.css
```diff
@@ -1 +1,83 @@
...
+}
+
+.nav_link:hover {
+  color: #00acdc;
+  cursor: pointer;
+}
+
+.nav_link:hover::after {
+  text-decoration: underline;
+}
```
witflash, line 60: Do you really need this selector? It's redundant I think. It doesn't do anything.

witflash (changes requested):
Great work!

src/style.css
```diff
@@ -1 +1,77 @@
...
+  position: absolute;
+  background-color: #00acdc;
+  right: 0;
+  left: 0;
+  bottom: 0;
+  height: 4px;
+  border-radius: 8px;
+  content: "";
+  margin-top: 19px;
+}
```
IvanFesenko, line 77: margin is useless

IvanFesenko (approved):
Good job

PR #2521, opened 2023-01-16
Develop

sergik94:
https://sergik94.github.io/layout_moyo-header/
https://sergik94.github.io/layout_moyo-header/report/html_report/

Не виходить пройти 2 тести.


src/index.html
```diff
@@ -7,8 +7,55 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css?family=Roboto:regular,500" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="header__container">
```
witflash, line 14: You don't need this additional wrapper. Remove this `div` and add all needed styles to `.header`

src/style.css
```diff
@@ -1 +1,68 @@
...
+  color: inherit;
+}
+
+body {
+  margin: 0;
+  --linkColor: #00acdc;
+}
+
+.header {
+  margin: 0 50px;
```
witflash, line 16: You don't need to use margin here. It looks like padding for inner content. So, use better:
```
padding-left: 50px;
padding-right: 50px;
```
And as a bonus - your tests will pass.

src/style.css
```diff
@@ -1 +1,68 @@
...
+}
+
+.header__container {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
```
witflash, line 22: General font rules move to the `body` selector.

src/style.css
```diff
@@ -1 +1,68 @@
...
+}
+
+a {
+  text-decoration: none;
+  color: inherit;
+}
+
+body {
+  margin: 0;
+  --linkColor: #00acdc;
```
witflash, line 12: It's better to use kebab-case for all CSS: `--link-color`. Or better `--color-link`.

witflash (changes requested):
Good job! 👍 

witflash (changes requested):
Good job! 👍

mykhalenych (approved):

PR #2535, opened 2023-01-20
MOYO header 

sich-mykhailo:


src/style.css
```diff
@@ -1 +1,73 @@
...
+  font-family: Roboto, serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  color: #000;
+}
+
+.header {
+  padding-left: 50px;
+  padding-right: 50px;
```
vkryvytskyy, line 21:
```
padding: 0 50px;

File: src/style.css [hunk 2]
```diff
@@ -1 +1,73 @@
...
+
+.logo {
+  display: flex;
+  align-items: center;
+}
+
+.nav-list {
+  display: flex;
+  position: relative;
+  top: 22.5px;
```
vkryvytskyy, line 37: don't use decimal values

src/style.css
```diff
@@ -1 +1,73 @@
...
+.nav-item:first-child {
+  padding-left: 0;
+}
+
+.nav-link {
+  text-decoration: none;
+  color: #000;
+  text-transform: uppercase;
+  padding-top: 22.5px;
+  padding-bottom: 22.5px;
```
vkryvytskyy, line 54: don't use decimal values

src/style.css
```diff
@@ -1 +1,73 @@
...
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  position: relative;
+  top: 19px;
```
vkryvytskyy, line 67: use bottom, so the element is always at the bottom of the item, no matter it's height

vkryvytskyy (changes requested):

PR #2547, opened 2023-01-23
MOYO added

MykolaLopushenko:
[DEMO LINK](https://MykolaLopushenko.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://MykolaLopushenko.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,64 @@
-/* Styles go here */
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  text-transform: uppercase;
+  padding-left: 50px;
+  padding-right: 50px;
+  cursor: pointer;
+}
+
+ul {
+  list-style: none;
+}
+
+a {
+  text-decoration: none;
+  color: inherit;
+}
+
+.logo__img {
+  display: flex;
+}
+
+.nav__menu {
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.nav__item {
+  position: relative;
+}
+
+.nav__item a {
+  display: flex;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -1 +1,64 @@
...
+}
+
+.nav__menu {
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.nav__item {
+  position: relative;
```
mykhalenych, line 38: is-active is relative

src/style.css
```diff
@@ -1 +1,64 @@
...
+}
+
+.nav__item a {
```
mykhalenych, line 41: don't use tag selector use classname instead

src/style.css
```diff
@@ -1 +1,64 @@
...
+  line-height: 60px;
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__item:hover,
+.is-active {
+  color: #00acdc;
```
mykhalenych, line 52: create variable and use it instead of 00acdc in two places

src/style.css
```diff
@@ -1 +1,64 @@
...
+  list-style: none;
+}
+
+a {
+  text-decoration: none;
+  color: inherit;
+}
+
+.logo__img {
+  display: flex;
```
mykhalenych, line 28: why does image need flex?

src/style.css
```diff
@@ -1 +1,64 @@
-/* Styles go here */
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
```
mykhalenych, line 3: font-family and font-size move to html selector

mykhalenych (changes requested):

src/style.css
MykolaLopushenko: @mykhalenych plz check

mykhalenych (approved):

PR #2549, opened 2023-01-24
create navbar menu

mstrilec:
- [DEMO LINK](https://mstrilec.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://mstrilec.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,59 @@
 /* Styles go here */
+body {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", Arial, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.nav {
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav__list {
+  display: flex;
+  align-items: center;
+  list-style: none;
+  padding: 23px 0;
+  margin: 0;
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  padding: 23px 0;
+  text-decoration: none;
+  color: #000;
+  font-family: inherit;
+  font-weight: 500;
+  line-height: 14px;
+  text-transform: uppercase;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  position: relative;
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: 4px;
+  border-radius: 8px;
+  background-color: #00acdc;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,59 @@
 /* Styles go here */
+body {
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 3: This value is not inherited, body don't have any paddings or margins by default, so this property is unnecessary

src/style.css
```diff
@@ -1 +1,59 @@
...
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 5: body don't have paddings

src/style.css
```diff
@@ -1 +1,59 @@
...
+  font-family: "Roboto", Arial, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
```
OleksiiBlezniuk, line 8: Usually we set font styles to `html` tag

src/style.css
```diff
@@ -1 +1,59 @@
...
+  justify-content: space-between;
+  align-items: center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav__list {
+  display: flex;
+  align-items: center;
+  list-style: none;
+  padding: 23px 0;
```
OleksiiBlezniuk, line 23: You already set same paddings to link, so no need to duplicate, container will het this child height (but your links should be blocks or inline-blocks, vertical paddings works incorrect for inline elements)

src/style.css
```diff
@@ -1 +1,59 @@
...
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  padding: 23px 0;
+  text-decoration: none;
+  color: #000;
+  font-family: inherit;
+  font-weight: 500;
```
OleksiiBlezniuk, line 36:
```suggestion
  font-weight: inherit;
```

OleksiiBlezniuk (changes requested):

src/style.css
mstrilec: I need padding in any case, because the indents for the link increase the clickable area, and the indents for the list are needed so that the elements do not stick to the top of the page. Thank you, I will take it into account, but I tried to fix it, but it didn't work for me. I must be doing something wrong.

OleksiiBlezniuk: Actually you don't need them 🙂 As I said you need to set `display: block` to links

OleksiiBlezniuk: https://user-images.githubusercontent.com/84275496/214909096-e63b71f9-2e2b-4f50-92cf-f3e51a971edb.mov

OleksiiBlezniuk: You can leave it as it is now, it's not critical, but keep in mind that vertical margins and paddings works incorrect for inline elements

OleksiiBlezniuk (approved):

src/style.css
mstrilec: Thank you, I understood what my mistake was.

PR #2550, opened 2023-01-24
header was created, styles were added

vladyslav-mosor:
- [DEMO LINK](https://vladyslav-mosor.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://vladyslav-mosor.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,70 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    >
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="header__logo">
+        <img src="images/logo.png"
+        alt="logo"
+        class="logo__img"
+        >
+      </a>
+      <nav class="header__nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">
+              Apple
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">
+              Samsung
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">
+              Smartphones
+            </a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#laptops"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -6,9 +6,70 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="header__logo">
+        <img src="images/logo.png"
+        alt="logo"
+        class="logo__img"
+        >
```
OleksiiBlezniuk, line 23: Screenshot 2023-01-25 at 15 45 55

OleksiiBlezniuk: Pay attention to code style

src/style.css
```diff
@@ -1 +1,74 @@
...
+ul {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.header {
+  position: fixed;
+  left: 0;
+  right: 0;
```
OleksiiBlezniuk, line 17: Ok, but this styles are not required for this task

src/style.css
```diff
@@ -1 +1,74 @@
...
+  max-width: 1200px;
+  margin: auto;
```
OleksiiBlezniuk, line 19:
```suggestion
  margin: 0 auto;
```

src/style.css
```diff
@@ -1 +1,74 @@
...
+
+.logo__img {
+  display: block;
+  height: 40px;
+}
+
+.nav__list {
+  list-style: none;
+  display: flex;
+  align-items: center;
```
OleksiiBlezniuk, line 36: you already used paddings for links, so they took all free height, so there is no need to align elements horizontally

src/style.css
```diff
@@ -1 +1,74 @@
...
+  flex-wrap: wrap;
+}
+
+.nav__item {
+  margin-right: 20px;
+  height: 60px;
```
OleksiiBlezniuk, line 42: set proper paddings instead
we don't need to align inner items using flex, so there is no need to hardcode height, let content to dictate height of its container

src/style.css
```diff
@@ -1 +1,74 @@
...
+
+.nav__item:last-child {
+  margin: 0;
+}
+
+.nav__link {
+  position: relative;
+  display: flex;
+  align-items: center;
+  height: 100%;
```
OleksiiBlezniuk, line 53: just set proper paddings to link

OleksiiBlezniuk (changes requested):
Nice idea to center this header, but in this case better to center content of header, this horizontal borders are not looks good
Screenshot 2023-01-25 at 19 22 01

src/style.css
```diff
@@ -1 +1,67 @@
...
+  font-weight: 500;
+  font-size: 12px;
+}
+
+body,
+ul {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
```
OleksiiBlezniuk, line 12: body don't have default paddings, and there is no need to set `box-sizing` to body (it's not inherited)

OleksiiBlezniuk (approved):

PR #2553, opened 2023-01-24
Moyo header

KuharchukDmytro:
https://KuharchukDmytro.github.io/layout_moyo-header/
https://KuharchukDmytro.github.io/layout_moyo-header/report/html_report/


src/style.css
```diff
@@ -1 +1,65 @@
...
+}
+
+.header__menu {
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+}
+
+.header__icon {
+  margin-top: 10px;
```
OleksiiBlezniuk, line 20:
```suggestion
.header__menu {
  padding: 0 50px;
  display: flex;
  justify-content: space-between;
  align-items: center;
}

.header__icon {
  line-height: 0; /* to avoid extra space added because of link */
```

src/style.css
```diff
@@ -1 +1,65 @@
...
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  display: flex;
+  align-items: center;
+  height: 60px;
```
OleksiiBlezniuk, line 40: Better to avoid height hardcoding, in that case better to set proper paddings instead

OleksiiBlezniuk (changes requested):
GJ. but uou forgot to add box-shadow. Fix it please
Screenshot 2023-01-25 at 19 39 24

OleksiiBlezniuk (approved):
👍

PR #2554, opened 2023-01-24
created moyo header css html

ksavelyeva:
[DEMO LINK](https://ksavelyeva.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://ksavelyeva.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,62 @@
-/* Styles go here */
+body {
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 3: body don't have default paddings

src/style.css
```diff
@@ -1 +1,62 @@
...
+  font-family: Roboto, Arial, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
```
OleksiiBlezniuk, line 6: set font styles to `html` tag

src/style.css
```diff
@@ -1 +1,62 @@
...
+  line-height: 14px;
+  text-transform: uppercase;
```
OleksiiBlezniuk, line 8: we don't need this rule for body, only for link (we don't want to uppercase all text on the page)

src/style.css
```diff
@@ -1 +1,62 @@
...
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  background: #fff;
```
OleksiiBlezniuk, line 16: it's a default background color

src/style.css
```diff
@@ -1 +1,62 @@
...
+  width: 40px;
+  height: 40px;
+  display: block;
+}
+
+.nav__list {
+  display: flex;
+  align-items: center;
+  list-style: none;
+  padding: 23px 0;
```
OleksiiBlezniuk, line 30: You've set this same paddings to links, no need to duplicate
(just set block or inline-block to links)

src/style.css
```diff
@@ -1 +1,62 @@
...
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  width: 100%;
+  height: 4px;
+  background: #00acdc;
+  border-radius: 8px;
+  bottom: 0;
+  left: 0;
```
OleksiiBlezniuk, line 61: you have `width: 100%`, no need to set `left`

OleksiiBlezniuk (changes requested):

src/style.css
```diff
@@ -1 +1,65 @@
...
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  right: 0;
+  height: 4px;
+  background: #00acdc;
+  border-radius: 8px;
+  bottom: 0;
+  left: 0;
```
OleksiiBlezniuk, line 64:
```suggestion
  content: "";
  position: absolute;
  height: 4px;
  background: #00acdc;
  border-radius: 8px;
  bottom: 0;
  left: 0;
  right: 0;
```

OleksiiBlezniuk (approved):

PR #2560, opened 2023-01-25
moyo-header added

NikitaStambul:
- [DEMO LINK](https://NikitaStambul.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://NikitaStambul.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,71 @@
 /* Styles go here */
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+  padding: 0 50px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,71 @@
...
+  align-items: center;
+  width: 100%;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
```
OleksiiBlezniuk, line 29:
```suggestion
.logo {
  line-height: 0;
}
```

OleksiiBlezniuk: This size are same as in image, so no need to duplicate
You can just set this line-height to remove extra few pixels at the bottom of link

src/style.css
```diff
@@ -1 +1,71 @@
...
+
+.nav__list {
+  display: flex;
+  justify-content: flex-end;
```
OleksiiBlezniuk, line 33: this rule is unnecessary here

OleksiiBlezniuk (approved):

PR #2561, opened 2023-01-25
moyo header is done

stanislavkulakovskyi:
[DEMO LINK](https://stanislavkulakovskyi.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://stanislavkulakovskyi.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,58 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    >
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo"><img src="./images/logo.png" alt="logo"></a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__list-item">
+            <a class="nav__link is-active" href="#apple">Apple</a>
+          </li>
+          <li class="nav__list-item">
+            <a class="nav__link" href="#samsung">Samsung</a>
+          </li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,14 +1,58 @@
...
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    >
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo"><img src="./images/logo.png" alt="logo"></a>
```
OleksiiBlezniuk, line 21:
```suggestion
      <a href="#home" class="logo">
        <img src="./images/logo.png" alt="logo">
      </a>
```

OleksiiBlezniuk: Not critical, but this way is more readable

src/style.css
```diff
@@ -1 +1,74 @@
-/* Styles go here */
+body {
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 2: Body don't have any paddings or borders, so I guess this rule is unnecessary here (this rule is not inherited)

src/style.css
```diff
@@ -1 +1,74 @@
...
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 4: body don't have default paddings, remove it

src/style.css
```diff
@@ -1 +1,74 @@
...
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
+  padding: 10px 0;
+}
```
OleksiiBlezniuk, line 17: Your header is flexbox, so it's better to align elements using `align-items` instead of padding. That way will allow us to don't worry about changes of logo size of header height (now it can be broken by this changes)

src/style.css
```diff
@@ -1 +1,74 @@
...
+  padding: 23px 0;
+}
+
+.nav__list-item {
+  margin-right: 20px;
+}
+
+.nav__list-item:last-child {
+  margin-right: 0;
+}
```
OleksiiBlezniuk, line 54: use `:not`

src/style.css
```diff
@@ -1 +1,74 @@
...
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+  position: relative;
```
OleksiiBlezniuk, line 62: I guess it's more logic to set this rule to nav__link

src/style.css
```diff
@@ -1 +1,74 @@
...
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  height: 4px;
+  width: 100%;
```
OleksiiBlezniuk, line 71:
```suggestion
  bottom: 0;
  height: 4px;
  width: 100%;
```

or

```suggestion
  bottom: 0;
  left: 0;
  right: 0;
  height: 4px;
```

OleksiiBlezniuk (changes requested):
GJ, but there are few issues to fix

PR #2563, opened 2023-01-25
completed task

sasha-dranuta:
- [DEMO LINK](https://sasha-dranuta.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://sasha-dranuta.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,72 @@
-/* Styles go here */
+
+html,
+body,
+ul {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
+  max-width: 1200px;
+  justify-content: space-between;
+  align-items: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,72 @@
-/* Styles go here */
+
+html,
+body,
+ul {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
```
OleksiiBlezniuk, line 8: Not every this tag has this default styles, better to reset styles for particular tags instead
And `box-sizing` rule is not inherited, so there is no sense to set it to html and body tags

src/style.css
```diff
@@ -1 +1,72 @@
...
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  background-color: #00acdc;
+  height: 4px;
+  width: 100%;
+  left: 0;
+  bottom: 0;
```
OleksiiBlezniuk, line 62:
```suggestion
  right: 0;
  left: 0;
  bottom: 0;
```

OleksiiBlezniuk: That way is more clear as for me

OleksiiBlezniuk (approved):

PR #2566, opened 2023-01-25
navbar created

timoxa2t:
- [DEMO LINK](https://timoxa2t.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://timoxa2t.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,82 @@
 /* Styles go here */
+a,
+li,
+ul,
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+a {
+  color: #000;
+  text-decoration: none;
+}
+
+li {
+  list-style: none;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+.nav__list {
+  min-width: max-content;
+}
+
+.nav__item {
+  display: inline-block;
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 17px;
+}
+
+.nav__link {
+  position: relative;
+  display: inline-block;
+  height: 60px;
+  line-height: 60px;
+  text-align: center;
+  text-transform: uppercase;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  display: inline-block;
+  bottom: 0;
+  left: 0;
+  width: 100%;
+  height: 4px;
+  border-radius: 2px;
+  background-color: #00acdc;
+}
+
+.logo__img {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,82 @@
...
+li {
+  list-style: none;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
```
OleksiiBlezniuk, line 31: Don't use tags for styling, only to reset some default styles

src/style.css
```diff
@@ -1 +1,82 @@
...
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  display: inline-block;
+  bottom: 0;
+  left: 0;
+  width: 100%;
+  height: 4px;
```
OleksiiBlezniuk, line 69:
```suggestion
  left: 0;
  right: 0;
  height: 4px;
```

OleksiiBlezniuk: Not critical, but as for me - more logical

src/style.css
```diff
@@ -1 +1,82 @@
...
+  border-radius: 2px;
+  background-color: #00acdc;
+}
+
+.logo__img {
+  width: 40px;
+  height: 40px;
+}
```
OleksiiBlezniuk, line 77: No need to set this sizes, they are same as in image

OleksiiBlezniuk (changes requested):

src/style.css
```diff
@@ -1 +1,75 @@
 /* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 4: Don't use `*` selector for styles reseting, only for `box-sizing` to apply it to all elements

src/style.css
```diff
@@ -1 +1,75 @@
...
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
```
OleksiiBlezniuk, line 12: Set font styles to `html` tag

src/style.css
```diff
@@ -1 +1,75 @@
...
+  min-width: max-content;
+}
+
+.nav__item {
+  display: inline-block;
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 17px;
+}
```
OleksiiBlezniuk, line 41: Why such strange value? Usually we use for sizes values that are divided by 4

src/style.css
```diff
@@ -1 +1,75 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav__list {
+  min-width: max-content;
+}
```
OleksiiBlezniuk, line 33: `display: flex` for this element will help to fix extra space between list items and to use correct margin from mock

OleksiiBlezniuk (changes requested):

illiarerun (approved):

PR #2588, opened 2023-01-26
Develop

vadosik15:
- [DEMO LINK](https://vadosik15.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://vadosik15.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,75 @@
...
+ul {
+  display: flex;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+}
+
+li:nth-child(-n+7) {
+  margin-right: 20px;
+}
```
Andr1yK, line 31: avoid global styling by tag 

What happened if I wanna use li in other palaces?

src/style.css
```diff
@@ -1 +1,75 @@
...
+  width: 40px;
+  height: 40px;
+}
+
+ul {
+  display: flex;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+}
```
Andr1yK, line 27: u can reset styles for tag, but adding `display` for `reset.css` it's bad practice

```suggestion
ul {
  list-style-type: none;
  margin: 0;
  padding: 0;
}
```

src/style.css
```diff
@@ -1 +1,75 @@
...
+  cursor: pointer;
+}
+
+.is-nav__item > :first-child {
+  color: #00acdc;
+}
+
+.nav__item:hover > :first-child {
+  color: #00acdc;
+}
```
Andr1yK, line 45: as for me, u can just use .nav__link class

src/style.css
```diff
@@ -1 +1,75 @@
...
+  align-items: center;
+  text-decoration: none;
+  text-align: center;
+  color: #000;
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  height: 60px;
+}
```
Andr1yK, line 64: the same as for `ul`

Andr1yK (changes requested):
why do u need to change the reference?

image

PR #2567, opened 2023-01-25
Header added2

yaroslav-vasilyev:
- [DEMO LINK](https://yaroslav-vasilyev.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://yaroslav-vasilyev.github.io/layout_moyo-header/report/html_report/)

PR #2593, opened 2023-01-27
Moyo Header was fixed

yaroslav-vasilyev:


src/style.css
```diff
@@ -1 +1,71 @@
...
+  right: 0;
+  left: 0;
+  bottom: 0;
+  height: 4px;
+  content: "";
+  background: hsl(193, 100%, 43%);
+  border-radius: 8px;
+}
+
+.link--1 {
```
illiarerun, line 68: Nope, that's not right. Imagine that I remove first element, or want to add another one, then I'll need to move this class to it. Use pseudoclasses for setting styles based on position of element.

src/style.css
```diff
@@ -1 +1,71 @@
...
+
+.is-active::after {
+  display: inline-block;
+  position: absolute;
+  right: 0;
+  left: 0;
+  bottom: 0;
+  height: 4px;
+  content: "";
+  background: hsl(193, 100%, 43%);
```
illiarerun, line 64: you use `rgba` everywhere, but decided to use `hsl` here, why? try to be consistent with color schemes that you use

src/style.css
```diff
@@ -1 +1,71 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.nav__item {
+  list-style: none;
+}
+
+.nav__link {
+  display: inline-block;
```
illiarerun, line 36: why `inline-block`? you already have flex around it
```suggestion
  display: block;
```

illiarerun (changes requested):

src/style.css
yaroslav-vasilyev: .nav__item:first-child > .nav__link {
  margin: 0;
}

@[OleksiiBlezniuk](https://github.com/OleksiiBlezniuk) OleksiiBlezniuk [4 days ago](https://github.com/mate-academy/layout_moyo-header/pull/2567#discussion_r1087957861)
avoid specificity increasing without need (you don't need it here)

What should I use?

illiarerun: @yaroslav-vasilyev add margin to `nav__item` instead, just to avoid "випадання марджинів"

src/style.css
```diff
@@ -1 +1,75 @@
...
+  bottom: 0;
+  height: 4px;
+  content: "";
+  background: rgb(0, 172, 219);
+  border-radius: 8px;
+}
+
+.nav__item:first-child > .nav__link {
+  margin: 0;
+}
```
illiarerun, line 74: don't forget to remove this
```suggestion
```

illiarerun (approved):

PR #2606, opened 2023-01-30
changed

vadosik15:
- [DEMO LINK](https://vadosik15.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://vadosik15.github.io/layout_moyo-header/report/html_report/)


Andr1yK (changes requested):
Where is your solution?)

Andr1yK (changes requested):
What do I need to check?)

vadosik15:
I don't know why this solution is not working, I tried share this task from another pc and it is still not working

PR #2607, opened 2023-01-30
initial commit

grinjk1993:
https://grinjk1993.github.io/layout_moyo-header/
https://grinjk1993.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,63 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin: 0;
+  padding: 0 50px;
+  border-radius: 4px;
+  box-shadow: 0 3px 5px rgba(0, 0, 0, 0);
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
+}
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,63 @@
...
+  padding: 22.6px 0;
+  font-size: 12px;
+  text-decoration: none;
+  text-transform: uppercase;
+}
+
+.nav__link.is-active,
+.nav__link:hover {
+  position: relative;
+  color: #00acdc;
```
mykhalenych, line 51: move to variable and use variable instead of 00acdc in two places

src/style.css
```diff
@@ -1 +1,63 @@
...
+  display: flex;
+}
+
+.nav__item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav__link {
+  color: #000;
+  padding: 22.6px 0;
```
mykhalenych, line 42: don't use fractional numbers for padding

mykhalenych (changes requested):
Check comments also don't forget add links to readme.md

src/style.css
grinjk1993: ok, done

src/style.css
grinjk1993: got it, amended the file

mykhalenych (approved):

PR #2620, opened 2023-02-02
create moyo header

semchuk-rodion:
https://semchuk-rodion.github.io/layout_moyo-header/
https://semchuk-rodion.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,71 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+.header {
+  padding: 0 50px;
+  margin: 0;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+}
+
+.nav__list {
+  display: flex;
+  justify-content: space-between;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.nav__item {
+  display: flex;
+}
+
+.nav__item:not(:first-child) {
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -1 +1,71 @@
...
+.is-active,
+.nav__link:hover {
+  position: relative;
+  color: #00acdc;
+}
+
+.is-active::after {
+  display: flex;
+  height: 4px;
+  width: 37px;
```
OlEzhA0, line 65: don't hardcode width
just add 100%

OlEzhA0 (changes requested):

src/style.css
```diff
@@ -1 +1,69 @@
...
+
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 60px;
+
+  text-transform: uppercase;
+  text-decoration: none;
+  color: #000;
+
+}
```
ivanchuk-yurii, line 52:
```suggestion
.nav__link {
  font-weight: 500;
  font-size: 12px;
  line-height: 60px;

  text-transform: uppercase;
  text-decoration: none;
  color: #000;
}
```

src/style.css
```diff
@@ -1 +1,69 @@
...
+.img {
+  height: 40px;
+  width: 40px;
+}
+
+.nav__link {
+
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 60px;
```
ivanchuk-yurii, line 46: it's better to set vertical padding instead

ivanchuk-yurii (approved):

PR #2624, opened 2023-02-03
Moyo header

Dimoshua:
Request description:
- [DEMO LINK](https://Dimoshua.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Dimoshua.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,96 @@
...
+body {
+  margin: 0;
+  --main-color: #00acdc;
+}
+
+.container {
+  padding: 0 50px;
+
+  display: flex;
+  flex-direction: row;
```
SanyaBratashchuk, line 28:
```suggestion
```
Default value

src/style.css
```diff
@@ -1 +1,96 @@
...
+  position: relative;
+  width: 40px;
+  height: 40px;
+  background-image: url("./images/logo.png");
+  background-repeat: no-repeat;
+}
+
+.nav {
+  display: flex;
+  flex-direction: row;
```
SanyaBratashchuk, line 46:
```suggestion
```

src/style.css
```diff
@@ -1 +1,96 @@
...
+
+.nav__list:first-child::after {
+  content: "";
+  position: absolute;
+  background: var(--main-color);
+  height: 4px;
+  width: 37px;
+  margin: 56px auto 0;
+  border-radius: 8px;
+}
```
SanyaBratashchuk, line 67: move it to `.is-active::after`

SanyaBratashchuk (changes requested):

src/style.css
Dimoshua: If they change like this, then the tests will not pass. Tests do not see the underlined line

Dimoshua: I solved the problem by inserting the is-active class twice:

<ul class="nav__list is-active">
            <a class="box is-active" href="#apple">Apple</a>
          </ul>

Dimoshua: I solved the problem by inserting the is-active class twice in *.html

PR #2638, opened 2023-02-06
create moyo header

LBondarchuk:
- [DEMO LINK](https://LBondarchuk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://LBondarchuk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -7,6 +7,7 @@ body {
   box-sizing: border-box;
   height: 100vh;
   background: #e5e5e5;
+  font-family: "Roboto", sans-serif;
 }
 
 .logo-link {
@@ -16,6 +17,7 @@ body {
 }
 
 .header {
+  margin: 0 auto;
   max-width: 1200px;
   min-width: 1024px;
   display: flex;
@@ -46,7 +48,6 @@ body {
   display: flex;
   align-items: center;
   text-transform: uppercase;
-  font-family: "Roboto", sans-serif;
   text-decoration: none;
   color: black;
   font-weight: 500;
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,75 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  box-sizing: border-box;
+  height: 100vh;
+  background: #e5e5e5;
+}
+
+.logo-link {
+  margin-left: 50px;
+  height: 40px;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -1 +1,75 @@
...
+
+.nav-item:last-child {
+  margin: 0;
+}
+
+.nav-link {
+  display: flex;
+  align-items: center;
+  text-transform: uppercase;
+  font-family: "Roboto", sans-serif;
```
slavanosatskyi, line 49: usually it is better to apply font styles on a higher level, like in body tag

src/style.css
```diff
@@ -1 +1,75 @@
...
+  max-width: 1200px;
+  min-width: 1024px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background-color: #fff;
+}
+
+.nav {
+  margin-right: 50px;
```
slavanosatskyi, line 28: instead you can add paddings to your header

src/style.css
```diff
@@ -1 +1,75 @@
...
+}
+
+body {
+  box-sizing: border-box;
+  height: 100vh;
+  background: #e5e5e5;
+}
+
+.logo-link {
+  margin-left: 50px;
```
slavanosatskyi, line 13: instead you can add paddings to your header

slavanosatskyi (changes requested):

slavanosatskyi:
Screenshot 2023-02-06 at 18 39 17
Please, pay attention that for really big screens header is not centered

src/style.css
LBondarchuk: I tried to do this from the beginning. But npm test treated it as an error.

LBondarchuk (commented):
I fixed the code.
But the padding did not change. The test then does not work correctly

src/style.css
```diff
@@ -1 +1,76 @@
...
+  text-decoration: none;
+}
+
+.header {
+  margin: 0 auto;
+  max-width: 1200px;
+  min-width: 1024px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
```
polosanya, line 25: Pay attention to CSS properties order

polosanya (changes requested):
Hi! I see that the problem from the previous comment was not fixed. Try to make the width 100% for the header.
image

PR #2653, opened 2023-02-09
added moyo header

helenmarth:
- [DEMO LINK](https://helenmarth.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://helenmarth.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,13 +13,14 @@
   </head>
   <body>
     <header class="header"> 
-      <img
-        src="./images/logo.png"
-        alt="logo"
-        class="header__logo"
-      >
-      <nav class="header__nav">
-        <ul class="header__list">
+      <a href="#" class="header__logo">
+        <img
+          src="./images/logo.png"
+          alt="logo"
+        >
+      </a>
+      <nav class="header__nav nav">
+        <ul class="nav__list">
           <li class="nav__item">
             <a href="#apple" class=" nav__link is-active">Apple</a>
           </li>
--- a/src/style.css
+++ b/src/style.css
@@ -1,40 +1,38 @@
-* {
-  margin: 0;
-  padding: 0;
-}
-
 body {
-  font-family: "Roboto", sans-serif;
+  box-sizing: border-box;
+  font-family: Roboto, sans-serif;
   font-style: normal;
+  margin: 0;
 }
 
 .header {
   display: flex;
   justify-content: space-between;
   align-items: center;
-  background: #fff;
   box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
   max-width: 100%;
   padding: 0 50px;
   --accent-color: #00acdc;
 }
 
-.logo {
-  padding-top: 5px;
+.header__logo {
+  height: 40px;
+  width: 40px;
 }
 
-.header__list {
+.nav__list {
   display: flex;
+  margin: 0;
+  padding-left: 0;
 }
 
 .nav__item {
   list-style: none;
   display: flex;
-  padding-left: 20px;
 }
 
-.nav__item:first-child {
-  padding-left: 0;
+.nav__item:not(:last-child) {
+  margin-right: 20px;
 }
 
 .nav__link {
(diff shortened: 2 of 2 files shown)


src/index.html
```diff
@@ -1,14 +1,64 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header"> 
+      <img
+        src="./images/logo.png"
+        alt="logo"
+        class="header__logo"
+      >
+      <nav class="header__nav">
```
Viktor-Kost, line 21: Your nav__list, nav__item etc should have a block
```suggestion
      <nav class="header__nav nav">
```

src/index.html
```diff
@@ -1,14 +1,64 @@
...
+        <ul class="header__list">
```
Viktor-Kost, line 22:
```suggestion
        <ul class="nav__list">
```

src/index.html
```diff
@@ -1,14 +1,64 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header"> 
+      <img
+        src="./images/logo.png"
+        alt="logo"
+        class="header__logo"
+      >
```
Viktor-Kost, line 20: Logo should be a link with an image inside

src/style.css
```diff
@@ -1 +1,74 @@
...
+}
+
+.header__list {
+  display: flex;
+}
+
+.nav__item {
+  list-style: none;
+  display: flex;
+  padding-left: 20px;
```
Viktor-Kost, line 33: Should be margin-left instead, because it's not an inner area of an element. It sets a distance between two elements
```suggestion
  margin-left: 20px;
```

src/style.css
```diff
@@ -1 +1,74 @@
...
+}
+
+.nav__item:first-child {
```
Viktor-Kost, line 36: just use `not:(:first-child)` to add margins for all elements besides first one instead of using two rules

src/style.css
```diff
@@ -1 +1,74 @@
...
+  color: var(--accent-color);
+}
+
+.nav__link.is-active {
+  position: relative;
+}
+
+.is-active.nav__link {
+  color: var(--accent-color);
+}
```
Viktor-Kost, line 64: combine rules with the same styles
```suggestion
.nav__link:hover,
.is-active.nav__link  {
  color: var(--accent-color);
}

.nav__link.is-active {
  position: relative;
}
```

src/style.css
```diff
@@ -1 +1,74 @@
...
+
+.nav__link {
+  padding: 23px 0;
+  text-decoration: none;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  color: #000;
+  transition: all 0.5s;
```
Viktor-Kost, line 48: it is better to specify what properties should be changed instead of default value `all`

src/style.css
```diff
@@ -1 +1,74 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
```
Viktor-Kost, line 3: only `box-sizing: border-box `is allowed to use at `*` selector

src/style.css
```diff
@@ -1 +1,74 @@
...
+body {
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background: #fff;
```
Viktor-Kost, line 15: it is a default value
```suggestion
```

Viktor-Kost (changes requested):

src/index.html
helenmarth: If i didn't use this class in scc, can i delete it?

vlad-demchuk (changes requested):
looks like you didn't push changes

check the previous comments

src/style.css
```diff
@@ -1 +1,69 @@
...
+  padding: 23px 0;
+  text-decoration: none;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  color: #000;
+  transition: color 0.5s;
+  display: flex;
+  align-items: center;
```
etojeDenys, line 48: redundant

etojeDenys (approved):
LGTM

PR #2663, opened 2023-02-14
Make header

ShplmnVtlk:
https://ShplmnVtlk.github.io/layout_moyo-header/

https://ShplmnVtlk.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,75 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  width: 100%;
+  max-width: 1200px;
+  margin: 0 auto;
+  display: flex;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,75 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
```
vlad-demchuk, line 3: don't reset styles for all elements (`*`)

src/style.css
```diff
@@ -1 +1,75 @@
...
+  height: 4px;
+  border-radius: 8px;
+  background-color: #00acdc;
+  margin-top: 18px;
+  position: absolute;
+  opacity: 0;
+}
+
+.nav__link.is-active {
+  color: #00acdc;
```
vlad-demchuk, line 70: better to create a variable for color

vlad-demchuk (approved):

PR #2717, opened 2023-02-22
Create HTML page with the header

ychaiun:
- [DEMO LINK](https://ychaiun.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ychaiun.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -66,7 +66,7 @@ body {
   display: block;
   position: absolute;
   bottom: 0;
-  width: 37px;
+  width: 100%;
   height: 4px;
   background-color: #00acdc;
   border-radius: 8px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,73 @@
...
+  position: relative;
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  position: absolute;
+  bottom: 0;
+  width: 37px;
```
anstsot, line 69: it's better not to hardcode the width
```suggestion
  width: 100%;
```

anstsot (approved):
Great job! I'm approving but it would be nice to add small change with width

PR #2731, opened 2023-02-22
moyo header

danylo-dobrovolskyi:
- [DEMO LINK](https://danylo-dobrovolskyi.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://danylo-dobrovolskyi.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,67 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+}
+
+.header {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+}
+
+.moyo_logo {
+  position: absolute;
+  width: 40px;
+  height: 40px;
+  padding: 10px 50px;
+}
+
+.navigation {
+  background-color: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  text-align: right;
+}
+
+.list {
+  list-style: none;
+  margin: auto 40px;
+}
+
+.list_item {
+  display: inline-block;
+  height: 60px;
+}
+
+.list_link {
+  display: inline-block;
+  color: #000;
+  text-decoration: none;
+  text-transform: uppercase;
+  padding: 23px 10px;
+}
+
+.list_link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  display: block;
+  position: absolute;
+  content: "";
+  border-radius: 8px;
+  background: #00acdc;
+  height: 4px;
+  width: 37px;
+  top: 56px;
+}
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@mate-academy/bemlint": "^0.1.1",
     "@mate-academy/eslint-config": "*",
     "@mate-academy/linthtml-config": "0.0.2",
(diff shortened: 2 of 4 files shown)


src/style.css
```diff
@@ -1 +1,67 @@
...
+  flex-direction: column;
+  justify-content: space-between;
+}
+
+.moyo_logo {
+  position: absolute;
+  width: 40px;
+  height: 40px;
+  padding: 10px 50px;
+}
```
Olexandr6, line 24: Don't use `position: absolute;` here

src/style.css
```diff
@@ -1 +1,67 @@
...
+
+.is-active::after {
+  display: block;
+  position: absolute;
+  content: "";
+  border-radius: 8px;
+  background: #00acdc;
+  height: 4px;
+  width: 37px;
```
Olexandr6, line 65:
```suggestion
  width: 100%;
```

Olexandr6 (changes requested):
Almost done🔥 Check the video review - https://www.loom.com/share/52d96337bc9a468f912920465dfb820f

src/style.css
```diff
@@ -1 +1,82 @@
...
+.moyo_logo {
+  width: 40px;
+  height: 40px;
+  margin: 10px 0;
+}
+
+.navigation {
+  display: flex;
+  background-color: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
```
Olexandr6, line 31: You should apply `box-shadow` and `background-color` for the whole header, not only for navigation:
image

src/style.css
```diff
@@ -1 +1,82 @@
...
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: stretch;
+  margin: auto;
+  padding: 0 50px;
+  height: 100%;
+}
```
Olexandr6, line 20: These styles are useless here, try to delete them and check
```suggestion
.header {
  display: flex;
  justify-content: space-between;
  padding: 0 50px;
}
```

src/style.css
```diff
@@ -1 +1,82 @@
...
+
+.nav__list {
+  padding: 0;
+  margin: 0;
+  list-style: none;
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+}
```
Olexandr6, line 42: You already applied space-between in `header`, so it's not needed here:
```suggestion
.nav__list {
  padding: 0;
  margin: 0;
  list-style: none;
  display: flex;
}
```

src/style.css
```diff
@@ -1 +1,82 @@
...
+  height: 100%;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+}
```
Olexandr6, line 70: You can combine these selectors

src/style.css
```diff
@@ -1 +1,82 @@
...
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+}
+
+.nav__item {
+  display: block;
+  margin-right: 20px;
+  height: 100%;
+}
```
Olexandr6, line 48: Why do you use `height: 100%;` here? It's not needed everywhen, check other selectors

Olexandr6 (changes requested):

src/style.css
```diff
@@ -1 +1,73 @@
...
+
+.is-active::after {
+  display: block;
+  position: absolute;
+  content: "";
+  border-radius: 8px;
+  background: #00acdc;
+  height: 4px;
+  width: 100%;
```
VinnieJ-2k20, line 70: if you use `left: 0;` it's better to use `right: 0;` as well

VinnieJ-2k20 (approved):

PR #2753, opened 2023-02-23
Implemented moyo header

Titech0:
[DEMO LINK](https://titech0.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://titech0.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,12 +3,110 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Moyo header</h1>
+
+  <body class="body">
+    <header class="header">
+      <a
+        href="/"
+        class="logo__link"
+      >
+        <img
+          class="logo"
+          src="./images/logo.png"
+          alt="./images/logo_copy.png"
+        >
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+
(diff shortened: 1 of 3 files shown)


vlad-demchuk (changes requested):
let's try to make header 100% width

src/index.html
```diff
@@ -3,12 +3,110 @@
...
+        <img
+          class="logo"
+          src="./images/logo.png"
+          alt="./images/logo_copy.png"
+        >
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+
```
vlad-demchuk, line 27: no need to add an empty blank between parent and child

src/style.css
```diff
@@ -1 +1,93 @@
...
+}
+
+.logo__link {
+  display: flex;
+}
+
+.nav {
+  display: flex;
+  align-items: stretch;
+  align-self: stretch;
```
vlad-demchuk, line 40: `stretch` is a default value

src/style.css
```diff
@@ -1 +1,93 @@
...
+
+  list-style-type: none;
+}
+
+.item__link {
+  display: flex;
+  align-items: center;
+  position: relative;
+  text-decoration: none;
+  height: 60px;
```
vlad-demchuk, line 67: don't hardcode height, use paddings instead

src/style.css
```diff
@@ -1 +1,93 @@
...
+
+  font-size: 12px;
+  line-height: 14px;
+  font-weight: 500;
+  color: #000;
+}
+
+.item__link:hover {
+  color: #00acdc;
```
vlad-demchuk, line 76: use variables for repeatable values

vlad-demchuk (changes requested):

src/style.css
Titech0: Dont know is this is right way to solve this problem, just trying my best
`.item__link {
  display: flex;
  align-items: center;
  position: relative;
  text-decoration: none;
  
  padding: calc(30px - 1em / 2) 0;

  font-size: 12px;
  line-height: 1em;
  font-weight: 500;
  color: #000;
}`

YegorKorenkov: Yes, it's correct

PR #2777, opened 2023-02-28
done first try

VladyslavKucher:
https://VladyslavKucher.github.io/layout_moyo-header/
https://VladyslavKucher.github.io/layout_moyo-header/report/html_report/


src/index.html
```diff
@@ -7,8 +7,55 @@
...
+      <nav class="box text active">
+        Tablets
+      </nav>
+      <nav class="box text active">
+        Gadgets
+      </nav>
+      <nav 
+      class="box text active"
+      data-qa="hover"
+      >
```
qoa11a, line 35:
```suggestion
      <nav 
        class="box text active"
        data-qa="hover"
      >
```
fix code style everywhere

src/style.css
```diff
@@ -1 +1,89 @@
...
+  top: 18.5px;
+  display: block;
+  border-bottom: #00acdc solid 4px;
+  border-radius: 8px;
+  opacity: 0;
+}
+
+.active:active::after {
+  opacity: 1;
+}
```
qoa11a, line 75: you shouldn't style only link with active class. Every link should have styles on hover but not on click in this task. Class active is only for tests

qoa11a (changes requested):
don't name classes using camelCase
don't forget to use required font-family

src/style.css
```diff
@@ -1 +1,84 @@
...
+
+.box {
+  padding-top: 15px;
+  padding-bottom: 15px;
+  margin-top: auto;
+  margin-bottom: auto;
+  margin-right: 20px;
+  display: inline-block;
+  font-family: "Roboto", sans-serif;
```
tetiana-ivanchenko, line 21: move it to the html selector

src/style.css
```diff
@@ -1 +1,84 @@
...
+  top: 50%;
+  transform: translateY(-50%);
+  display: flex;
+  flex-grow: 1;
+  margin-left: 50px;
+}
+
+.box:hover {
+  cursor: pointer;
+  color: #00acdc;
```
tetiana-ivanchenko, line 54: move repeatable colors in the variables

tetiana-ivanchenko (changes requested):

PR #2802, opened 2023-03-07
add navigation menu

Zibi95:
[DEMO LINK](https://Zibi95.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://Zibi95.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,13 +2,55 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap">
     <link rel="stylesheet" href="./style.css">
+    <title>Moyo header</title>
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img
+        src="./images/logo.png"
+        alt="Moyo logo"
+        width="40"
+        height="40"
+        class="header__img_logo"
+      >
+      <nav class="header__nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__item_link is-active">apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="nav__item_link">samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__item_link">smartphones</a>
+          </li>
+          <li class="nav__item" data-qa="hover">
+            <a href="#laptop&computers" class="nav__item_link">
+              laptops & computers
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__item_link">gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#tablets" class="nav__item_link">tablets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#photo" class="nav__item_link">photo</a>
+          </li>
+          <li class="nav__item">
+            <a href="#video" class="nav__item_link">video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -17,48 +17,54 @@
   </head>
   <body>
     <header class="header">
-      <img
-        src="./images/logo.png"
-        alt="Moyo logo"
-        width="40"
-        height="40"
-        class="header__img_logo"
-      >
+      <a href="#home">
+        <img
+          src="./images/logo.png"
+          alt="Moyo logo"
+          width="40"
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -1,14 +1,67 @@
...
+    <header class="header">
+      <img
+        src="./images/logo.png"
+        alt="Moyo logo"
+        width="40"
+        height="40"
+        class="header__img_logo"
+      >
+
+      <nav class="header__nav">
```
Radoslaw-Czerniawski, line 28: If you are trying to use BEM methodology here, you have to specify a separate class for a block `class="header__nav nav"`. I would personally do something like this here `class="nav"`

src/index.html
```diff
@@ -1,14 +1,67 @@
...
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__item_link is-active">apple</a>
```
Radoslaw-Czerniawski, line 31: In BEM methodology every element is connected to the block. In this case your block is the nav tag. I would change the name of the classes from `nav__item_link` to `nav__link`

src/index.html
```diff
@@ -1,14 +1,67 @@
...
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__item_link">samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__item_link">smartphones</a>
+          </li>
+
+          <li class="nav__item" data-qa="hover">
```
Radoslaw-Czerniawski, line 42: According to the requirements data-qa attribute should be given to link tag not list item.

src/style.css
```diff
@@ -1 +1,60 @@
...
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  padding: 0 50px;
+}
+
+.header__img_logo {
+  margin: 10px 0;
```
Radoslaw-Czerniawski, line 21: There is no need for margin here.

src/style.css
```diff
@@ -1 +1,60 @@
...
+
+.nav__list {
+  font-size: 12px;
+  font-weight: 500;
+  text-transform: uppercase;
+  list-style: none;
+  display: flex;
+}
+
+.nav__item:nth-of-type(n + 2) {
```
Radoslaw-Czerniawski, line 35: You could also do here `.nav__item:not(:first-of-type)` but your solution still works

src/style.css
```diff
@@ -1 +1,60 @@
...
+  margin-left: 20px;
+}
+
+.nav__item_link {
+  display: block;
```
Radoslaw-Czerniawski, line 40: No need for this.

src/style.css
```diff
@@ -1 +1,60 @@
...
+  text-decoration: none;
+  color: black;
+}
+
+.nav__item_link:hover,
+.is-active {
+  color: var(--hover-active-color);
+}
+
+.is-active::after {
```
Radoslaw-Czerniawski, line 52: Generally you want your after element to be positioned absolute according to his relative as not to take space in the block. I would personally set first `nav__item` as position relative and set `::after` on that `nav__item` with position absolute

src/style.css
```diff
@@ -1 +1,60 @@
...
+  content: "";
+  display: block;
```
Radoslaw-Czerniawski, line 54: This then will be redundant.

src/style.css
```diff
@@ -1 +1,60 @@
...
+  position: relative;
+  bottom: 4px;
+  height: 4px;
+  border-radius: 2px;
+  background-color: var(--hover-active-color);
+}
```
Radoslaw-Czerniawski, line 60: What do you think about such structure?

```
.nav__item {
  height: 64px;
  position: relative;
}

.nav__item:first-of-type {
  margin-left: 20px;
}

.nav__item_link {
  height: 100%;
  display: flex;
  align-items: center;
  text-decoration: none;
  color: black;
}

.nav__item_link:hover {
  color: var(--hover-active-color);
}

.is-active {
  color: var(--hover-active-color);
}

.nav__item_link.is-active::after {
  content: "";
  position: absolute;
  left: 0;
  right: 0;
  bottom: 0;
  height: 4px;
  border-radius: 2px;
  background-color: var(--hover-active-color);
}
```

Radoslaw-Czerniawski (changes requested):
Some more comments on this one. It looks fine apart from mistakes like missing `#home` link on logo. Please correct according to review and ask if you have any questions.

src/style.css
Zibi95: less math less problems 😂

src/style.css
Zibi95: `.nav__item {
  height: 64px;
  position: relative;
}`
height should be 60 px

`.nav__item:first-of-type {
  margin-left: 20px;
}`
I think you mean here :not(first-of-type)

Yea I think position absolute is better. I don't have to take into account this ::after height. Thanks for the insights!
I added external reset.css file as you suggested with [Meyer reset](https://meyerweb.com/eric/tools/css/reset/) but ES lint is giving me hard time with that 😂

Radoslaw-Czerniawski: - Yup now I see the design in Figma says 60px
- Also I ment something like
```
.nav__item {
  height: 64px;
  position: relative;
  margin-left: 20px;
}

.nav__item:first-of-type /* or :first-child */ {
  margin-left: 0;
}
```
My bad with the advice, reviewing css is always challenging as you are not actually seeing the results 😅
- I would import the reset pretty similarly
- also for the future if you have more than 2 css files you may create one that only imports the others then you only insert that one to the html -> however that's my personal preference

Radoslaw-Czerniawski (approved):
Solid work!

mkyloa (approved):

PR #2823, opened 2023-03-09
add task solution

marshalek37:
[DEMO LINK](https://marshalek37.github.io/layout_moyo-header/)
 [TEST REPORT LINK](https://marshalek37.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,6 +1,6 @@
 :root {
-  --link-color: #00acdc;
-  --link-color2: #000;
+  --link-moyoBlue: #00acdc;
+  --link-moyoBlack: #000;
 }
 
 html {
@@ -42,7 +42,7 @@ body {
   padding: 23px 0;
   font-size: 12px;
   font-weight: 500;
-  color: var(--link-color2);
+  color: var(--link-moyoBlack);
   text-decoration: none;
   text-transform: uppercase;
   white-space: nowrap;
@@ -53,16 +53,16 @@ body {
 }
 
 .navLink:hover {
-  color: var(--link-color);
+  color: var(--link-moyoBlue);
   cursor: pointer;
 }
 
 .is-active {
-  color: var(--link-color);
+  color: var(--link-moyoBlue);
 }
 
 .is-active::after {
-  background-color: #00acdc;
+  background-color: var(--link-moyoBlue);
   border-radius: 2px;
   display: block;
   height: 4px;


michalcywka (changes requested):
Please look at the font weight in the mockup ;)

src/style.css
```diff
@@ -1 +1,73 @@
-/* Styles go here */
+@font-face {
+  font-family: "Roboto";
+  src: url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
```
michalcywka, line 3: Try to add this font in `link` tag in HTML to see if it helps :)

michalcywka (approved):

michalcywka (changes requested):
.

src/style.css
```diff
@@ -1 +1,68 @@
...
+.is-active {
+  color: rgb(0, 183, 255);
+}
+
+.is-active::after {
+  background-color: rgb(0, 183, 255);
+  border-radius: 8px;
+  display: block;
+  height: 4px;
+  width: 37px;
```
solaryasha, line 51:
```suggestion
  width: 100%;
```

src/style.css
```diff
@@ -1 +1,68 @@
...
+  bottom: 0;
+  margin-bottom: 0;
```
solaryasha, line 53: Why do you need this rule?

solaryasha: Let's remove this rule

src/style.css
```diff
@@ -1 +1,68 @@
...
+  margin: 0;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+}
+
+* {
+  box-sizing: border-box;
+}
```
solaryasha, line 68: Let's move these rules to the top of the file!

solaryasha (changes requested):

michalcywka (changes requested):
There is an unresolved comment from Volodymyr, please have a look.

src/style.css
```diff
@@ -1 +1,67 @@
...
+  display: flex;
+  position: relative;
+}
+
+.navLink {
+  padding: 23px 0;
+  font-size: 12px;
+  font-weight: 500;
+  color: rgb(41, 41, 41);
+  cursor: pointer;
```
d-wasilewski, line 40: `cursor: pointer` should be set in a :hover pseudo-class, not in the element itself (we need it only when the element is hovered)

src/style.css
```diff
@@ -1 +1,67 @@
...
+.navItem {
+  display: flex;
+  position: relative;
+}
+
+.navLink {
+  padding: 23px 0;
+  font-size: 12px;
+  font-weight: 500;
+  color: rgb(41, 41, 41);
```
d-wasilewski, line 39: Please use hexadecimal color values from figma

src/style.css
```diff
@@ -1 +1,67 @@
...
+  margin-right: 20px;
+}
+
+.navLink:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+}
```
d-wasilewski, line 56: This can be simplified :)

src/style.css
```diff
@@ -1 +1,67 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
+.header {
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+}
```
d-wasilewski, line 17: You might have missed it, but header should have a "shadow". Check figma

d-wasilewski (changes requested):

src/style.css
```diff
@@ -1 +1,73 @@
-/* Styles go here */
+:root {
+  --link-color: #00acdc;
+  --link-color2: #000;
```
d-wasilewski, line 3: try to give them more descriptive names

d-wasilewski (changes requested):
Great job!

d-wasilewski (approved):

PR #2835, opened 2023-03-11
add task solution

vpanch2:
[DEMO LINK] https://vpanch2.github.io/layout_moyo-header/

[TEST REPORT LINK] https://vpanch2.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,16 +15,37 @@
       </a>
       <nav class="nav">
         <ul class="list">
-          <li class="nav-item"><a href="#apple" class="nav-link is-active">Apple</a></li>
-          <li class="nav-item"><a href="#samsung" class="nav-link">Samsung</a></li>
-          <li class="nav-item"><a href="#smartphones" class="nav-link">Smartphones</a></li>
+          <li class="nav-item">
+            <a href="#apple" class="nav-link is-active">Apple</a>
+          </li>
+
+          <li class="nav-item">
+            <a href="#samsung" class="nav-link">Samsung</a>
+          </li>
+
+          <li class="nav-item">
+            <a href="#smartphones" class="nav-link">Smartphones</a>
+          </li>
+
           <li class="nav-item" data-qa="hover">
             <a href="#laptops&computers" class="nav-link">Laptops &amp; computers</a>
           </li>
-          <li class="nav-item"><a href="#gadgets" class="nav-link">Gadgets</a></li>
-          <li class="nav-item"><a href="#tablets" class="nav-link">Tablets</a></li>
-          <li class="nav-item"><a href="#photo" class="nav-link">Photo</a></li>
-          <li class="nav-item"><a href="#video" class="nav-link">Video</a></li>
+
+          <li class="nav-item">
+            <a href="#gadgets" class="nav-link">Gadgets</a>
+          </li>
+
+          <li class="nav-item">
+            <a href="#tablets" class="nav-link">Tablets</a>
+          </li>
+
+          <li class="nav-item">
+            <a href="#photo" class="nav-link">Photo</a>
+          </li>
+
+          <li class="nav-item">
+            <a href="#video" class="nav-link">Video</a>
+          </li>
         </ul>
       </nav>
     </header>
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,24 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,86 @@
...
+.logo {
+  width: 40px;
+  height: 40px;
+  display: block;
+}
+
+.nav {
+  display: flex;
+  align-items: center;
+  flex-flow: row;
```
agniya-i, line 28: looks redundant, it's `row` by default

src/index.html
```diff
@@ -9,6 +9,24 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img src="images/logo.png" alt="logo">
+      </a>
+      <nav class="nav">
+        <ul class="list">
+          <li class="nav-item"><a href="#apple" class="nav-link is-active">Apple</a></li>
```
agniya-i, line 18:
```suggestion
          <li class="nav-item">
             <a href="#apple" class="nav-link is-active">Apple</a>
          </li>
```
format code in this way to make code more readable

src/style.css
```diff
@@ -1 +1,86 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@1,200&family=Roboto:wght@500&display=swap");
+
+* {
+  margin: 0;
+  padding: 0;
```
agniya-i, line 5: better to not reset these values in `*`

agniya-i (changes requested):

src/style.css
polosanya: Not fixed. Don't reset margins and padding using * selector. Specify the exact tags you want to reset
image
```
body,
nav, 
list {
  margin: 0;
  padding: 0;
}
```

src/style.css
```diff
@@ -1 +1,85 @@
...
+  box-sizing: border-box;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  padding-left: 50px;
+  padding-right: 50px;
```
polosanya, line 13: You can use shorthand property instead
https://developer.mozilla.org/en-US/docs/Web/CSS/padding#syntax
```suggestion
  padding: 0 50px;
```

src/style.css
```diff
@@ -1 +1,85 @@
...
+
+.nav-item {
+  display: flex;
+  justify-content: center;
+}
+
+.nav-link {
+  padding: 23px 0;
+  display: flex;
+  font-family: "Roboto", Helvetica, sans-serif;
```
polosanya, line 48: You can set the font for the whole page and just inherit it where needed

src/style.css
```diff
@@ -1 +1,85 @@
...
+}
+
+.nav-link.is-active::after {
+  position: absolute;
+  content: "";
+  background-color: #00acdc;
+  border-radius: 2px;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
polosanya, line 78: Pay attention to CSS properties order

polosanya (changes requested):
Almost done!

src/style.css
```diff
@@ -1 +1,87 @@
...
+@import url("https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@1,200&family=Roboto:wght@500&display=swap");
+
+body,
+nav,
+.list {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: "Roboto", Helvetica, sans-serif;
+}
```
agniya-i, line 10:
```suggestion
* { 
  box-sizing: border-box;
}

body,
.nav,
.list {
  margin: 0;
  padding: 0;
}

html {
  font-family: "Roboto", Helvetica, sans-serif;
}
```
Let's improve and structurize a little bit) 

1. You should not set properties in `*` selector because it can affect the page performance, but in the case of `box-sizing `it's ok, it's better when all elements have same `box-sizing`
2. Better to set `font-family` to `html`

src/style.css
```diff
@@ -1 +1,87 @@
...
+.logo {
+  width: 40px;
+  height: 40px;
+  display: block;
+}
+
+.nav {
+  display: flex;
+  align-items: center;
+  margin: 0;
```
agniya-i, line 30: pay attention that you already reset margin and padding above, so do not repeat yourself

agniya-i (changes requested):

polosanya (approved):
Well done!

PR #2845, opened 2023-03-14
solution

Kinderwagen:


src/style.css
```diff
@@ -1 +1,67 @@
...
+
+.is-active {
+  color: rgba(0, 172, 220, 1);
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  position: relative;
+  top: 18.5px;
```
OlEzhA0, line 63: here is should be
```
position: absolute;
bottom: 0;
left: 0;
width: 100%;
```

and add `position: relative;` to `.is-active`

OlEzhA0 (changes requested):

src/style.css
```diff
@@ -1 +1,70 @@
...
+}
+
+.nav__item {
+  list-style: none;
+  margin-left: 20px;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
```
etojeDenys, line 38:
```suggestion
.nav__item {
  list-style: none;
}

.nav__item:not(:first-child) {
  margin-left: 20px;
}
```

etojeDenys (changes requested):
small improvement

VitaliyBondarenko1982 (approved):

PR #2853, opened 2023-03-15
moyo-header

yevheniia-biliaieva:
https://yevheniia-biliaieva.github.io/layout_moyo-header/
https://yevheniia-biliaieva.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,121 @@
-/* Styles go here */
+/*Обнуление*/
+*,
+html,
+body,
+div,
+nav {
+  padding: 0;
+  margin: 0;
+  border: 0;
+}
+
+nav,
+footer,
+header,
+aside {
+  display: block;
+}
+
+a {
+  color: #000;
+  text-decoration: none;
+}
+
+a:visited,
+a:hover,
+a:focus,
+a:active {
+  text-decoration: none;
+  outline: none;
+}
+
+ul li {
+  list-style: none;
+}
+
+/*------------------*/
+
+@font-face {
+  font-family: "Roboto";
+  src: url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,121 @@
...
+
+@font-face {
+  font-family: "Roboto";
+  src: url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+  font-style: normal;
+  font-weight: 500;
+}
+
+html {
+  box-sizing: border-box;
```
etojeDenys, line 46: move it into * selector

src/style.css
```diff
@@ -1 +1,121 @@
...
+}
+
+body {
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 500;
```
etojeDenys, line 52: and these lines move to html selector

src/style.css
```diff
@@ -1 +1,121 @@
...
+
+.is-active {
+  color: #00acdc;
+  position: relative;
+}
+
+.nav__link.is-active::after {
+  position: absolute;
+  content: "";
+  background-color: #00acdc;
```
etojeDenys, line 110: we can move this color into a variable and then reuse it

etojeDenys (changes requested):
almost done

src/style.css
IvanFesenko: Not fixed

src/style.css
```diff
@@ -1 +1,122 @@
...
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+  padding-left: 0;
+}
```
IvanFesenko, line 101: Redudant

PR #2859, opened 2023-03-16
Solution

OlhaKissa:
[DEMO LINK](https://olhakissa.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://olhakissa.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,72 @@
...
+  position: relative;
+  color: #000;
+  text-decoration: none;
+  font-size: 12px;
+  text-transform: uppercase;
+  padding: 23px 0;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
etojeDenys, line 57: let`s move this color to variable and then reuse it

src/style.css
```diff
@@ -1 +1,72 @@
...
+  margin: 0;
+}
+
+.nav__item {
+  margin-right: 20px;
+}
+
+.nav__item:last-child {
+  margin-right: 0;
+}
```
etojeDenys, line 43:
```suggestion
.nav__item:not(:last-child) {
  margin-right: 20px;
}
```

etojeDenys (changes requested):
GJ

PR #2867, opened 2023-03-17
Solution

offsset:
- [DEMO LINK](https://offsset.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://offsset.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,59 @@
-/* Styles go here */
+* {
+	font-family: 'Robot';
+	font-size: 12px;
+	margin: 0;
+	padding: 0;
+	box-sizing: border-box;
+}
+
+header {
+	width: 1200px;
+	min-width: 1024px;
+	height: 60px;
+	margin: auto;
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	padding: 0 50px;
+	box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+
+
+.nav__list {
+	display: flex;
+	list-style: none;
+
+}
+
+.nav__link {
+	display: block;
+	text-decoration: none;
+	padding:  23px 0;
+	margin-left: 20px;
+	color: black;
+	line-height: 14px;
+	text-transform: uppercase;
+
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,59 @@
-/* Styles go here */
+* {
+	font-family: 'Robot';
+	font-size: 12px;
```
VitaliyBondarenko1982, line 3: move font family and font size to html selector
Robot => Roboto
Remember to use fallback fonts - alternative font-family in case the main one doesn't work [like this](https://www.w3schools.com/cssref/pr_font_font-family.asp)

src/style.css
```diff
@@ -1 +1,59 @@
...
+	margin: 0;
+	padding: 0;
+	box-sizing: border-box;
+}
```
VitaliyBondarenko1982, line 7: reset margin and padding  for tags it is necessary
```suggestion
	box-sizing: border-box;
}
```

src/style.css
```diff
@@ -1 +1,59 @@
...
+
+header {
+	width: 1200px;
+	min-width: 1024px;
+	height: 60px;
```
VitaliyBondarenko1982, line 12: do not hardcode height. make for links vertical paddings which will make necessary height

VitaliyBondarenko1982, line 12:
```suggestion
```

src/style.css
```diff
@@ -1 +1,59 @@
...
+	margin: auto;
```
VitaliyBondarenko1982, line 13:
```suggestion
```

src/style.css
```diff
@@ -1 +1,59 @@
...
+	display: flex;
+	list-style: none;
+
+}
+
+.nav__link {
+	display: block;
+	text-decoration: none;
+	padding:  23px 0;
+	margin-left: 20px;
```
VitaliyBondarenko1982, line 33: better to set margin left for nav_item. and do not set for first-child

VitaliyBondarenko1982 (changes requested):
check your demo links.

PR #2875, opened 2023-03-19
solution

VadymKapeliusnyi:
- [DEMO LINK](https://VadymKapeliusnyi.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://VadymKapeliusnyi.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,61 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,61 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
```
vlad-demchuk, line 3: image

src/style.css
```diff
@@ -1 +1,61 @@
...
+.header__logo {
+  margin-top: 3px;
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+}
+
+.nav__list a {
```
vlad-demchuk, line 29: do not use tags for styling
do not increase specificity

vlad-demchuk: just add classes and use them

vlad-demchuk (changes requested):

PR #2883, opened 2023-03-22
add task solution

alexmik71:
https://alexmik71.github.io/layout_moyo-header/

https://alexmik71.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,72 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,72 @@
...
+.nav__item:first-child {
+  margin-left: 0;
+}
+
+.nav__link {
+  position: relative;
+  display: flex;
+  align-items: center;
+  height: 60px;
+  font-family: "Roboto", sans-serif;
```
IvanFesenko, line 45: Move `font-family` to `html` selector

IvanFesenko (changes requested):
Over all all good, some small fix required

PR #2897, opened 2023-03-26
add task solution

akozlovska:
[DEMO LINK](https://akozlovska.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://akozlovska.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,56 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.header {
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+  --header-links-color: #00acdc;
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+  margin: 10px 0;
+}
+
+.nav__list {
+  list-style-type: none;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,56 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
```
tetiana-ivanchenko, line 3: you shouldn't reset indentation for the * selector

src/style.css
```diff
@@ -1 +1,56 @@
...
+}
+
+.nav__list {
+  list-style-type: none;
+  display: flex;
+}
+
+.nav__list-link {
+  display: block;
+  padding: 22.5px 0;
```
tetiana-ivanchenko, line 27: use integer
image

src/style.css
```diff
@@ -1 +1,56 @@
...
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
```
tetiana-ivanchenko, line 29: move it to the html selector

tetiana-ivanchenko (changes requested):
fix tests, you have little bit different indentation for the active link

PR #2909, opened 2023-03-29
add task solution

shelyaa:
https://shelyaa.github.io/layout_moyo-header/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,66 @@
-/* Styles go here */
+body {
+  margin: 0;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+}
+
+.nav__link {
+  text-transform: uppercase;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-decoration: none;
+  color: rgba(0, 0, 0, 1);
+  --hover-color: rgba(0, 172, 220, 1);
+  display: block;
+}
+
+.nav__link:hover {
+  color: var(--hover-color);
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+  margin: 10px 50px;
+}
+
+.nav__item {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,66 @@
...
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  color: rgba(255, 255, 255, 1);
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav__list {
```
vlad-demchuk, line 43: let's change classes order a bit
it will be better to understand
```
header
logo
nav__list
nav__item
nav__link

File: src/style.css [hunk 2]
```diff
@@ -1 +1,66 @@
...
+  display: flex;
+}
+
+.last {
```
vlad-demchuk, line 47: what is `.last`?
use` :last-child `CSS pseudo-class

vlad-demchuk (changes requested):
test report link doesn't work
try to re-deploy your work

anstsot (changes requested):
Looks good but the report isn't working. Try to write in `fe_chat` and mentors will help you with this problem

src/style.css
```diff
@@ -1 +1,66 @@
...
+}
+
+.nav__link {
+  text-transform: uppercase;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-decoration: none;
+  color: rgba(0, 0, 0, 1);
+  --hover-color: rgba(0, 172, 220, 1);
```
OleziO, line 39: Use variables for all repeatable colors and move all variables to `:root` selector

OleziO (changes requested):

VitaliyBondarenko1982 (approved):

PR #2912, opened 2023-03-30
moyo header

NazarVasylyshyn:
- [DEMO LINK](https://NazarVasylyshyn.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://NazarVasylyshyn.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,66 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.body {
+  margin: 0;
+  color: #00acdc;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav {
+  margin: 0;
+  padding: 0;
+}
+
+.logo {
+  margin: 10px 0;
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,66 @@
...
+  text-decoration: none;
+  text-transform: uppercase;
+  font-size: 12px;
+  padding: 23px 0;
+  color: #000;
+}
+
+.nav__link:hover,
+.is-active {
+  color: #00acdc;
```
slavanosatskyi, line 55: you can also move same colors to css variables

slavanosatskyi (changes requested):
Very nice 👍 just few comments from me

src/style.css
```diff
@@ -1 +1,62 @@
...
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav,
+.nav__list {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
```
vlad-demchuk, line 25: https://github.com/mate-academy/layout_moyo-header/pull/2912#discussion_r1153406210

src/style.css
```diff
@@ -1 +1,62 @@
-/* Styles go here */
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.body {
+  margin: 0;
+  --best-color: #00acdc;
```
vlad-demchuk, line 9: move the variable to `:root` pseudo class

vlad-demchuk (changes requested):

src/style.css
NazarVasylyshyn: I do not understand what do you mean. I did it again but I don't know if it's right

OleziO (approved):

PR #2921, opened 2023-04-04
add task solution

vassavka:
https://vassavka.github.io/layout_moyo-header/
https://vassavka.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,65 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,65 @@
...
+:root {
+  --color-cyan: #00acdc;
+}
+
+body {
+  margin: 0;
+}
+
+.header {
+  height: 60px;
```
YegorKorenkov, line 12: try to not hardcode height

YegorKorenkov (approved):
Looks good

PR #2926, opened 2023-04-05
moyo header1

mralexlihuta:
- [DEMO LINK](https://mralexlihuta.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://mralexlihuta.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,72 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  padding-right: 30px;
+}
+
+.nav {
+  margin-right: 20px;
+}
+
+.logo-site {
+  height: 40px;
+  margin-left: 50px;
+  margin-top: 10px;
+}
+
+.nav__list {
+  list-style: none;
+  display: flex;
+
+}
+
+.nav__item {
+  margin-right: 20px;
+  font-style: normal;
+  text-transform: uppercase;
+  font-size: 12px;
+  letter-spacing: 0;
+  font-weight: 500;
+}
+
+.nav__item:nth-last-child(1) {
+  margin-right: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,72 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+}
```
etojeDenys, line 8:
```suggestion
html {
  font-family: "Roboto", sans-serif;
}
```

src/style.css
```diff
@@ -1 +1,72 @@
...
+  font-style: normal;
+  text-transform: uppercase;
+  font-size: 12px;
+  letter-spacing: 0;
+  font-weight: 500;
+}
+
+.nav__item:nth-last-child(1) {
+  margin-right: 0;
+}
```
etojeDenys, line 43:
```suggestion
.nav__item:not(:last-child) {
  margin-right: 20px;
}
```

src/style.css
```diff
@@ -1 +1,72 @@
...
+  width: 100%;
+  right: 0;
+  bottom: 0;
+  height: 4px;
+  border-radius: 5px;
+  background-color: #00adcd;
+}
+
+.nav__link:hover {
+  color: #00adcd;
```
etojeDenys, line 71: let's move this color to some var and reuse it everywhere

etojeDenys (changes requested):

src/style.css
mralexlihuta: can't pass test with that

src/style.css
```diff
@@ -1 +1,76 @@
-/* Styles go here */
+* {
+  margin: 0;
+  padding: 0;
```
loralevitska, line 3: Only `box-sizing: border-box` allowed for this selector. Move it to body

src/style.css
loralevitska: acepted

loralevitska (approved):

PR #2928, opened 2023-04-05
add task solution

obadyda:
https://github.com/obadyda.github.io/layout_moyo-header/
https://github.com/obadyda.github.io/layout_moyo-header/report/html_report


src/style.css
```diff
@@ -1 +1,84 @@
-/* Styles go here */
+
+* {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  font-style: normal;
+  font-display: swap;
+  line-height: 14px;
```
etojeDenys, line 9: move to html selector

etojeDenys (changes requested):
fix your links they do not work

PR #2931, opened 2023-04-06
Develop

VolodymyrPst:


src/style.css
```diff
@@ -1 +1,79 @@
 /* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Comfortaa:wght@300;400;700&family=Inter:wght@300;400;700;800&family=Roboto+Condensed:ital,wght@1,300;1,400;1,700&family=Roboto:wght@500&display=swap");
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  outline: none;
```
slavanosatskyi, line 10: body doesn't have outline

src/style.css
```diff
@@ -1 +1,79 @@
...
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  margin: auto;
+  align-items: center;
+  max-width: 1200px;
+  height: 60px;
```
slavanosatskyi, line 20: try to avoid setting height, let its content do define height of the component

src/style.css
```diff
@@ -1 +1,79 @@
...
+.link__logo {
+  text-decoration: none;
+  margin-left: 50px;
+  width: 40px;
+  height: 40px;
+}
+
+.nav {
+  height: 100%;
+  margin-right: 50px;
```
slavanosatskyi, line 32: do not use external geometry on block level

src/style.css
```diff
@@ -1 +1,79 @@
...
+  display: flex;
+  align-items: center;
+  border-bottom: 4px solid white;
+  border-radius: 4px;
+  padding-top: 4px;
+  text-decoration: none;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
slavanosatskyi, line 73: define variable for reused colors

slavanosatskyi (changes requested):

PR #2936, opened 2023-04-07
add task solution

Dolnys:
[HEADER](https://Dolnys.github.io/layout_moyo-header/)

[TEST](https://Dolnys.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,72 @@
 <!doctype html>
 <html lang="en">
   <head>
-    <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <meta
+    charset="UTF-8">
+    <meta
+    name="viewport"
+          content="width=device-width, user-scalable=no,
+          initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <meta
+    http-equiv="X-UA-Compatible"
+    content="ie=edge">
     <title>Moyo header</title>
-    <link rel="stylesheet" href="./style.css">
+    <link
+    rel="stylesheet"
+    href="./style.css">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -1,14 +1,72 @@
...
+      alt="moyo-logo"
+      >
+      </a>
+      <nav class="navigation">
+        <ul class="navigation--list">
+          <li class="nav-item">
+            <a class="nav-item-link is-active" href="#">Apple</a>
+          </li>
+          <li class="nav-item">
+            <a class="nav-item-link" href="#Samsung">Samsung</a>
```
choeqq, line 47: `href` attribute is typically lowercased, it doesn't matter in this particular case, but I would keep that in mind, when you'll start working with HTTP and routing

src/style.css
```diff
@@ -1 +1,64 @@
 /* Styles go here */
+body {
```
choeqq, line 2: I would probably provide more generous CSS styles reset like:
```
*, *::after, *::before {
  box-sizing: border-box;
  margin: 0;
  padding: 0;
}
```
but that's a good start 😄 

src/index.html
```diff
@@ -1,14 +1,72 @@
...
+    >
+    <link
+    href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+    rel="stylesheet"
+    >
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="logo-link">
```
choeqq, line 34: I am thinking of wrapping this `<a />` tag with `<div>` so it is easier to move around, what do you think?
Typically when we use flexbox we want to operate on block elements

choeqq (approved):

choeqq (approved):

src/index.html
Dolnys: Hey! It sounds like a reasonable idea. Thank you for the review. I appreciate your suggestions, and I have implemented all the proposed changes.

mvjl000 (approved):

PR #2953, opened 2023-04-10
add moyo header

cephy77:
[DEMO LINK](https://cephy77.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://cephy77.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,18 +1,22 @@
 # Moyo header
+
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_moyo-header/)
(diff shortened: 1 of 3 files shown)


naziamoff (approved):
nailed it 💅 🔥 

The only thing - I would have added some spacing between blocks because it is now quite difficult to read and distinguish between `nav__item`s. I mean would be better to make it this way:

```
        <li class="nav__item">
            <a href="#" class="nav__link">
              Gadgets
            </a>
          </li>
          
          <li class="nav__item">
            <a href="#" class="nav__link">
              Tablets
            </a>
          </li>
```

PR #2963, opened 2023-04-11
add task solution

prostonadobobb:
[DEMO LINK](https://prostonadobobb.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://prostonadobobb.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,53 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+
+    <header class="header">
+      <div class="container">
+        <div class="header-body">
+          <a href="#home" class="logo-link">
+            <img
+              src="./images/logo.png"
+              alt="logo"
+              class="logo"
+            >
+          </a>
+
+          <nav class="nav">
+            <ul class="menu-list">
+              <li class="menu-item">
+                <a class="menu-link is-active" href="#">Apple</a>
+              </li>
+              <li class="menu-item">
+                <a class="menu-link" href="#">Samsung</a>
+              </li>
+              <li class="menu-item">
+                <a class="menu-link" href="#">Smartphones</a>
+              </li>
+              <li class="menu-item"
+                  data-qa="hover"
+              >
+                <a class="menu-link" href="#">Laptops & Computers</a>
+              </li>
+              <li class="menu-item">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,53 @@
...
+              </li>
+              <li class="menu-item">
+                <a class="menu-link" href="#">Samsung</a>
+              </li>
+              <li class="menu-item">
+                <a class="menu-link" href="#">Smartphones</a>
+              </li>
+              <li class="menu-item"
+                  data-qa="hover"
+              >
```
Viktor-Kost, line 38: Just write in one line if you have two or less attributes 
```suggestion
              <li class="menu-item" data-qa="hover">
```

src/style.css
```diff
@@ -1 +1,77 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
Viktor-Kost, line 4: it's allowed to use only box-sizing: border-box; inside * selector. If you need to reset margin and padding for other tags just use other selectors

Example:
```
body {
  margin: 0;
}
```

src/style.css
```diff
@@ -1 +1,77 @@
...
+}
+
+.is-active::after {
+  content: " ";
+  position: absolute;
+  border-bottom: 4px solid var(--header-hover-color);
+  bottom: 0;
+  left: 0;
+  right: 0;
+}
```
Viktor-Kost, line 77: pseudo-element ::after is not a border. Just make it as an element
```suggestion
.is-active::after {
  content: "";
  position: absolute;
  width: 100%;
  height: 4px;
  background-color: var(--header-hover-color);
  bottom: 0;
  left: 0;
}
```

src/style.css
```diff
@@ -1 +1,77 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
+:root {
+  --header-hover-color: #00acdc;
```
Viktor-Kost, line 8: it would be better to name something like `accent-color` in case you use this variable not only to the hover effect

Viktor-Kost (changes requested):

src/index.html
```diff
@@ -7,8 +7,53 @@
...
+              src="./images/logo.png"
+              alt="logo"
+              class="logo"
+            >
+          </a>
+
+          <nav class="nav">
+            <ul class="menu-list">
+              <li class="menu-item">
+                <a class="menu-link is-active" href="#">Apple</a>
```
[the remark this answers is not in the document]
vlad-demchuk: not fixed

I know you haven't learnt BEM yet, but just follow suggestions above

nazarmatsevych (approved):
👍

PR #2983, opened 2023-04-16
Develop2

timohahaaa:
https://timohahaaa.github.io/layout_moyo-header/

https://timohahaaa.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -3,39 +3,37 @@ body {
   font-family: "Roboto", sans-serif;
 }
 
-header {
+.header {
+  display: flex;
   width: 100%;
+  justify-content: space-between;
 }
 
 .logo {
-  position: absolute;
   width: 40px;
   height: 40px;
-  left: 50px;
-  top: 10px;
+  margin: 10px 50px;
 }
 
 .nav {
   display: flex;
   justify-content: end;
   height: 60px;
-  left: 0;
-  top: 0;
   background: #fff;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,57 @@
...
+  margin: 10px 50px;
+}
+
+.nav {
+  display: flex;
+}
+
+.nav__link {
+  font-family: "Roboto", sans-serif;
+  display: flex;
```
polosanya, line 24: Add `align-items: center` property to center the text vertically.
Also, position should be relative, so bottom line is absolute to the link

src/style.css
```diff
@@ -1 +1,57 @@
...
+}
+
+.is-active {
+  position: relative;
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  width: 37px;
```
polosanya, line 46: Don't hardcode the width. It should take the whole possible width
```suggestion
  width: 100%;
```

src/style.css
```diff
@@ -1 +1,57 @@
...
+  height: 4px;
+  top: 36px;
+  left: 0;
```
polosanya, line 49:
```suggestion
  bottom: 0
```
Let's try this instead

polosanya (changes requested):
Try to reset the default margin and paddings of ul and body
image
Also, pay attention to the comments. I hope they will help you to pass the tests

PR #2984, opened 2023-04-16
TIC1

timohahaaa:
https://timohahaaa.github.io/layout_moyo-header/report/html_report/

https://timohahaaa.github.io/layout_moyo-header/

I'm so sorry, but I created a new pull request cause the last one had a glitch on my PC!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,40 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans" rel="stylesheet">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@400;700&family=Roboto:wght@400;500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <a href="#home">
+        <img
+          src="images/logo.png"
+          alt="#"
+          class="logo"
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item"><a href="#" class="is-active nav__link">Apple</a></li>
+          <li class="nav__item"><a href="#" class="nav__link">Samsung</a></li>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,65 @@
-/* Styles go here */
+body {
+  margin: 0;
+  padding: 0;
```
Volodymyr-Melnyk-Fron10der, line 3: there is no default padding for body, so there is no need to reset it

src/style.css
```diff
@@ -1 +1,65 @@
...
+  height: 40px;
+  margin: 10px 50px;
+}
+
+.nav {
+  display: flex;
+}
+
+.nav__link {
+  font-family: "Roboto", sans-serif;
```
Volodymyr-Melnyk-Fron10der, line 29: just set it for html selector and inherit it

Volodymyr-Melnyk-Fron10der (changes requested):
Pls try to fix failed report points. If you have any troubles - feel free to ask them in the Slack

src/index.html
```diff
@@ -7,8 +7,40 @@
...
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@400;700&family=Roboto:wght@400;500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home">
+        <img
+          src="images/logo.png"
+          alt="#"
```
polosanya, line 20: The `alt` attribute for the logo image should descriptive. Something like "Moyo logo" would be great

src/style.css
```diff
@@ -1 +1,69 @@
...
+
+.nav {
+  display: flex;
+  margin-right: 30px;
+}
+
+.nav__link {
+  font-family: inherit;
+  position: relative;
+  align-items: center;
```
polosanya, line 33: The 'align-items' property is used with a flex container, but '.nav__link' is not a flex container. Consider removing it, as it does not have any effect on the element

src/style.css
```diff
@@ -1 +1,69 @@
...
+  left: 0;
+  top: 34px;
+}
+
+.nav__item {
+  display: inline-block;
+  list-style-type: none;
+  margin-left: 20px;
+  padding-top: 19px;
+  padding-bottom: 19px;
```
polosanya, line 68: Paddings should be added to the links, not li, so links are clickable for the whole height

polosanya (changes requested):
A few more comments to fix

PR #2987, opened 2023-04-18
add task solution for moyo nav bar

ruslan-ihnatenko:
- [DEMO LINK](https://ruslan-ihnatenko.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ruslan-ihnatenko.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,67 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="main_header" id="home">
+      <div class="list_container">
+        <a class="logo_link" href="#home">
+          <img
+            class="logo"
+            src="images/logo.png"
+            alt="MOYO"
+          >
+        </a>
+        <nav class="main_nav">
+          <ul class="nav__list">
+            <li class="list__item ">
+              <a class="link_list_item is-active" href="/">
+                apple
+              </a>
+            </li>
+            <li class="list__item">
+              <a class="link_list_item" href="/">
+                samsung
+              </a>
+            </li>
+            <li class="list__item">
+              <a class="link_list_item" href="/">
+                smartphones
+              </a>
+            </li>
+            <li class="list__item">
+              <a
+                class="link_list_item"
+                href="/"
+                data-qa="hover"
+              >
+                laptops & computers
+              </a>
+            </li>
+            <li class="list__item">
+              <a class="link_list_item" href="/">
+                gadgets
+              </a>
+            </li>
+            <li class="list__item">
+              <a class="link_list_item" href="/">
+                tablets
+              </a>
+            </li>
+            <li class="list__item">
+              <a class="link_list_item" href="/">
+                photo
+              </a>
+            </li>
+            <li class="list__item">
+              <a class="link_list_item" href="/">
+                video
+              </a>
+            </li>
+          </ul>
+        </nav>
+      </div>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,84 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Roboto , sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+body {
+  margin: 0;
+}
+
+.main_header {
+  display: flex;
+  justify-content: space-between;
+  padding: auto 50px;
+  width: 100%;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav__list {
+  margin: 0;
+  padding-inline-start: 0;
+  display: flex;
+  list-style: none;
+  flex-wrap: nowrap;
+  flex-direction: row;
+  justify-content: space-between;
+  background-color: #fff;
+}
+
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,67 @@
...
+          <img
+            class="logo"
+            src="images/logo.png"
+            alt="MOYO"
+          >
+        </a>
+        <nav class="main_nav">
+          <ul class="nav__list">
+            <li class="list__item ">
+              <a class="link_list_item is-active" href="/">
```
tetiana-ivanchenko, line 25:
```suggestion
              <a class="link_list_item is-active" href="#apple">
```

tetiana-ivanchenko: fix all similar cases

tetiana-ivanchenko (changes requested):

src/style.css
```diff
@@ -1 +1,85 @@
...
+
+.is-active {
+  color: var(--focus_color);
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  height: 4px;
+  width: 37px;
```
IvanFesenko, line 73: If active will be another element it will looks like this one 
2023-04-18_13h01_16
But it should be 100% link width
2023-04-18_13h02_36

IvanFesenko (changes requested):
Check my comments

src/style.css
```diff
@@ -1 +1,82 @@
...
+* {
+  box-sizing: border-box;
+}
+
+:root {
+  --focus_color: #00acdc;
+}
+
+html {
+  font-family: Roboto , sans-serif;
```
qoa11a, line 10:
```suggestion
  font-family: Roboto, sans-serif;
```

src/style.css
```diff
@@ -1 +1,82 @@
...
+  font-weight: 500;
+}
+
+body {
+  margin: 0;
+}
+
+.main_header {
+  display: flex;
+  width: 100%;
```
qoa11a, line 21:
```suggestion
```
There's no need to do it as `display: flex;` makes the element 100% in width anyway.

src/style.css
```diff
@@ -1 +1,82 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav__list {
+  margin: 0;
+  padding-inline-start: 0;
+  display: flex;
+  list-style: none;
+  flex-wrap: nowrap;
+  flex-direction: row;
```
qoa11a, line 31: It's the default value. No need to specify it.

src/style.css
```diff
@@ -1 +1,82 @@
...
+  width: 100%;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav__list {
+  margin: 0;
+  padding-inline-start: 0;
+  display: flex;
+  list-style: none;
+  flex-wrap: nowrap;
```
qoa11a, line 30: Again, it's the default value. No need to specify it.

src/style.css
```diff
@@ -1 +1,82 @@
...
+  flex-direction: row;
+  justify-content: space-between;
+  background-color: #fff;
```
qoa11a, line 33: The background of the page is white anyway. No need to specify it.

src/style.css
```diff
@@ -1 +1,82 @@
...
+}
+
+.list__item:nth-child(n + 2) {
```
qoa11a, line 36:
```suggestion
.list__item:not(:first-child) {
```
It's actually the same but looks more understandable in my opinion. Not so important, just want to mention it.

src/style.css
```diff
@@ -1 +1,82 @@
...
+  margin-left: 20px;
+}
+
+.link_list_item {
+  white-space: nowrap;
+  padding: 22.5px 0;
+  display: flex;
+  background-color: #fff;
```
qoa11a, line 44: No need to specify it.

src/style.css
```diff
@@ -1 +1,82 @@
...
+  top: 56px;
+  height: 4px;
+  width: 100%;
+  background-color: var(--focus_color);
+  border-radius: 8px;
+}
+
+.list_container {
+  width: 100%;
+  display: inline-flex;
```
qoa11a, line 78: Why do you need inline here? It can be just flex.

src/style.css
```diff
@@ -1 +1,82 @@
...
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  top: 56px;
+  height: 4px;
+  width: 100%;
+  background-color: var(--focus_color);
+  border-radius: 8px;
+}
```
qoa11a, line 74: It's not good to specify constant values for such things. If you change the height of the navbar it's going to brake.

What you could do instead is:
1. Move the height of this element into the variable.
2. Calculate the top in this way: `calc(100% - var(--is-active_underline_height))`.

qoa11a (changes requested):
Almost done! You're doing a great job! 👍

src/style.css
ruslan-ihnatenko: `width:100%` removed

src/style.css
ruslan-ihnatenko: here i removed `flex-wrap` and `flex-direction` , but if i removed `display:flex` and `list-style` i will have this case
image

src/style.css
ruslan-ihnatenko: if will remove 
`white-space: nowrap;
  padding: 22.5px 0;
  display: flex;` 
  i will have this situation
image

i wrote `white-space: nowrap` because i dont want to have this case with "laptops & computers"
image

src/style.css
ruslan-ihnatenko: here i set `padding-inline-start: 0` because in default i have this value 40px
image

polosanya (approved):
Good job!

PR #3004, opened 2023-04-19
add task solution

maryanmats:
- [DEMO LINK](https://maryanmats.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://maryanmats.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,79 @@
 /* Styles go here */
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  background-color: #fff;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,79 @@
...
+  font-size: 12px;
+  line-height: 14px;
+  color: #000;
+  text-decoration: none;
+  text-transform: uppercase;
+  transition: color 0.5s;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
naziamoff, line 61: It is a great practice to create variables for the colors/fonts you use in your code. 
Imagine the page was redesigned - what will you do? You will need to fix all the colors/fonts manually. But if you use a variable - you can just change it in a single place

naziamoff (approved):
Great job! Here are suggestions for improvements:

as your `<li>` are on the same level of nesting and there are quite a lot of them - you can add some empty lines between them for better readability

PR #3035, opened 2023-04-20
add task solution

Huracann:
- [DEMO LINK](https://Huracann.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Huracann.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,106 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <!-- <h1>Moyo header</h1> -->
+
+    <header class="header">
+
+      <!-- logo -->
+      <a class="header_logo-link" href="#home">
+        <img
+          class="header__logo"
+          src="./images/logo.png"
+          alt="logo"
+        >
+      </a>
+
+      <!-- nav -->
+      <nav class="nav">
+        <ul class="nav__list">
+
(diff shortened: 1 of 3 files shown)


naziamoff (changes requested):
remove unnecessary comments, fix indentation and we're good

src/index.html
```diff
@@ -9,6 +9,88 @@
...
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+
+          <li class="nav__item">
+            <a
+              href="#apple"
+              class="nav__link is-active"
+            >
```
naziamoff, line 30: I'd say it is not necessary to move those to a new line. Wouldn't say it is a mistake though

src/index.html
```diff
@@ -9,6 +9,88 @@
...
-    <h1>Moyo header</h1>
+    <!-- <h1>Moyo header</h1> -->
+
+    <header class="header">
+
+      <a class="header_logo-link" href="#home">
+        <img
+          class="header__logo"
+          src="./images/logo.png"
+          alt="logo"
```
naziamoff, line 20: Alt attribute value should be maximally clear and maximally descriptive. That's an alternate description of the image, so I would come up with sth like that:
```suggestion
          alt="MOYO logo"
```

naziamoff (approved):
Good job! See some extra suggestions

PR #3039, opened 2023-04-20
add task solution

SergiyVS:
[DEMO LINK](https://SergiyVS.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://SergiyVS.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,84 @@
...
+* {
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  color: #000;
+
+  box-sizing: border-box;
```
polosanya, line 17:
```suggestion
* {
  box-sizing: border-box;
```
Do you really need to change the font styles and color of ALL elements?
Just set if for body or html and inherit where needed
Leave only box-sizing under * selector

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments

PR #3046, opened 2023-04-22
add task solution

sasdaiv:
- [DEMO LINK](https://sasdaiv.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://sasdaiv.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1 +1,59 @@
-/* Styles go here */
+@font-face {
+  font-family: "Roboto";
+  src: url(fonts/Roboto/Roboto-Medium.ttf);
+}
+
+html {
+  font-family: "Roboto", serif;
```
etojeDenys, line 7: Consider using a more modern font fallback like 'sans-serif' instead of 'serif' for the Roboto font.

etojeDenys (changes requested):

src/style.css
sasdaiv: "serif" has been changed to "sans-serif"
thx

IvanFesenko (approved):

PR #3058, opened 2023-04-26
my solution

rigency:


src/style.css
```diff
@@ -1 +1,61 @@
 /* Styles go here */
```
OleziO, line 1: Remove all comments.

src/style.css
```diff
@@ -1 +1,61 @@
...
+html {
+  --links-color: #00acdc;
```
OleziO, line 3: Move to :root selector.

src/style.css
```diff
@@ -1 +1,61 @@
...
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  display: block;
+  position: relative;
+  padding: 23px 0;
+  font-family: Roboto, sans-serif;
```
OleziO, line 39: Move to html selector and remove from all other places.

OleziO (changes requested):

src/style.css
rigency: > Move to :root selector.
I didn't understand. HTML is just the root for all tags

src/style.css
```diff
@@ -1 +1,60 @@
...
+}
+
+.header__logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
+  --links-color: #00acdc;
```
sTorba24, line 24: mentor meant to move it to :root selector
https://developer.mozilla.org/en-US/docs/Web/CSS/:root
it's good practice to declare variables there

sTorba24 (approved):
Good work! I left you comment with explanation about variables

PR #3062, opened 2023-04-28
Develop

kbekher:
- [DEMO LINK](https://kbekher.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://kbekher.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,57 @@
-/* Styles go here */
+body,
+ul {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,57 @@
...
+}
+
+.nav-link:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  background-color: #00acdc;
```
IvanFesenko, line 50: Use varaibles for repeatable colors

IvanFesenko (approved):
Check my comment to avoid same mistakes in the future

src/style.css
kbekher: done, thank you! 🙂

PR #3089, opened 2023-05-05
add task solution

MakaronX:


src/style.css
```diff
@@ -1 +1,77 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.header {
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  max-width: 100%;
```
YegorKorenkov, line 21: redundant

src/index.html
```diff
@@ -3,12 +3,57 @@
...
+        <img
+          src="./images/logo.png"
+          alt="logo_MOYO"
+          class="header__images"
+        >
+      </a>
+      <nav class="header__nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#Apple" class="nav__link is-active">Apple</a>
```
YegorKorenkov, line 26: image

src/style.css
```diff
@@ -1 +1,77 @@
...
+  align-items: center;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+  margin: 0;
```
YegorKorenkov, line 64: Please remove margin-left for first nav item, but not for is-active. If, for example, the third element becomes active, then its margin-right will also be deleted

src/style.css
```diff
@@ -1 +1,77 @@
...
+  position: relative;
+}
+
+.is-active:after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  width: 37px;
```
YegorKorenkov, line 73: make this line on fullwidth of parent

src/style.css
```diff
@@ -1 +1,77 @@
...
+  text-align: center;
+  align-items: center;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
```
YegorKorenkov, line 63: you can also move this blue color to variable

YegorKorenkov (changes requested):

MakaronX:
Is everything okay now?

src/style.css
MakaronX: I don't know how to do it

PR #3105, opened 2023-05-09
Add task solution

spojrzenie:


DorotaLeniecDev (changes requested):
Few changes needed

src/style.css
DorotaLeniecDev, line 1: add box shaddow for the headers, as it is specified in the designs:
box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.05);

DorotaLeniecDev (changes requested):
changes needed

src/style.css
```diff
@@ -1 +1,76 @@
...
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  align-items: center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo img {
```
DorotaLeniecDev, line 17: Style the image using class :) element selectors increases specificity

DorotaLeniecDev (changes requested):
Good job, one more comment, style all elements using classes

spojrzenie:
Thank, please check now.

src/style.css
```diff
@@ -1 +1,76 @@
...
+}
+
+.menu-item--link {
+  color: black;
+  text-decoration: none;
+  display: block;
+  padding: 23px 0;
+}
+
+.menu-item a:hover {
```
darokrk, line 49: look Dorota's suggestion before, you added .`menu-item--link` so you should style this element by this class name not by tag `a` name, just use your class for styling :)

src/style.css
```diff
@@ -1 +1,76 @@
...
+  padding: 0;
+  display: flex;
+}
+
+.menu-item {
+  float: left;
+  margin-right: 20px;
+}
+
+.menu-item--link {
```
darokrk, line 42: this class name should be `.menu-item__link`, -- is used as a modifier class but this is a separate element so for BEM convention please use __

darokrk (approved):
Nicely done, looks good 🥇  ! Approved, but please check my last suggestion and I would suggest improving it :)

src/style.css
spojrzenie: Ok, thank you.

PR #3151, opened 2023-05-19
Develop

MaksymKolesnyk:
https://MaksymKolesnyk.github.io/layout_moyo-header/
https://MaksymKolesnyk.github.io/layout_moyo-header/report/html_report/


src/style.css
```diff
@@ -1 +1,77 @@
...
+  margin: 0 10px;
+}
+
+.menu-link {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 23px 0 19px;
+  text-decoration: none;
+  color: rgba(0, 0, 0, 1);
```
OleziO, line 59: Try to use HEX instead of RGBA

src/style.css
```diff
@@ -1 +1,77 @@
...
+}
+
+.down-line {
+  height: 4px;
+  width: 100%;
+  background-color: rgba(0, 172, 220, 1);
```
OleziO, line 65: Move all repeatable colors to CSS variables.

[More info](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/style.css
```diff
@@ -1 +1,77 @@
...
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  font-size: 12px;
+  line-height: 14px;
+  font-family: Roboto, sans-serif;
```
OleziO, line 13: Move it to html selector and remove it from all other places.

OleziO (changes requested):
image

Make this logic with `before/after

[More info.](https://developer.mozilla.org/en-US/docs/Web/CSS/::before)

OleziO (approved):

PR #3154, opened 2023-05-19
Moyo header solution

oleksandrusov:
Demo link: https://oleksandrusov.github.io/layout_moyo-header/

Test report link: https://oleksandrusov.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,8 +17,7 @@
     <link rel="stylesheet" href="./style.css">
   </head>
 
-  <body class="body">
-
+  <body>
     <header class="header">
       <a href="#home" class="logo">
         <img
@@ -27,35 +26,43 @@
           alt="Moyo logo"
         >
       </a>
+
       <nav class="nav">
         <ul class="nav__list">
           <li class="nav__item">
             <a class="nav__link is-active" href="index.html">Apple</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="index.html">Samsung</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="index.html">Smartphones</a>
           </li>
+
           <li class="nav__item">
             <a
               class="nav__link"
               data-qa="hover"
               href="index.html"
             >
-            Laptops & Computers
+              Laptops & Computers
             </a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="index.html">Gadgets</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="index.html">Tablets</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="index.html">Photo</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="index.html">Video</a>
           </li>
--- a/src/style.css
+++ b/src/style.css
@@ -4,12 +4,11 @@ html {
   --background-color: #e5e5e5;
   --header-color: #fff;
   --header-height: 60px;
-  --logo-height: 40px;
-  --logo-width: 40px;
+  --logo-size: 40px;
   --link-color: #00acdc;
 }
 
-.body {
+body {
   margin: 0;
   background-color: var(--background-color);
 }
@@ -19,6 +18,8 @@ html {
   position: relative;
   align-items: center;
   justify-content: space-between;
+  padding: 0 50px;
+  box-sizing: border-box;
   background-color: var(--header-color);
   box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
 }
@@ -26,17 +27,12 @@ html {
 .nav {
   display: flex;
   align-items: center;
-  margin-right: 50px;
-}
-
-.logo {
-  padding: 10px 0 10px 50px;
 }
 
 .logo__img {
   display: block;
-  height: var(--logo-height);
-  width: var(--logo-width);
+  height: var(--logo-size);
+  width: var(--logo-size);
 }
 
 .nav__list {
@@ -49,15 +45,17 @@ html {
 
 .nav__link {
   display: inline-block;
-  margin-left: 20px;
   line-height: var(--header-height);
   color: #000;
   text-decoration: none;
 }
 
+.nav__item:not(:first-child) {
+  margin: 0 0 0 20px;
+}
+
 .is-active {
   color: var(--link-color);
-  margin-left: 0;
   position: relative;
 }
 


src/style.css
```diff
@@ -1 +1,95 @@
...
+  --logo-width: 40px;
+  --link-color: #00acdc;
+}
+
+.body {
+  margin: 0;
+  background-color: var(--background-color);
+}
+
+.header {
```
YegorKorenkov, line 17: Please add padding for logo, instead of adding margin right/left for logo and nav
image

src/style.css
```diff
@@ -1 +1,95 @@
...
+  display: flex;
+  position: relative;
+  align-items: center;
+  justify-content: space-between;
+  height: var(--header-height);
```
YegorKorenkov, line 22: try to not hardcode height

src/style.css
```diff
@@ -1 +1,95 @@
...
+  align-items: center;
+  margin-right: 50px;
+}
+
+.logo__img {
+  display: block;
+  height: var(--logo-height);
+  width: var(--logo-width);
+  margin-left: 50px;
```
YegorKorenkov, line 38: Pay attention on first comment

src/style.css
```diff
@@ -1 +1,95 @@
...
+  height: var(--header-height);
+  background-color: var(--header-color);
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav {
+  display: flex;
+  height: var(--header-height);
+  align-items: center;
+  margin-right: 50px;
```
YegorKorenkov, line 31: Pay attention on first comment

src/style.css
```diff
@@ -1 +1,95 @@
...
+  display: block;
+  position: absolute;
+  bottom: 0;
+  height: 4px;
+  width: 100%;
+  background: var(--link-color);
+  border-radius: 8px;
+}
+
+[data-qa="hover"] {
```
YegorKorenkov, line 81: Why you nedd this? Redundant

src/style.css
```diff
@@ -1 +1,95 @@
...
+  color: var(--link-color);
+  margin-left: 0;
+}
+
+.nav__link:hover {
+  color: var(--link-color);
+}
+
+.nav__link.is-active {
+  position: relative;
```
YegorKorenkov, line 67: move to  .is-active selector on 57 row

YegorKorenkov (changes requested):

oleksandrusov:
Hello, Yegor! I made changes:
- added padding to logo instead of margin but with .nav I can pass the test only with margin right
- removed height from .header but still have var for it, is it right?
- removed [data-qa=‘hover’]. Cant pass the test and I thought it was because of it.
- moved position: relative, like you said
So now have question about margin-right for .nav and header height.
Thank you!

src/index.html
```diff
@@ -2,13 +2,65 @@
...
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    >
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Moyo header</h1>
+
+  <body class="body">
```
Viktor-Kost, line 20: no needs to add class to a body, since you can use tag selector in this case

src/index.html
```diff
@@ -2,13 +2,65 @@
...
+
```
Viktor-Kost, line 21: an extra empty line between parent tag and child one 
```suggestion
```

src/style.css
```diff
@@ -1 +1,89 @@
...
+  --logo-width: 40px;
+  --link-color: #00acdc;
+}
+
+.body {
+  margin: 0;
+  background-color: var(--background-color);
+}
+
+.header {
```
Viktor-Kost, line 17: sorry for a little discoordination with the previous mentor, but the correct solution is to add padding: 0 50px; to a header instead of adding paddings to your logo and nav. Since header is a wrapper and if you'd like to fit content inside you should add padding to this wrapper (header), not its children

src/style.css
```diff
@@ -1 +1,89 @@
...
+  align-items: center;
+  justify-content: space-between;
+  background-color: var(--header-color);
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.nav {
+  display: flex;
+  align-items: center;
+  margin-right: 50px;
```
Viktor-Kost, line 29: redundant since we use padding for a wrapper
```suggestion
```

src/style.css
```diff
@@ -1 +1,89 @@
...
+}
+
+.logo {
+  padding: 10px 0 10px 50px;
+}
```
Viktor-Kost, line 34: redundant since we use padding for a wrapper
```suggestion
```

src/style.css
```diff
@@ -1 +1,89 @@
...
+  display: flex;
+  margin: 0;
+  padding: 0;
+  list-style: none;
+  text-transform: uppercase;
+}
+
+.nav__link {
+  display: inline-block;
+  margin-left: 20px;
```
Viktor-Kost, line 52: You should avoid margin collapsing, when a child margins go through a parent and push off another parent
[useful_link](https://www.smashingmagazine.com/2019/07/margins-in-css/#parent-and-first-or-last-child-element)
So, if you'd like one element to be push off from another element use margins at THESE elements, not their children

P.S. and the first element shouldn't have margin-left. Use `:not(:first-child)`

src/style.css
```diff
@@ -1 +1,89 @@
...
+  line-height: var(--header-height);
+  color: #000;
+  text-decoration: none;
+}
+
+.is-active {
+  color: var(--link-color);
+  margin-left: 0;
```
Viktor-Kost, line 60: Use :not(:first-child) it's not a good practice to reset margin as you did, since what will you get, if your `is-active` element will be the second in a row?

src/style.css
```diff
@@ -1 +1,89 @@
...
+}
+
+.logo {
+  padding: 10px 0 10px 50px;
+}
+
+.logo__img {
+  display: block;
+  height: var(--logo-height);
+  width: var(--logo-width);
```
Viktor-Kost, line 39: as the logo width and height have the same params you can use just one value instead (logo-size)

src/index.html
```diff
@@ -2,13 +2,65 @@
...
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+          src="./images/logo.png"
+          class="logo__img"
+          alt="Moyo logo"
+        >
+      </a>
+      <nav class="nav">
```
Viktor-Kost, line 30: add an empty line between _multiline_ siblings. Check similar cases 
```suggestion
      </a>
      
      <nav class="nav">
```

Viktor-Kost (changes requested):

src/style.css
oleksandrusov: But if I add padding: 0 50px; header size will be 50 + 1200 + 50 instead of 1200. In this case logo and nav will be closer to each one. If I understand right.

oleksandrusov: Screenshot 2023-05-25 at 10 04 30

oleksandrusov: But it works if I use header with box-sizing: border-box;

Screenshot 2023-05-25 at 10 23 19
.header {
  display: flex;
  position: relative;
  align-items: center;
  padding: 0 50px;
  box-sizing: border-box;
  background-color: var(--header-color);
  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
}

oleksandrusov:
Made all requested changes.

nazarmatsevych (approved):

PR #3161, opened 2023-05-20
Add task solution

adv555:
- [DEMO LINK](https://adv555.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://adv555.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,13 +2,87 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <meta
+      http-equiv="X-UA-Compatible"
+      content="ie=edge"
+    >
     <title>Moyo header</title>
-    <link rel="stylesheet" href="./style.css">
+    <link
+      rel="stylesheet"
+      href="./style.css"
+    >
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
+        <a
+          href="#home"
+          class="logo"
+        >
+          <img
+            src="/images/logo.png"
+            alt="logo"
+            class="logo__img"
+          >
+        </a>
+      </div>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="#apple"
+              class="nav__link is-active"
+            >Apple</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#samsung"
+              class="nav__link"
+            >Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#smartphones"
+              class="nav__link"
+            >Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#laptops & computers"
+              class="nav__link"
+              data-qa="hover"
+            >Laptops & Computers</a>
+          </li>
+          <li class="nav__item">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -2,13 +2,87 @@
...
+            <a
+              href="#apple"
+              class="nav__link is-active"
+            >Apple</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#samsung"
+              class="nav__link"
+            >Samsung</a>
```
ManoylovAC, line 45: In case of multiline tag It will be more clear if you place content on separate line also
```suggestion
            >
              Samsung
            </a>
```

ManoylovAC (changes requested):
Hi 👋
Good time to some improvements.
Please check requirements, for example your layout haven't fit on different widths (1024px, 1200px). 
Hint: for adaptive styles you can youse media-queries

image

src/style.css
```diff
@@ -1 +1,96 @@
...
+  max-width: 1200px;
+
+  @media (min-width: 1024px) {
+    max-width: 1024px;
+  }
+
+  @media (min-width: 1200px) {
+    max-width: 1200px;
+  }
+}
```
ManoylovAC, line 34: Yes, it will work in css-preprocessors like `SASS`. But you are in `css` now, so there is no hierarchy included.
In this case, your media rules should be separate. Please update this part with appropriate modifications.

ManoylovAC (changes requested):
You are almost done)
One thing it's you add ccs media inside other rule. But this possible only in css-preprocessors. See my comments and fix it please. As I see everything else is ok.

src/style.css
adv555: Ths :blush: Really mixed up. Fixed.  All test shows[ Ok](https://adv555.github.io/layout_moyo-header/report/html_report/), but build still failed 😞 
Would be very grateful for the advice. [Screenshort](https://prnt.sc/tUGUW9_J8GPE)

src/style.css
```diff
@@ -1 +1,99 @@
...
+  --background-color: #fff;
+  --header-line-height: 60px;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Roboto, sans-serif;
```
tetiana-ivanchenko, line 15: image

src/index.html
```diff
@@ -2,13 +2,108 @@
...
+      <div class="container">
+        <div class="header__content">
+          <div class="logo">
+            <a
+              href="#home"
+              class="logo"
+            >
+              <img
+                src="/images/logo.png"
+                alt="logo"
```
tetiana-ivanchenko, line 31: image

src/index.html
```diff
@@ -2,13 +2,108 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
+        <div class="header__content">
+          <div class="logo">
+            <a
+              href="#home"
+              class="logo"
+            >
```
tetiana-ivanchenko, line 28: optional:
If the HTML-element has long attribute values or number of attributes is more than 2 - start each one, including the first, on the new line with 2-space indentation related to tag. (so, 2 or less, if their length is less than the maximum allowable, you do not need to move to a new line)

tetiana-ivanchenko (changes requested):

src/style.css
adv555: Hi! 👋  As a "fallback" system, it shares a common generic-family as "sans-serif", but added an additional family-name of 'Helvetica' . Done :white_check_mark:

src/index.html
adv555: Fixed :white_check_mark:

src/index.html
adv555: If you don’t mind, I’ll leave it as it is, because I use automatic formatting from Vscode, which does not allow to limit the number of attributes on a new line, but speeds up the writing of code. Maybe there are some additional settings or plugins that I don’t know about. I would appreciate any advice on how to better tune it.😊

YegorKorenkov (approved):
Nice work!

PR #3172, opened 2023-05-23
Develop

IgorOksentyuk:
- [DEMO LINK](https://igoroksentyuk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://igoroksentyuk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,10 +5,45 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#">
+        <img class="header__logo" src="./images/logo.png" alt="moyo-logo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="nav__link is-active" href="#">apple</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#">samsung</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#">smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#" data-qa="hover">laptops & computers</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#">gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#">tablets</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#">photo</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#">video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -5,10 +5,53 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#">
+        <img class="header__logo"
+          src="./images/logo.png"
+          alt="moyo-logo"
+        >
```
ManoylovAC, line 20: if you are placing attributes on separate lines, it would be better to do this for all attributes. It is more clear for readability
```suggestion
        <img
          class="header__logo"
          src="./images/logo.png"
          alt="moyo-logo"
        >
```

src/index.html
```diff
@@ -5,10 +5,53 @@
...
+          <li class="nav__item">
+            <a class="nav__link" href="#">smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+              class="nav__link"
+              href="#"
+              data-qa="hover">
+              laptops & computers
+            </a>
```
ManoylovAC, line 39: visually, the content with "laptops and computers" looks like part of the attributes, so it is better to place the closing corner on a separate line
```suggestion
            <a
              class="nav__link"
              href="#"
              data-qa="hover"
            >
              laptops & computers
            </a>
```

ManoylovAC (changes requested):
Good start 👍 
Please recheck requirements, for example your layout haven't fit on different widths (1024px, 1200px). And check my comments I left.
Hint: for adaptive styles you can youse `media-queries`
Good luck!

src/index.html
IgorOksentyuk: Got it! Ty!

src/index.html
IgorOksentyuk: +

src/style.css
```diff
@@ -1 +1,78 @@
...
+  bottom: 0;
+  left: 0;
+  content: "";
+  height: 4px;
+  width: 100%;
+  border-radius: 8px;
+  background-color: var(--blueHover);
+}
+
+@media screen and (min-width: 1200px) {
```
ManoylovAC, line 73: I guess you also should have another one rule for 1024 with
image

src/style.css
```diff
@@ -1 +1,78 @@
...
+  .header {
+    max-width: 1200px;
+    margin: auto;
+  }
```
ManoylovAC, line 77: I would recommend do it for some new inner element, because now your header looks with some shades but it should not
image

src/style.css
IgorOksentyuk: Done!

src/style.css
IgorOksentyuk: Done! I created inner div with class .header__container, and added media query for it!

src/style.css
```diff
@@ -1 +1,88 @@
...
+.nav__item {
+  margin-left: 20px;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
+
+.nav__link {
+  font-family: "Roboto", sans-serif;
```
tetiana-ivanchenko, line 43: image

tetiana-ivanchenko (changes requested):

src/style.css
IgorOksentyuk: Hello, did you mean that in my case it is not necessary to use quotation marks? Or I need to add one more alternative fonts, except sans-serif?

tetiana-ivanchenko: Hello, if you have any questions about how to resolve the comments, feel free to ask it in the fe_chat, instead of re-requesting review
no, i mean you should use fallback fonts - alternative font-family in case the main one doesn't work [like this](https://www.w3schools.com/cssref/pr_font_font-family.asp)

tetiana-ivanchenko (changes requested):

src/style.css
IgorOksentyuk: I found that the best fallback fonts for "Roboto" are "Helvetica Neue" and "sans-serif". Is it a good decision?)

YegorKorenkov (approved):
Good job! About fallback fonts, you did everything right!

PR #3181, opened 2023-05-24
add task solution

RDNCry:
- [DEMO LINK](https://rdncry.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://rdncry.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,34 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <a href="#home" class="logo"
+      alt="logo"><img src="./images/logo.png" alt="moyo_logo"></a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__link"><a href="#" class="is-active">Apple</a></li>
+
+          <li class="nav__link"><a href="#">Samsung</a></li>
+
+          <li class="nav__link"><a href="#">Smartphones</a></li>
+
+          <li class="nav__link"><a href="#" data-qa="hover">Laptops & Computers</a></li>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -6,9 +6,34 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <a href="#home" class="logo"
+      alt="logo"><img src="./images/logo.png" alt="moyo_logo"></a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__link"><a href="#" class="is-active">Apple</a></li>
```
ManoylovAC, line 20: It will be clearer if you separate the tags on different lines for all your cases.
Currently you don't have actual link addresses and only the `#' character, but imagine when that happens, the length will increase and it will make the markup even more difficult to reading.
```suggestion
          <li class="nav__link">
            <a href="#" class="is-active">Apple</a>
          </li>
```

src/index.html
```diff
@@ -6,9 +6,34 @@
...
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <a href="#home" class="logo"
+      alt="logo"><img src="./images/logo.png" alt="moyo_logo"></a>
```
ManoylovAC, line 17: Please check [codestyle checklist](https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md) especially about indentations (point 13)

src/style.css
```diff
@@ -1 +1,54 @@
...
+  margin: 10px 0;
+}
+
+.nav__list {
+  display: flex;
+  padding: 0;
+  list-style: none;
+}
+
+.nav__link a {
```
ManoylovAC, line 26: please try to avoid using simple tags in your rules
it is better add to element you want special class
because now it will affects all links that you will added in the next work

src/style.css
```diff
@@ -1 +1,54 @@
...
+  line-height: 60px;
+  text-decoration: none;
+  color: #000;
+}
+
+.nav__link:not(:last-child) {
+  margin-right: 20px;
+}
+
+a.is-active,
```
ManoylovAC, line 39: so yes, please add some new classname to your links and use them instead of tag

ManoylovAC (changes requested):
Hi 👋 
It's a good time for some improvements.
Please recheck requirements, for example your layout haven't styles to fit your header on different screen sizes (width: 1024px, 1200px).
Hint: for adaptive styles you can youse media-queries

Also check my suggestions in comment I left.
Good luck 🍀

PR #3187, opened 2023-05-25
Add task solution

parmenid89:
[DEMO LINK](https://parmenid89.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://parmenid89.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -3,12 +3,55 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="index.html" class="header__logo">
+        <img src="images/logo.png" alt="Moyo logo">
+      </a>
+
+      <nav clas="'nav">
```
And678, line 20: Two typos here
```suggestion
      <nav class="nav">
```

And678 (changes requested):
Hello, looks good, one issue though, layout breaks on screen width's >1024, <1200, check the image below. To fix, try using [max-width](https://developer.mozilla.org/en-US/docs/Web/CSS/max-width) on `header` instead of `width`.
image

src/index.html
parmenid89: Changed, tnx

src/style.css
```diff
@@ -1 +1,77 @@
...
+  --color-main: #00acdc;
+}
+
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+}
```
tetiana-ivanchenko, line 11: image
image

src/style.css
```diff
@@ -1 +1,77 @@
...
+
+.nav__item {
+  margin-left: 20px;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
+
+.nav__item :hover {
```
tetiana-ivanchenko, line 44:
```suggestion
.nav__item:hover {
```

tetiana-ivanchenko (changes requested):

src/style.css
parmenid89: Is thit correct that i used font-family: "Roboto", sans-serif; (sans-serif like alternative font-family) or i should add one more font to the project?

src/style.css
parmenid89: Fixed, tnx

anstsot (approved):

PR #3203, opened 2023-05-27
add task solution

MariiaMonakhova:
- [DEMO LINK](https://MariiaMonakhova.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://MariiaMonakhova.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!Doctype html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,72 @@
...
+          class="logo"
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item active__link">
+            <a
+            href="#apple"
+            class="nav__link is-active"
+            >Apple</a>
```
anstsot, line 29: fix the code style according to the checklist (points 5 and 13)

anstsot: fix all across the HTML document

src/style.css
```diff
@@ -1 +1,75 @@
-/* Styles go here */
+* {
+  --active-color: #00acdc;
```
anstsot, line 2: it's better to add the CSS variable to `:root` selector

src/style.css
```diff
@@ -1 +1,75 @@
...
+  margin: 0;
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
```
anstsot, line 36: add font-family style to body or html selector. then you can use font-family: inherit
do not forget to add more fallback fonts
image

src/style.css
```diff
@@ -1 +1,75 @@
...
+.is-active {
+  color: var(--active-color);
+}
+
+.is-active::after {
+  position: absolute;
+  content: "";
+  left: 0;
+  bottom: 0;
+  width: 37px;
```
anstsot, line 65: do not hardcode the `width` because it won't work for other active links with different width

src/style.css
```diff
@@ -1 +1,75 @@
...
+}
+
+.nav__item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav__link {
+  text-decoration: none;
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
```
anstsot, line 37: it's default style so there is no need to add it

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.

src/style.css
MariiaMonakhova: done!

src/style.css
```diff
@@ -1 +1,75 @@
-/* Styles go here */
+:root {
+  --active-color: #00acdc;
+
+}
+
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif, serif;
```
tetiana-ivanchenko, line 8: Remember to use fallback fonts - alternative font-family in case the main one doesn't work [like this](https://www.w3schools.com/cssref/pr_font_font-family.asp)

tetiana-ivanchenko (changes requested):

src/style.css
```diff
@@ -1 +1,75 @@
...
+
+body {
+  margin: 0;
+  font-family: "Roboto", "Helvetica Neue", Helvetica, sans-serif;
+  box-sizing: border-box;
+}
+
+.header {
+  display: flex;
+  width: 1100px;
```
YegorKorenkov, line 14: 1. Please use max-width so that the header does not go beyond the screen borders
image
2. Why value is 1100px? I understand that you did this in order to have 50 pixel intend on the sides, but it’s better to make the width as in the task requirments and explicitly set the side paddings
image

src/style.css
```diff
@@ -1 +1,75 @@
...
+  display: inline-block;
+  height: 60px;
+  line-height: 60px;
+}
+
+.nav__link:hover {
+  color: var(--active-color);
+}
+
+.logo {
```
YegorKorenkov, line 52: Please move this selector to the top so that the css style order is the same as the html code order. It structures your code better, and is more readable

YegorKorenkov (approved):
Nice work! 🔥Please fix minor mistakes

src/style.css
MariiaMonakhova: Done! Thank you for your feedback

meforoma (approved):

meforoma (approved):

PR #3214, opened 2023-05-29
completed task

ovynarchuk:
- [DEMO LINK](https://ovynarchuk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ovynarchuk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,10 +5,61 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="menu">
+      <div class="menu__container">
+        <a href="#home" class="menu__logo">
+          <img src="images/logo.png" alt="logo">
+        </a>
+
+        <nav class="nav">
+
+          <ul class="nav__list">
+
+            <li class="nav__item">
+              <a href="#apple" class="is-active nav__link">apple</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#samsung" class="nav__link">samsung</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#smartphones" class="nav__link">smartphones</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#laptops&computers"
+              class="nav__link"
+              data-qa="hover"
+              >laptops & computers</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#gadgets" class="nav__link">gadgets</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#tablets" class="nav__link">tablets</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#photo" class="nav__link">photo</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#video" class="nav__link">video</a>
+            </li>
+
+          </ul>
+        </nav>
+      </div>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,82 @@
-/* Styles go here */
+body {
+  margin: 0;
+  --color: #00acdc;
+}
+
+.menu {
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -5,10 +5,61 @@
...
+            <li class="nav__item">
+              <a href="#samsung" class="nav__link">samsung</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#smartphones" class="nav__link">smartphones</a>
+            </li>
+
+            <li class="nav__item">
+              <a href="#laptops&computers"
```
YegorKorenkov, line 38: Please re-read this paragraph in checklist
image

src/index.html
```diff
@@ -5,10 +5,61 @@
...
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="menu">
```
YegorKorenkov, line 15: It would be more correct to change the class name to header, because this element has both the site logo and navigation and it is not quite correct to name it as you named it

src/style.css
```diff
@@ -1 +1,82 @@
...
+
+.menu {
+  max-width: 1200px;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.menu__logo {
+  display: inline-block;
+  padding: 10px 0;
```
YegorKorenkov, line 14: It is not quite right to center the element through the padding, if the height of the header changes, then you will need to change the padding. You have already set a property on the parent flex element that centers all children along the Y axis (align-items:center)

src/style.css
```diff
@@ -1 +1,82 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.nav__item {
+  margin-right: 20px;
+}
+
+.nav__item:last-child {
+  margin-right: 0;
```
YegorKorenkov, line 40: You can use .nav__item:not(:last-child) selector to shorten your code a bit

src/style.css
```diff
@@ -1 +1,82 @@
...
+}
+
+.nav__link {
+  font-family: "Roboto", sans-serif;
```
YegorKorenkov, line 44: Please move this to html selector, it's more correctly

src/style.css
```diff
@@ -1 +1,82 @@
...
+  display: flex;
+  align-items: center;
+  height: 60px;
+  font-weight: 500px;
```
YegorKorenkov, line 48: Wrong value for font-weight,please check

src/style.css
```diff
@@ -1 +1,82 @@
...
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  text-decoration: none;
+  position: relative;
+  color: black;
```
YegorKorenkov, line 54: 1. More correctly provide color value in #hex or rgba format
2. If you already use variable for blue color, please also move this black color to variable, because if you need to change this 'default' color, it will be enough for you to change the value of the variable and not look for it through the whole code

YegorKorenkov (changes requested):
Please check remarks, and also it would be cool if you move the header in the center of the page
image

src/index.html
ovynarchuk: I've made changes to correct ones. Please check them.
And about the font-weight - pretty sure it was 500, but I cannot open figma to check what was it

src/index.html
```diff
@@ -5,10 +5,63 @@
...
+            <li class="nav__item">
+              <a href="#smartphones" class="nav__link">smartphones</a>
+            </li>
+
+            <li class="nav__item">
+              <a
+                href="#laptops&computers"
+                class="nav__link"
+                data-qa="hover"
+              >laptops & computers
```
YegorKorenkov, line 42: Text shoud be on new line
```
>
  laptops & computers
</a>
```

src/style.css
```diff
@@ -1 +1,82 @@
...
+  position: relative;
+  color: var(--default_color);
+  font-style: normal;
+}
+
+.nav__link:hover {
+  color: var(--color);
+}
+
+.is-active {
```
YegorKorenkov, line 62: you can shorten your code a bit by moving your selector to nav__link:hover separated by commas (.nav__link:hover, .is-active), because both of selector have the same rule

YegorKorenkov (approved):
Nice work, please pay attention on comments

src/style.css
ovynarchuk: got it, will do

src/index.html
ovynarchuk: will make changes

PR #3219, opened 2023-05-29
add task solution

melnyksergey7:
[DEMO LINK]
[TEST REPORT LINK]

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,59 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link
+      href="https://fonts.googleapis.com/css?family=Roboto:500&display=swap"
+      rel="stylesheet"
+    >
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
+        <a href="#home" class="header__link">
+          <img
+          class="header__logo"
+          src="./images/logo.png"
+          alt="moyo"
+          >
+        </a>
+        <nav>
+          <ul class="nav__list">
+            <li class="nav__item">
+              <a
+              href="#apple"
+              class="nav__link is-active">
+              Apple
+              </a>
+            </li>
+            <li class="nav__item">
+              <a href="#samsung" class="nav__link">Samsung</a>
+            </li>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,81 @@
...
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  font-family: Roboto, Arial, Helvetica, sans-serif;
+  font-weight: 500;
+  font-style: normal;
```
anstsot, line 10: it's default style so there is no need to add it
```suggestion
```

src/style.css
```diff
@@ -1 +1,81 @@
...
+  .container {
+    max-width: 1024px;
+  }
+}
+
+a,
+a:visited,
+a:hover,
+a:focus,
+a:active {
```
anstsot, line 35: better to use class name instead of tag

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.

src/style.css
melnyksergey7: I couldn`t understand which class name i can use instead of tag?

src/index.html
```diff
@@ -7,8 +7,59 @@
...
+            src="./images/logo.png"
+            alt="moyo"
+          >
+        </a>
+        <nav>
+          <ul class="nav__list">
+            <li class="nav__item">
+              <a
+              href="#apple"
+              class="nav__link is-active">
```
YegorKorenkov, line 30: Please re-read point 13 in checklist

src/index.html
```diff
@@ -7,8 +7,59 @@
...
+              <a href="#samsung" class="nav__link">Samsung</a>
+            </li>
+            <li class="nav__item">
+              <a href="#smartphones" class="nav__link">Smartphones</a>
+            </li>
+            <li class="nav__item">
+              <a
+              href="#laptops&computers"
+              class="nav__link"
+              data-qa="hover">
```
YegorKorenkov, line 44: The same, re-read point 13 in checklist

YegorKorenkov (changes requested):
Hi, please carefully check your code according to the checklist, making sure that everything is correct, only after that send the task to review

anstsot (approved):

PR #3230, opened 2023-05-29
add task solution

AndriiMelnyk94:


src/index.html
```diff
@@ -3,12 +3,57 @@
...
+            </li>
+            <li class="nav__item">
+              <a class="nav__link" href="#smartphones">smartphones</a>
+            </li>
+            <li class="nav__item">
+              <a
+                class="nav__link"
+                href="#computers"
+                data-qa="hover">
+                laptops & computers
```
YegorKorenkov, line 38: Please don't forget about proper formatting
image

src/style.css
```diff
@@ -1 +1,77 @@
...
+.nav__item:last-child {
+  margin-right: 0;
+}
+
+.nav__link {
+  display: flex;
+  align-items: center;
+  height: 60px;
+  position: relative;
+  font-family: "Roboto", sans-serif;
```
YegorKorenkov, line 51: It's better to initialize the font-family in the selector because usually the site uses one font and it should be available for all elements. Also better to add more fallback fonts accordingly to paragraph 3 in checklist

YegorKorenkov (approved):
Approve, but, please, pay attention on comments

PR #3253, opened 2023-05-30
add task solution

melarkkkk:


src/style.css
```diff
@@ -1 +1,78 @@
 /* Styles go here */
+
+* {
```
YegorKorenkov, line 3: usually this selector is used only to add a `box-sizing: border-box;` rule
image

src/style.css
```diff
@@ -1 +1,73 @@
...
+  font-size: 12px;
+  line-height: 60px;
+  height: 60px;
+}
+
+.nav__link:hover {
+  color: var(--active-color);
+}
+
+.is-active:link {
```
anstsot, line 60: you can merge these styles
```suggestion
.nav__link:hover,
.is-active:link {
```

anstsot (approved):

PR #3264, opened 2023-05-30
layout_moyo-header

katya-sn:


src/style.css
```diff
@@ -1 +1,83 @@
...
+  color: #00acdc;
+}
+
+.nav__item:hover {
+  text-decoration: none;
+}
+
+.is-active {
+  position: relative;
+  color: #00acdc;
```
OleziO, line 69: Move all repeatable colors to CSS variables.

OleziO (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #3270, opened 2023-05-30
add task solution

NikitaOrliak:
- [DEMO LINK](https://NikitaOrliak.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://NikitaOrliak.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,56 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+    rel="preconnect"
+    href="https://fonts.gstatic.com"
+    crossorigin="anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,500;0,700;1,400;1,700&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="home-page">
+        <img
+        src="./images/logo.png"
+        alt="moyo-logo"
+        class="logo"
+        >
+      </a>
+      <nav class="navigation">
+        <ul class="nav__list">
+          <li class="nav__element_of_list">
+            <a href="apple" class="nav__link is-active">Apple</a>
+          </li>
+          <li class="nav__element_of_list">
+            <a href="samsung" class="nav__link">Samsung</a>
+          </li>
+          <li class="nav__element_of_list">
+            <a href="smartphones" class="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__element_of_list">
+            <a
+            href="laptops&computers"
+            class="nav__link"
+            data-qa="hover"
+            >Laptops & Computers</a>
+          </li>
+          <li class="nav__element_of_list">
+            <a href="gadgets" class="nav__link">Gadgets</a>
+          </li>
+          <li class="nav__element_of_list">
+            <a href="tablets" class="nav__link">Tablets</a>
+          </li>
+          <li class="nav__element_of_list">
+            <a href="photo" class="nav__link">Photo</a>
+          </li>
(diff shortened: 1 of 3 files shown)


OleziO (changes requested):

src/style.css
```diff
@@ -1 +1,79 @@
...
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  max-height: var(--max__height);
+
+  box-sizing: border-box;
+
+  /* border: 2px solid black; */
```
slavanosatskyi, line 21: remove redundant comments and empty lines

src/style.css
```diff
@@ -1 +1,79 @@
...
+  display: flex;
+  color: #000;
+  text-decoration: none;
+  /* padding: calc((var(--max__height) - 14px) / 2) 0; */
+  height: var(--max__height);
+  line-height: var(--max__height);
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
slavanosatskyi, line 62: move it to a variable

src/index.html
```diff
@@ -6,9 +6,58 @@
...
+      <a href="#home-page">
+        <img
+          src="./images/logo.png"
+          alt="moyo-logo"
+          class="logo"
+        >
+      </a>
+      <nav class="navigation">
+        <ul class="nav__list">
+          <li class="nav__element_of_list">
```
slavanosatskyi, line 29:
```suggestion
          <li class="nav__element_list-item">
```

slavanosatskyi (changes requested):
nice work 👍 some code styling is requried

src/index.html
```diff
@@ -6,9 +6,58 @@
...
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home-page">
+        <img
+          src="./images/logo.png"
+          alt="moyo-logo"
+          class="logo"
+        >
+      </a>
+      <nav class="navigation">
```
Viktor-Kost, line 27: you can simplify
```suggestion
      <nav class="nav">
```

src/index.html
```diff
@@ -6,9 +6,58 @@
...
+        <ul class="nav__list">
+          <li class="nav__element_list-item">
```
Viktor-Kost, line 29: simplify
```suggestion
          <li class="nav__item">
```

src/style.css
```diff
@@ -1 +1,73 @@
...
+  margin: 0 10px;
+}
+
+.nav__element_list-item:last-child {
+  margin-right: 0;
+}
+
+.nav__element_list-item:first-child {
+  margin-left: 0;
+}
```
Viktor-Kost, line 46: you can just do like this
```suggestion
.nav__item {
  list-style: none;
}

.nav__item:not(:last-child) {
  margin-right: 20px;
}
```

src/style.css
```diff
@@ -1 +1,73 @@
...
+
+.nav__link {
+  display: flex;
+  color: #000;
+  text-decoration: none;
+  height: var(--max__height);
+  line-height: var(--max__height);
+}
```
Viktor-Kost, line 54: 1) you can just use display: inline-block since you don't use flex as it usually used
2) it's not good to hardcode height for element with content inside. Use `padding: 23px 0;` instead

src/style.css
```diff
@@ -1 +1,73 @@
-/* Styles go here */
+html {
+  font-family: "Roboto", sans-serif;
+  font-style: normal;
+  --max__height: 60px;
```
Viktor-Kost, line 4: will be a redundant variable
```suggestion
```

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

src/style.css
NikitaOrliak: Wow, this is a very interesting idea, i had a lot of troubles with tests before i used separate margins, thanks a lot!

src/style.css
NikitaOrliak: I'm using display: flex; cos i wanna to put my line inder link, if i use inline-block - line shifts to the right and becomes, as it were, after the text. Like that:
image

I didn't find a better solution than to leave flex and set a specific height for the entire header. Btw, one mentor in the chat said that i can use height: 100%, so i just add max-height in a variable and can use it wherever i want.

PR #3278, opened 2023-06-01
add task solution

Shpurka:


src/index.html
```diff
@@ -1,14 +1,63 @@
...
     <h1>Moyo header</h1>
+    <header class="container">
+      <div class="nav-container">
+        <div class="nav-logo-container">
+          <a href="#home" class="logo-link">
+            <img
+            src="./images/logo.png"
+            alt="logo"
+            class="logo"
+            >
```
volodymyr-soltys97, line 23:
```suggestion
            <img
              src="./images/logo.png"
              alt="logo"
              class="logo"
            >
```

src/style.css
```diff
@@ -1 +1,93 @@
...
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+
+  --main-color: #00acdc;
+}
+
+li {
+  list-style: none;
+
```
volodymyr-soltys97, line 14:
```suggestion
```

src/style.css
```diff
@@ -1 +1,93 @@
...
+a {
+  text-decoration: none;
+  color: #000;
+}
+
+.container {
+  padding: 0 50px;
+  margin: 0 auto;
+  max-width: 1200px;
+
```
volodymyr-soltys97, line 26:
```suggestion
```

src/style.css
```diff
@@ -1 +1,93 @@
...
+a {
+  text-decoration: none;
+  color: #000;
+}
+
+.container {
+  padding: 0 50px;
+  margin: 0 auto;
+  max-width: 1200px;
```
volodymyr-soltys97, line 25: There is no need to limit the maximum width of the header, you can make it responsive

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's to fix some mistakes. Тhis text should be deleted
Знімок екрана 2023-06-01 о 17 21 59

src/style.css
Shpurka: updated all

PR #3296, opened 2023-06-05
add task solution

kyslytsyn:
- [DEMO LINK](https://kyslytsyn.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://kyslytsyn.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,62 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+          src="/images/logo.png"
+          alt="logo-img"
+          class="logo-img"
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="nav__link is-active" href="#apple">
+              Apple
+            </a>
+          </li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -9,6 +9,62 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+          src="/images/logo.png"
```
volodymyr-soltys97, line 15: The src attribute in the img tag should use a relative path, like './images/logo.png'

src/style.css
```diff
@@ -1 +1,79 @@
-/* Styles go here */
+@import url("https://fonts.googleapis.com/css2?family=Roboto&display=swap");
+
+@media (max-width: 1024px) {
+  .header {
+    max-width: 1024px;
+  }
```
volodymyr-soltys97, line 6: Remove the 'max-width' rule for the .header class inside the media query, as it's not necessary for the given task.

volodymyr-soltys97 (changes requested):
Good job
But let's make your code more perfect
Let's make it so that on large screens we have a menu with the entire available width, so that there is no such situation as on the screen
image

polosanya (approved):
👍

PR #3315, opened 2023-06-08
moyo-header

lisabogdantseva:
- [DEMO LINK](https://lisabogdantseva.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://lisabogdantseva.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,90 @@
-/* Styles go here */
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,90 @@
...
+  left: 0;
+  border-radius: 8px;
+}
+
+.nav__item:last-child {
+  margin-right: 0;
+
+}
+
+.nav__last__child {
```
anastasiiavorobiova, line 88: Avoid duplication of styles
It's better to use just pseudoclass here because we need just the last element no matter what is it

anastasiiavorobiova (approved):

PR #3337, opened 2023-06-11
Create MOYO header 

rnagiyev:
- [DEMO LINK](https://rnagiyev.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://rnagiyev.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,95 @@
+html {
+  --header-font-family: "Roboto", "sans-serif";
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
+  padding: 0;
+  --color-main: #000;
+  --color-link-hover: #00acdc;
+}
+
+.nav__list {
+  list-style: none;
+  display: flex;
+  padding: 0;
+  margin: 0;
+  height: 60px;
+}
+
+.nav {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,95 @@
...
+}
+
+.nav__item {
+  margin-right: 20px;
+}
+
+.nav__item:last-child {
+  margin-right: 0;
+}
+
```
etojeDenys, line 45:
```suggestion
.nav__item:not(:last-child) {
  margin-right: 20px;
}

```

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+.nav {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  height: 60px;
+}
+
+.header {
+  font-family: var(--header-font-family);
```
etojeDenys, line 29: we can set it in html selector

etojeDenys (changes requested):

src/style.css
rnagiyev: Done

src/style.css
rnagiyev: Done

mykhalenych (changes requested):

mykhalenych (approved):

PR #3360, opened 2023-06-16
add task solution

serdiukm:


src/index.html
```diff
@@ -5,10 +5,47 @@
...
   <body>
     <h1>Moyo header</h1>
+    <header>
+      <nav>
+        <a href="index.html"><img src="images/logo.png" alt="logo"></a>
+        <ul class="nav__list">
+
+          <li class="nav__item">
+            <a href="index.html" class="blue_link">Apple</a>
+            <div class="rectangle"></div>
```
etojeDenys, line 23: Instead of adding a separate div with class 'rectangle', use a pseudo-element (::before or ::after) in CSS to create the active link underline effect.

src/index.html
```diff
@@ -5,10 +5,47 @@
...
+          </li>
+          <li class="nav__item">
+            <a href="index.html" class="link">Samsung</a>
```
etojeDenys, line 26: For all the 'a' elements with the 'link' class, replace it with a more meaningful name like 'nav__link'.

src/index.html
```diff
@@ -5,10 +5,47 @@
...
   </head>
   <body>
     <h1>Moyo header</h1>
+    <header>
+      <nav>
+        <a href="index.html"><img src="images/logo.png" alt="logo"></a>
+        <ul class="nav__list">
+
+          <li class="nav__item">
+            <a href="index.html" class="blue_link">Apple</a>
```
etojeDenys, line 22: Replace the 'blue_link' class with a more meaningful name like 'nav__link-active'.

src/style.css
```diff
@@ -1,3 +1,66 @@
 body {
   margin: 0;
 }
+
+header {
+  font-family: Roboto, sans-serif;
+}
+
+nav {
```
etojeDenys, line 9: Consider using a more specific selector like '.nav' to avoid applying styles to other 'nav' elements that might be added to the page in the future.

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  height: 4px;
+  background: #00acdc;
+}
+
+a:hover {
+  text-decoration-line: underline;
+  text-decoration-color: #00acdc;
+}
+
+@media (max-width: 1200px) {
```
etojeDenys, line 62: Remove the '@media (max-width: 1200px)' query as the layout should fit on both 1024px and 1200px screens without any extra styles.

etojeDenys (changes requested):
and add links to your PR

etojeDenys (changes requested):
fix all prev comments and then re-request

src/index.html
serdiukm: https://github.com/serdiukm/layout_moyo-header.git I don't understand where to add the link

etojeDenys (changes requested):
if you have any questions, please, fill free to ask about them in slack chat 🙏

PR #3368, opened 2023-06-19
Develop

bormaxvel:
https://bormaxvel.github.io/layout_moyo-header/

https://bormaxvel.github.io/layout_moyo-header/report/html_report/


src/index.html
```diff
@@ -7,8 +7,33 @@
...
+    <header class="header">
+      <a href="#home" class="logo_link">
+        <img
+        class="logo_img"
+        src="images\logo.png"
+        alt="logo">
+      </a>
+      <nav class="nav">
+        <ul class="nav_list">
+          <li class="nav_list_item"><a class="nav_list_item_link is-active" href="/">Apple</a></li>
```
sTorba24, line 27:
```suggestion
          <li class="nav__item">
            <a 
              class="nav__link is-active" 
              href="/"
            >Apple</a>
          </li>
```

sTorba24: Fix formatting, also in BEM you don't need to follow name of class before.
Your block is `nav`. Other things inside - elements of block or can be block.
naming should be `{block}__{element}`. Doesn't matter how deep element is.

src/style.css
```diff
@@ -1 +1,71 @@
-/* Styles go here */
+body {
+  margin: 0;
+  --accent_colour: #00acdc;
```
sTorba24, line 3: move variables to :root selector

src/style.css
```diff
@@ -1 +1,71 @@
...
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  font-style: normal;
```
sTorba24, line 7: default value, can be removed

src/style.css
```diff
@@ -1 +1,71 @@
...
+
+.nav_list_item_link {
+  height: 100%;
+  text-decoration: none;
+  color: black;
+  display: block;
+}
+
+.nav_list_item {
+  height: 60px;
```
sTorba24, line 42: don't set element height directly, you can do it with elements inside

sTorba24 (changes requested):
Almost good! Lets improve it

src/style.css
bormaxvel: I agree, in this case it was inappropriate

src/style.css
bormaxvel: ok

src/style.css
bormaxvel: I don't quite understand what you mean. If you remove these 4 lines, the font will change to the standard one

sTorba24: i meant only last one, sorry for confusing

src/index.html
bormaxvel: I'm not quite sure about the formatting. If you write everything so widely, it will be inconvenient to read. Our list simply won't fit on one screen, and it's hard to follow all the text with eyes because it's far from each other. 
Maybe it's better to format it like this?
```
          <li class="nav_list_item">
            <a class="nav_list_item_link is-active" href="/">
              Apple
            </a>
          </li>
```
Or like this
```
          <li class="nav_list_item">
            <a class="nav_list_item_link" href="/">
              Samsung</a>
          </li>
```

src/style.css
etojeDenys: still not fixed

src/index.html
```diff
@@ -7,8 +7,65 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <!-- <link rel="preconnect"
+    href="https://fonts.gstatic.com"
```
etojeDenys, line 12: remove all comments

etojeDenys (changes requested):

bormaxvel:
I think I've fixed everything

.linthtmlrc.json
loralevitska, line 1: For future - Do not add extra files to PR. To add the file to a commit use git add src/.

src/index.html
```diff
@@ -7,8 +7,61 @@
...
   </head>
-  <body>
-    <h1>Moyo header</h1>
+  <body id="#home">
+    <header class="header">
+      <a href="#home" class="logo_link">
+        <img
+        class="logo_img"
+        src="images\logo.png"
+        alt="logo">
```
loralevitska, line 19:
```suggestion
        <img
          class="logo_img"
          src="images\logo.png"
          alt="logo"
        >
```

src/index.html
```diff
@@ -7,8 +7,61 @@
...
+      </a>
+      <nav class="nav">
+        <ul class="nav_list">
+          <li class="nav_item">
+            <a class="nav_link is-active" href="/">
```
loralevitska, line 24: According to BEM, it is correct to add 2 underscores, because with this syntax, you currently do not have an element in blocks, but only blocks whose names consist of two words. Fix all cases with BEM naming please
```suggestion
        <ul class="nav__list">
          <li class="nav__item">
            <a class="nav__link is-active" href="/">
```

loralevitska (approved):

PR #3399, opened 2023-06-26
add task solution

bereg9187:
[DEMO LINK](httpsbereg9187.github.iolayout_moyo-header)
- [TEST REPORT LINK](httpsbereg9187.github.iolayout_moyo-headerreporthtml_report)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,68 @@
-body {
+:root {
+  --current-text: #000;
+  --brend-color: #00acdc;
+}
+
+.body {
   margin: 0;
+  font-family: "Roboto", "Arial", sans-serif;
+  font-size: 12px;
+}
+
+.header {
+  display: flex;
+  padding: 0 50px;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
+}
+
+.nav__list {
+  list-style: none;
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+a {
+  text-decoration: none;
+  position: relative;
+  text-transform: uppercase;
+  color: var(--current-text);
+  height: 60px;
+  display: flex;
+  align-items: center;
+}
+
+a:hover {
+  color: var(--brend-color);
+}
+
+a[href^="#home"] {
+  line-height: 0;
+}
+
+.is-active {
+  color: var(--brend-color);
+}
+
+.is-active::after {
+  content: "";
+  width: 100%;
+  height: 4px;
+  background-color: var(--brend-color);
+  display: block;
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  left: 0;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,68 @@
...
+}
+
+.nav__list {
+  list-style: none;
+  display: flex;
   margin: 0;
+  padding: 0;
+}
+
+a {
```
mykhalenych, line 31: use classes instead of tag selector

mykhalenych (changes requested):
Nice work, but check comment, also check your links in description, must be link to demo and test

src/style.css
bereg9187: I cant add class to tag "a" because   "29:48   error  Only 2 attributes per line are permitted  attr-new-line "

bereg9187: It "Laptops & computers" has 3 atribute

bereg9187: please check

src/style.css
```diff
@@ -1,3 +1,69 @@
-body {
+:root {
+  --current-text: #000;
+  --brend-color: #00acdc;
+}
+
+.body {
+  margin: 0;
+  font-family: "Roboto", "Arial", sans-serif;
+  font-size: 12px;
```
etojeDenys, line 9: better to move these styles to html selector

usually, we just reset some default styles in the body

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+
+.nav__list {
+  list-style: none;
+  display: flex;
   margin: 0;
+  padding: 0;
+}
+
+.nav__list__link,
+a[href^="#laptops"] {
```
etojeDenys, line 32: why do you need this tag here

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  align-items: center;
+}
+
+a:hover {
+  color: var(--brend-color);
+}
+
+a[href^="#home"] {
+  line-height: 0;
+}
```
etojeDenys, line 48: looks like a redundant selector

etojeDenys (changes requested):

src/style.css
bereg9187: i will fix it

src/style.css
bereg9187: becuse i cant add class to tag "a" because "29:48 error Only 2 attributes per line are permitted attr-new-line "

src/style.css
```diff
@@ -1,3 +1,67 @@
-body {
+:root {
+  --current-text: #000;
+  --brend-color: #00acdc;
```
VitaliyBodnar, line 3: variable name mistake
```suggestion
  --brand-color: #00acdc;
```

VitaliyBodnar (changes requested):
Well done )

PR #3416, opened 2023-06-29
add task solution

1umamaster:
- [DEMO LINK](https://1umamaster.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://1umamaster.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,54 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  background-color: black;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background-color: white;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,54 @@
...
+.nav__link:hover,
+.is-active {
+  color: #00acdc;
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  height: 4px;
+  margin-top: -4px;
```
Andriy-Shymkiv, line 51: `margin-top: -4px` ? strange decision. maybee let's try to do the same with `position` property? set `position: relative` for `is-active` , `position: absolute` here, and try to play with `top` and `bottom`

Andriy-Shymkiv (approved):
very good job! appoved in advance

src/style.css
1umamaster: Thx for feedback. Just updated.

PR #3421, opened 2023-06-30
add task solution

myehor:


src/style.css
```diff
@@ -1,3 +1,68 @@
...
 body {
   margin: 0;
 }
+
+.header {
+  justify-content: space-between;
+  align-items: center;
+  display: flex;
+  padding: 0 50px;
+  /* min-width: 1024px; */
```
OleziO, line 17: Remove al comments

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  list-style: none;
+}
+
+.logo__img {
+  width: 40px;
+  height: 40px;
+}
+
+.nav :hover {
+  color: #00acdc;
```
OleziO, line 46: Move all repeatable colors to the CSS variable.

OleziO (changes requested):

src/style.css
```diff
@@ -1,3 +1,68 @@
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
 body {
   margin: 0;
+  --task-color: #00acdc;
```
OleziO, line 10: Move all variables to the :root selector.

OleziO (changes requested):

loralevitska (approved):
Well done)

PR #3425, opened 2023-07-01
Develop

robsza1122:
ready to check


src/index.html
```diff
@@ -7,8 +7,55 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
+
   <body>
     <h1>Moyo header</h1>
```
darokrk, line 16: Think this header `<h1>Moyo header</h1>`, regarding the designs should be removed

Screenshot 2023-07-04 at 07 44 21

src/index.html
```diff
@@ -7,8 +7,55 @@
...
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
```
darokrk, line 35: think class is-active does not work, please make sure that active link has blue color

image

backstopConfig.js
```diff
@@ -47,8 +47,8 @@ const config = {
     },
     {
       ...basic,
-      label: 'Link with class_is-active',
-      selectors: ['a.is-active'],
+      label: 'Link with class_nav__item',
+      selectors: ['a.nav__item'],
```
darokrk, line 51: Are you sure you should change something here? think after those changes you have problem with active link :)

darokrk (changes requested):
Good PR and the layout looks fine, small fixes are needed and we are ready to move next 👍🏽

backstopConfig.js
robsza1122: I did replacement everywhere by mistake. :)

robsza1122:
Ready to check and let's move on!.

src/style.css
```diff
@@ -1,3 +1,59 @@
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  --color: #00acdc;
```
DorotaLeniecDev, line 4: It is a good practice to name colors more descriptively so your code is readable
```suggestion
  --color-blue: #00acdc;
```

DorotaLeniecDev (changes requested):
You can remove Roboto directory as you are using fonts by link, also few small improvements suggestion from my side

darokrk (approved):
Good job 👏 approved 🏆

PR #3426, opened 2023-07-01
MOYO Header Solution

WaveOfDandelions:
[HEADER](https://WaveOfDandelions.github.io/layout_moyo-header/)

[TEST](https://WaveOfDandelions.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,50 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <div class="logo">
+        <img src="./images/logo.png" alt="logotype">
+      </div>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+
+          <li class="nav__item">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,50 @@
...
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <div class="logo">
+        <img src="./images/logo.png" alt="logotype">
```
choeqq, line 17: I think we need `< a/>` tag around `img` - our logo probably links to main page 😉

src/style.css
```diff
@@ -1,3 +1,69 @@
+:root {
+  --main-color-font: #000;
+  --main-color-font-is-hover: #00acdc;
+}
+
 body {
```
choeqq, line 6: I would probably provide more generous CSS styles reset like:
```
*, *::after, *::before {
  box-sizing: border-box;
  margin: 0;
  padding: 0;
}
```
but that's a good start 😄 

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
+}
+
+.nav__item {
+  list-style: none;
+}
+
+.nav__item:not(:last-child) {
```
choeqq, line 38: we can also just remove margin from last child 😄

src/style.css
```diff
@@ -1,3 +1,69 @@
+:root {
```
choeqq, line 1: props for using CSS variables 👍

choeqq (approved):
It would be useful to provide the TEST and HEADER links in PR description, I updated it for you 😉

src/style.css
WaveOfDandelions: Is it better to use normalize css for that?

choeqq: For this kind of project I would stick to the basics, and those simple styles are more than enough, but for larger projects I would definitely take that into consideration 😉

mvjl000 (approved):

PR #3437, opened 2023-07-03
Solving task

domalewski:


src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  height: 60px;
+  cursor: pointer;
+}
+
+.is-active,
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active:after {
```
choeqq, line 57: to create `after` pseudoelement we need to use `::` instead of `:` which is used for pseudoclasses

choeqq, line 57: we are missing `display: block` property, and specific width instead of `100%`

src/style.css
```diff
@@ -1,3 +1,66 @@
 body {
```
choeqq, line 1: I would probably provide more generous CSS styles reset like:
```
*, *::after, *::before {
  box-sizing: border-box;
  margin: 0;
  padding: 0;
}
```
but that's a good start 😄 

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  position: relative;
+  padding: 0 50px;
+}
+
+.logo__image {
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
```
choeqq, line 18: I think I would use `font family` property here, instead of `nav__link` class

src/index.html
```diff
@@ -3,12 +3,53 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
```
choeqq, line 10:
```
    <link
    rel="stylesheet"
    href="./style.css">
    <link
    rel="preconnect"
    href="https://fonts.googleapis.com"
    >
    <link
    rel="preconnect"
    href="https://fonts.gstatic.com"
    crossorigin="anonymous "
    >
    <link
    href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
    rel="stylesheet"
    >
```
here is a proper way to import google font 😉

src/index.html
```diff
@@ -3,12 +3,53 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="logo__link">
```
choeqq, line 14: I am thinking about wrapping this `<a />` tag with `<div>` so it is easier to move around, what do you think?
Typically when we use flexbox we want to operate on block elements

choeqq (changes requested):
It would be useful to provide the TEST and HEADER links in PR description, I updated it for you 😉

src/index.html
domalewski: but but ... do I need this gstatic.com stuff?

choeqq: I think so, but you can test it both with and without those links 😉

domalewski:
[HEADER](https://domalewski.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://domalewski.github.io/layout_moyo-header/report/html_report/)

choeqq (approved):
I tested the solution on my computer - it's alright, so I am approving it, we will see if problems will still persist in next tasks

mvjl000 (approved):

PR #3439, opened 2023-07-03
Marekder solution

marekder:
- [DEMO LINK](https://marekder.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://marekder.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,4 +1,44 @@
 # Moyo header
-
+Replace `<your_account>` with your Github username and copy the links to Pull Request description:
 - [DEMO LINK](https://marekder.github.io/layout_moyo-header/)
 - [TEST REPORT LINK](https://marekder.github.io/layout_moyo-header/report/html_report/)
+
+> Follow [these instructions](https://mate-academy.github.io/layout_task-guideline/#how-to-solve-the-layout-tasks-on-github)
+
+## ❗️❗️❗️ DON'T FORGET TO PROOFREAD YOUR CODE WITH [CHECKLIST](https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md) BEFORE SENDING YOUR PULL REQUEST❗️❗️❗️
+
+## The task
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -3,12 +3,54 @@
...
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
```
choeqq, line 15: good that you wrapped `<a />` tag with `<div/>` 👍

src/index.html
```diff
@@ -3,12 +3,54 @@
...
+        <a href="index.html">
```
choeqq, line 16: `href="#"` is actually the same thing - please go with `"#"` if you want to link to top of the page

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  position: relative;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px #0000000d;
+}
+
+.logo {
+  display: block;
```
choeqq, line 18: div's are `display: block` by deafult, so we don't need to overrite that

src/style.css
choeqq, line 1: if you'll import only 1 weight of font it won't be necessary to assign it every time - it will always be 500, and that's seems to be the reason your tests are failing - your font is too thin sometimes😄

choeqq (changes requested):

readme.md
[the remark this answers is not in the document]
marekder: Still tests on GitHub passed, page deployed, on the other hand in terminal on VSC - 4 tests failed, I cannot find any reason for that.

PR #3441, opened 2023-07-03
Solution

Erchu420:
- [DEMO LINK](https://Erchu420.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Erchu420.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,8 +13,11 @@
   </head>
   <body>
     <header class="header">
-      <a href="#" class="logo-link">
-      </a>
+      <div class="logo-link">
+        <a href="#">
+          <img src="images/logo.png" alt="logo">
+        </a>
+      </div>
       <nav class="nav">
         <ul class="nav-list">
           <li class="nav-item">
--- a/src/index.html
+++ b/src/index.html
@@ -29,8 +29,9 @@
           <li class="nav-item">
             <a href="#smartphones" class="nav-link">Smartphones</a>
           </li>
-          <li class="nav-item" data-qa="hover">
-            <a href="#laptopsandcomputers" class="nav-link">Laptops & Computers</a>
+          <li class="nav-item">
+            <a href="#laptopsandcomputers" data-qa="hover"
+             class="nav-link">Laptops & Computers</a>
           </li>
           <li class="nav-item">
             <a href="#gadgets" class="nav-link">Gadgets</a>
--- a/src/index.html
+++ b/src/index.html
@@ -1,6 +1,9 @@
 <!doctype html>
 <html lang="en">
   <head>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -9,6 +12,37 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="logo-link">
```
choeqq, line 16: I think it would be more useful to include `` tag in link - that's what it's for, instead of setting element's background-image 😄

choeqq, line 16: Please wrap `<a/>` with `<div />` - div's are more predictable when it comes to `flex-box`, and moving things around the page generally

src/index.html
```diff
@@ -9,6 +12,37 @@
...
+            <a href="#apple" class="nav-link is-active">Apple</a>
+          </li>
+          <li class="nav-item">
+            <a href="#samsung" class="nav-link">Samsung</a>
+          </li>
+          <li class="nav-item">
+            <a href="#smartphones" class="nav-link">Smartphones</a>
+          </li>
+          <li class="nav-item" data-qa="hover">
+            <a href="#laptopsandcomputers" class="nav-link">Laptops & Computers</a>
```
choeqq, line 30: please move the `data-qa="hover"` attribute to `< a/>` tag - it's more explicit like that

src/style.css
```diff
@@ -1,3 +1,58 @@
...
+}
+
+.nav-item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav-link {
+  text-decoration: none;
+  text-transform: uppercase;
+  color: rgba(0,0,0);
```
choeqq, line 36: you can just use "black" - but good that you're keeping everything within one convention😄

choeqq (changes requested):

src/index.html
Erchu420: done in solution2 it's <div><a></a>/</div>

Erchu420: done in solution2

src/style.css
Erchu420: done in solution2

choeqq (approved):
To address your concerns about `` tag and background-image property:

Both solutions are good, but when it comes to learning basics it is far more useful to learn how to use `` tag, and I think it is way more common to see this kind of implementation in most of the projects

[More about img tag vs background-image](https://stackoverflow.com/questions/492809/when-to-use-img-vs-css-background-image)|

Good job! 😉

src/index.html
Erchu420: done in solution3

choeqq (approved):

mvjl000 (approved):

PR #3449, opened 2023-07-03
add task solution

klaudia-klis:
- [DEMO LINK](https://klaudia-klis.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://klaudia-klis.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,10 +1,13 @@
+* {
+  box-sizing: border-box;
+}
+
 body {
   font-family: Roboto, sans-serif;
-  font-weight: 500;
   margin: 0;
 }
 
-header {
+.header {
   display: flex;
   justify-content: space-between;
   padding: 0 50px;
@@ -12,7 +15,6 @@ header {
 }
 
 .logo {
-  margin: 10px 0;
   width: 40px;
   height: 40px;
 }
@@ -21,24 +23,20 @@ header {
   display: flex;
   margin: 0;
   padding: 0;
-  text-transform: uppercase;
-  list-style-type: none;
 }
 
-.nav__item {
+li {
   list-style: none;
 }
 
 .nav__link {
   color: black;
-  padding: 23px 0;
   text-decoration: none;
   text-transform: uppercase;
   font-size: 12px;
-  line-height: 14px;
-  display: flex;
-  align-items: center;
+  line-height: 60px;
   font-weight: 500;
+  display: inline-block;
 }
 
 .nav__item:not(:last-child) {
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,66 @@
 body {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
   margin: 0;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  align-items: center;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  font-weight: 500;
   margin: 0;
 }
+
+header {
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  align-items: center;
+}
```
asakevych, line 12: 1. it's a good practice to write all styles using classes -> add class to tag header 
2. add height here

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+
+.logo {
+  margin: 10px 0;
```
asakevych, line 15: margins are unnecessary, because header has `align-items: center`

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  text-transform: uppercase;
```
asakevych, line 24: unnecessary, you have uppercase on links

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  list-style-type: none;
+}
+
+.nav__item {
+  list-style: none;
+}
```
asakevych, line 30: `list-style: none` should be on <ul> tag, not <li>, then `list-style-type: none;` is unnecessary

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+
+.nav__link {
+  color: black;
+  padding: 23px 0;
```
asakevych, line 34: it's better not to use padding here, try to make height 100% and just ceter text

src/style.css
```diff
@@ -1,3 +1,66 @@
 body {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
```
asakevych, line 3: you set font-weight for each link, it's better not to set it for whole body

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+
+.nav__link {
+  color: black;
+  padding: 23px 0;
+  text-decoration: none;
+  text-transform: uppercase;
+  font-size: 12px;
+  line-height: 14px;
+  display: flex;
+  align-items: center;
```
asakevych, line 40: no sense to use flex here

asakevych (changes requested):
Good job! just some small changes needed :)

asakevych (changes requested):
So far so good, but still some small changes are requested 😉

src/style.css
```diff
@@ -1,3 +1,61 @@
...
+
+.logo {
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
+  display: flex;
   margin: 0;
+  padding: 0;
```
choeqq, line 25: you can also just reset that in `*` or body, so none of the elements has that 😉

choeqq (approved):
Wel done, one small suggestion, you don't need to apply this, but please keep that in mind 🚀

mvjl000 (approved):

PR #3451, opened 2023-07-03
add task solution

ogospadzipalina:
[DEMO LINK](https://ogospadzipalina.github.io/layout_moyo-header/)
 [TEST REPORT LINK](https://ogospadzipalina.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,49 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+    crossorigin="-">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logonpm">
+        <a href="#">
+          <img class="logo" src="images/logo.png"
+           alt="moyo logo">
+        </a>
+      </div>
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#Apple" class="nav__link  is-active">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Samsung" class="nav__link">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Smartphones" class="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Laptops" class="nav__link"
+            data-qa="hover">Laptops & Computers</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Gadgets" class="nav__link">Gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Tablets" class="nav__link">Tablets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Photo" class="nav__link">Photo</a>
+          </li>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,67 @@
...
+.is-active::after {
+  position: absolute;
+  content: "";
+  display: block;
+  height: 4px;
+  width: 37px;
+  border-radius: 8px;
+  background-color: #00acdc;
+  bottom: 0;
+}
```
asakevych, line 67: it's the same color, you can create variable and use it in both places

src/style.css
```diff
@@ -1,3 +1,67 @@
 body {
   margin: 0;
 }
+
+.header {
+  justify-content: space-between;
+  display: flex;
+  align-items: center;
+  padding: 0 50px;
+}
```
asakevych, line 10: add height of your header, then you can just set links to have 100%

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+
+.logo {
+  display: flex;
+  align-items: center;
```
asakevych, line 14: flex not needed

src/index.html
```diff
@@ -7,8 +7,49 @@
...
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+    crossorigin="-">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logonpm">
```
asakevych, line 17: not best name for class here ;)

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.nav__link:link {
+  text-decoration: none;
+  display: flex;
+  align-items: center;
+  position: relative;
+  height: 60px;
```
asakevych, line 45: it's better to set height for the whole header and make here just 100%

src/style.css
```diff
@@ -1,3 +1,67 @@
 body {
   margin: 0;
 }
```
asakevych, line 3: 1. add font-family here
2. it's a good practice to set for all elements `box-sizing: border-box`

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+}
+
+.nav__item {
+  margin-left: 20px;
+  position: relative;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
```
asakevych, line 26: position is not needed here and its better to write `.nav__item:not(:last-child) { margin-left: 20px ;}` ,  check here https://developer.mozilla.org/en-US/docs/Web/CSS/:not

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+.nav .is-active {
+  color: hsl(193, 100%, 43%);
+}
+
+.is-active::after {
+  position: absolute;
+  content: "";
+  display: block;
+  height: 4px;
+  width: 37px;
```
asakevych, line 63: width here should be 100%, if active would have loger text, width of underline would also change

asakevych (changes requested):
Great job! just some small changes needed :)

src/style.css
ogospadzipalina: @asakevych when I do it like this, my clickable area around link stops being 60px and ::after appears underneath the text, not a bit below as in an example, so it is not accepted by tests.

Oskinavara (approved):
Everything from the previous review was fixed. Good job. I would avoid writing comments in code in Polish though.

mvjl000 (approved):

PR #3452, opened 2023-07-03
add task solution

ihor-pavlov:
[DEMO LINK](https://ihor-pavlov.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://ihor-pavlov.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -50,6 +50,5 @@
         </ul>
       </nav>
     </header>
-    <h1>Moyo header</h1>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -9,7 +9,6 @@ header {
   justify-content: space-between;
   background-color: #fff;
   padding: 0 50px;
-  max-width: 1200px;
 }
 
 .go-home {


src/style.css
```diff
@@ -1,3 +1,80 @@
...
+
+.nav__item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav__link {
+  display: flex;
+  align-items: center;
+  text-decoration: none;
+  line-height: normal;
```
Andriy-Shymkiv, line 54: `normal` is a default value for `line-height`, so you can remove it

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  font-size: 12px;
+  font-weight: 500;
+  color: #000;
+  font-style: normal;
```
Andriy-Shymkiv, line 58: same for `font-style`, remove it

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  height: 100%;
```
Andriy-Shymkiv, line 59: why you need to set this `height`?

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.nav__item {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
```
Andriy-Shymkiv, line 43: you don't need this `height: 100%;`

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  display: flex;
+  justify-content: space-between;
+  background-color: #fff;
+  padding: 0 50px;
+  max-width: 1200px;
+}
+
+.go-home {
+  margin: 10px 0;
+  height: 40px;
```
Andriy-Shymkiv, line 17: no need to set `height` for this because you doing this for class `logo`

Andriy-Shymkiv: and you can remove `go-home` class from your link and set `margin` for logo as well

Andriy-Shymkiv (changes requested):
good job! left comments

src/style.css
ihor-pavlov: When I remove this height or this class, the logo remains 40px, but the link increases to 64px, the header increases as well, but there is a condition about 60px.

src/style.css
ihor-pavlov: Because links should have clickable area above and below the text. If I remove this height, clickable area disappears.

src/style.css
ihor-pavlov: If I remove this height, clickable area disappears and the underline of the active page rises to the text.

src/style.css
ihor-pavlov: > why you need to set this `height`?

Because links should have clickable area above and below the text. If I remove this height, clickable area disappears.

src/style.css
ihor-pavlov: > you don't need this `height: 100%;`

If I remove this height, clickable area disappears and the underline of the active page rises to the text.

src/style.css
ihor-pavlov: > no need to set `height` for this because you doing this for class `logo`

When I remove this height or this class, the logo remains 40px, but the link increases to 64px, the header increases as well, but there is a condition about 60px.

Andriy-Shymkiv: so in such case you can remove height from `logo` class, because there is no sense to set it twice

Andriy-Shymkiv (changes requested):

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
+}
+
+.nav {
+  display: flex;
+}
```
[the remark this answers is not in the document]
ihor-pavlov: > you don't need this prop, because you setting flex-box to `header`

Automated tests require two separate classes: .nav and .nav__list. I have to specify them and give them this feature and 100% height, because clickable area above and below the text.

src/index.html
```diff
@@ -6,9 +6,50 @@
...
+          <li class="nav__item">
+            <a class="nav__link" href="#photo">Photo</a>
+          </li>
+          <li class="nav__item">
+            <a class="nav__link" href="#video">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
     <h1>Moyo header</h1>
```
volodymyr-soltys97, line 53: Also you can remove it
```suggestion
```

src/style.css
```diff
@@ -1,3 +1,75 @@
...
   margin: 0;
+  padding: 0;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
+  background-color: #fff;
+  padding: 0 50px;
+  max-width: 1200px;
```
volodymyr-soltys97, line 12:
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your solution
You need to make the menu stretch to the full width of the screen, so you don't need to limit the header to the maximum width
image

Andriy-Shymkiv (approved):
good job!

PR #3455, opened 2023-07-03
Develop

Sancz0pansa:
[DEMO LINK](https://Sancz0pansa.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://Sancz0pansa.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,76 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+}
+
+.img__link {
+  display: block;
+  height: 40px;
+}
+
+header {
+  background-color: #fff;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 60px;
+  padding: 0 50px;
+}
+
+.nav {
+  display: block;
+}
+
+.nav__list {
+  margin: 0;
+  list-style: none;
+  display: flex;
+  padding: 0;
+}
+
+.nav_img {
+  height: 40px;
+  width: 40px;
+  margin-right: auto;
+}
+
+.nav__item {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  display: block;
+  height: 40px;
+}
+
+header {
+  background-color: #fff;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 60px;
```
mvjl000, line 18: You are not supposed to set height for header. Instead let links to dictate the height of it ;)

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  font-weight: 500;
+  font-size: 12px;
+}
+
+.img__link {
+  display: block;
+  height: 40px;
+}
+
+header {
```
mvjl000, line 13: I think it would be better to add a class to a header. For instance `class="header"` and style it with class selector `.header {}`

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  background-color: #fff;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 60px;
+  padding: 0 50px;
+}
+
+.nav {
+  display: block;
```
mvjl000, line 23: I'm not convinced if it is necessary. Does anything break when you remove it?

mvjl000 (changes requested):
Your code needs just a couple of improvements but besides of that, it looks great!
Good job ;)

src/style.css
Sancz0pansa: yeah, it's not necessary.

choeqq (approved):
Looks good to me! 👍

mvjl000 (approved):

PR #3463, opened 2023-07-03
add task solution

HaniaNassalska:
https://hanianassalska.github.io/layout_moyo-header/
https://hanianassalska.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,82 @@
+html {
+  box-sizing: border-box;
+}
+
+*,
+::after,
+::before {
+  box-sizing: inherit;
+}
+
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  height: 60px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,82 @@
...
   margin: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  height: 60px;
+  align-items: center;
+  margin: 10px 50px;
```
Oskinavara, line 21: This vertical 10px margin is not necessary. The header should be only 60px in height and it is like that without the header :)

Oskinavara (changes requested):

src/style.css
HaniaNassalska: Great! Thank you for your advice!

choeqq (approved):
Looks good to me! 😄 
GJ 😉

mvjl000 (approved):

PR #3473, opened 2023-07-04
add task solution

GavriliukArtem:


src/style.css
```diff
@@ -1,3 +1,79 @@
...
+  list-style: none;
+}
+
+.nav__item {
+  margin-left: 20px;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
```
Andriy-Shymkiv, line 40: here you can go like this instead:
```css
.nav__item:not(:first-child) {
	margin-left: 20px;
} 
```

andriy-shymkiv (approved):
left comments. approved in advance. good job!

PR #3475, opened 2023-07-04
add task solution

DanielToporek:
https://DanielToporek.github.io/layout_moyo-header/
https://DanielToporek.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -19,17 +19,15 @@
           alt="logo of moyo-header">
       </a>
       <nav class="nav">
-        <ul>
-          <a class="link1 left is-active" href="#apple">APPLE</a>
-          <a class="link" href="#samsung">SAMSUNG</a>
-          <a class="link" href="#smartphones">SMARTPHONES</a>
-          <a class="link" data-qa="hover"
-            href="#laptops">LAPTOPS & COMPUTERS</a>
-          <a class="link" href="#gadgets">GADGETS</a>
-          <a class="link" href="#tablets">TABLETS</a>
-          <a class="link" href="#photo">PHOTO</a>
-          <a class="link right " href="#video">VIDEO</a>
-        </ul>
+        <a class="nav__link is-active" href="#apple">APPLE</a>
+        <a class="nav__link" href="#samsung">SAMSUNG</a>
+        <a class="nav__link" href="#smartphones">SMARTPHONES</a>
+        <a class="nav__link" data-qa="hover"
+          href="#laptops">LAPTOPS & COMPUTERS</a>
+        <a class="nav__link" href="#gadgets">GADGETS</a>
+        <a class="nav__link" href="#tablets">TABLETS</a>
+        <a class="nav__link" href="#photo">PHOTO</a>
+        <a class="nav__link" href="#video">VIDEO</a>
       </nav>
     </header>
   </body>
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,31 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+      crossorigin="anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo" >
+        <img class="image" src="images\logo.png"
+          alt="logo of moyo-header">
+      </a>
+      <nav class="nav">
+        <ul>
+          <a class="link1 left is-active" href="#apple">APPLE</a>
+          <a class="link" href="#samsung">SAMSUNG</a>
+          <a class="link" href="#smartphones">SMARTPHONES</a>
+          <a class="link" data-qa="hover"
+            href="#laptops">LAPTOPS & COMPUTERS</a>
+          <a class="link" href="#gadgets">GADGETS</a>
+          <a class="link" href="#tablets">TABLETS</a>
+          <a class="link" href="#photo">PHOTO</a>
+          <a class="link right " href="#video">VIDEO</a>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -20,30 +20,30 @@
       </a>
       <nav class="nav">
         <ul class="nav__ul">
-          <li class="nav__list" >
+          <li>
             <a class="nav__link is-active" href="#apple">APPLE</a>
           </li>
-          <li class="nav__list" >
+          <li>
             <a class="nav__link" href="#samsung">SAMSUNG</a>
           </li>
-          <li class="nav__list">
+          <li>
             <a class="nav__link" href="#smartphones">SMARTPHONES</a>
           </li>
-          <li class="nav__list" >
+          <li>
             <a class="nav__link" data-qa="hover"
             href="#laptops">LAPTOPS & COMPUTERS</a>
           </li>
-          <li class="nav__list">
+          <li>
             <a class="nav__link" href="#gadgets">GADGETS</a>
           </li>
-          <li class="nav__list">
+          <li>
             <a class="nav__link" href="#tablets">TABLETS</a>
           </li>
-          <li class="nav__list">
+          <li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -6,9 +6,31 @@
...
+          <a class="link1 left is-active" href="#apple">APPLE</a>
+          <a class="link" href="#samsung">SAMSUNG</a>
+          <a class="link" href="#smartphones">SMARTPHONES</a>
+          <a class="link" data-qa="hover"
+            href="#laptops">LAPTOPS & COMPUTERS</a>
+          <a class="link" href="#gadgets">GADGETS</a>
+          <a class="link" href="#tablets">TABLETS</a>
+          <a class="link" href="#photo">PHOTO</a>
+          <a class="link right " href="#video">VIDEO</a>
+        </ul>
```
asakevych, line 32: 1. please remember about list structure, `<ul>` should have children `<li>` elements,
2. remember about BEM block__element -> if you have links inside block nav, so class should be `.nav__link`
3.  left, right classes are not needed

src/style.css
```diff
@@ -1,3 +1,82 @@
 body {
   margin: 0;
+  margin-right: 0;
```
asakevych, line 3: if you have all margins 0, so you don't need to set margi-right once more

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  width: auto;
```
asakevych, line 4: unnecessary here, it's default value :)

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  font-family: Roboto, sans-serif;
+}
+
+.logo {
+  display: flex;
+  text-decoration: none;
+  margin: 10px 0;
```
asakevych, line 11: unnecessary here, you have `align-items: center` on header, so the position of logo would be right

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  align-items: center;
+  text-align: -webkit-match-parent;
+  margin-right: 20px;
+}
+
+.right {
+  text-decoration: none;
+  color: black;
+}
+
```
asakevych, line 46: unnecessary, you have the same styles for all your links

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+ul {
+  display: flex;
+  height: 100%;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0;
+  margin: 0;
+}
```
asakevych, line 54: it's a good practice to write all styles using classes, better chenage it to `.nav_list`

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+.is-active {
+  color: #00acdc;
+  position: relative;
+  text-decoration: none;
+  font-family: Roboto, sans-serif;
+  display: flex;
+  height: 100%;
+  align-items: center;
+  text-align: -webkit-match-parent;
+  margin-right: 20px;
```
asakevych, line 69: unnecessary to repeat all styles from .link , just add `position: relative`

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+}
+
+.link:not(:last-child) {
+  text-decoration: none;
+  color: black;
+  font-family: Roboto, sans-serif;
+  display: flex;
+  height: 100%;
+  align-items: center;
+  text-align: -webkit-match-parent;
```
asakevych, line 38: unnecessary here, it works without this property ;)

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  margin-right: 20px;
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  height: 4px;
+  width: 37px;
```
asakevych, line 76: make width 100%, then when other element woukd be active with longer or shorter text, your underline will fit it

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  border-radius: 8px;
+  background-color: #00acdc;
```
asakevych, line 78: create a variable for this color, it is used in several places

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+
+.nav {
+  display: flex;
+  height: 100%;
+}
+
+.link:not(:last-child) {
+  text-decoration: none;
+  color: black;
+  font-family: Roboto, sans-serif;
```
asakevych, line 34: you can add font-family to body styles

asakevych (changes requested):
Great job! just some small chenges needed 👍

src/index.html
```diff
@@ -6,9 +6,29 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo" >
+        <img class="image" src="images\logo.png"
+          alt="logo of moyo-header">
+      </a>
+      <nav class="nav">
+        <a class="nav__link is-active" href="#apple">APPLE</a>
```
Oskinavara, line 22: This is still not fixed. Links should be a `<ul>` list which contains multiple `<li>` elements.

Oskinavara (changes requested):

src/index.html
```diff
@@ -6,9 +6,47 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo" >
+        <img class="image" src="images\logo.png"
+          alt="logo of moyo-header">
+      </a>
+      <nav class="nav">
+        <ul class="nav__ul">
+          <li class="nav__list" >
```
mvjl000, line 23: As I can see in your styles, you are not using the `nav__list` class selector anywhere. You could get rid of that, it will make the HTML look cleaner ;)

P.S. if you were to use such a class for this `li` a better name for it would be something like `nav__item`

src/style.css
```diff
@@ -1,3 +1,72 @@
...
+.nav__link {
+  text-decoration: none;
+  color: black;
+  display: flex;
+  height: 100%;
+  margin-right: 20px;
+  align-items: center;
+}
+
+.nav__link--last {
```
mvjl000, line 44: There is just one difference between styles applied to `.nav__link` and `.nav__link--last` which is margin-right set to 0 in --last.
If I were you, I would simply put only `margin-right: 0;` to --last
It would look like this:
`
.nav__link--last {
  margin-right: 0;
}
`
and in the HTML set two classes to the last element
`<a class="nav__link nav__link--last" href="#video">VIDEO</a>`

mvjl000: This approach is better because if you wanted to change some styles of that link, you would only change them in one place - `.nav__link`

mvjl000 (changes requested):

src/index.html
```diff
@@ -6,9 +6,47 @@
...
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo" >
+        <img class="image" src="images\logo.png"
+          alt="logo of moyo-header">
+      </a>
+      <nav class="nav">
+        <ul class="nav__ul">
+          <li>
+            <a class="nav__link is-active" href="#apple">APPLE</a>
```
mvjl000, line 24: One thing I missed - content of links should be lowercase and you should style it to be uppercase using CSS.
So please add text-transform to `.nav__link {}`

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+}
+
+.is-active::after {
+  content: "";
+  display: block;
+  height: 4px;
+  width: 100%;
+  border-radius: 8px;
+  background-color: var(--main-color);
+  color: var(--main-color);
```
mvjl000, line 64: You can remove color from after. Only background color affects the color of this 'underline'.

mvjl000 (changes requested):

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+  align-items: center;
+  height: 60px;
+}
+
```
asakevych, line 20: please add .logo and give it size => width and height should be 40px
if your image will be bigger or smaller, your header will crush without those styles

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+.nav {
+  display: flex;
+  height: 100%;
+  cursor: pointer;
```
asakevych, line 24: it's unnecessary here
- when cursor is pointer, it suggets to user that he can click on the element, but navigation do nothing, it's not ckickable
- just links inside navigation are clickable and `cursor: pointer` is default value of `<a>` element

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+}
+
+.nav__ul {
+  list-style-type: none;
```
asakevych, line 28: it's better to use `list-style: none`, it will remove all list dedicatetd styles

asakevych (changes requested):
So far so good!! you are really close to perfect solution, just some small changes left 😉

mvjl000 (approved):
All good 👌

PR #3476, opened 2023-07-04
Develop

Noini111:
- [DEMO LINK](https://Noini111.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Noini111.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,52 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="https://fonts.google.com/specimen/Roboto">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <div class="navigation box">
+
+      <header class="nav-logo">
+        <a href="#home">
+          <img src="images/logo.png" alt="Logo_Moyo" >
+        </a>
+      </header>
+      <div class="box1"></div>
+      <nav class="box">
+
+
+
+        <ul class="nav-list">
+          <li>
+            <a href="#Apple" class="is-active">Apple</a>
+          </li>
+          <li>
+            <a href="#Samsung" class="is-active">Samsung</a>
+          </li>
+          <li>
+            <a href="#Smatphones" class="is-active">Smartphones</a>
+          </li>
+          <li>
+            <a
+              href="#Laptops&Computers"
+              class="is-active"
+              data-qa="hover">Laptops & Computers</a>
+          </li>
+          <li>
+            <a href="#Gadgets" class="is-active">Gadgets</a>
+          </li>
+          <li>
+            <a href="#Tablets" class="is-active">Tablets</a>
+          </li>
+          <li>
+            <a href="#Photo" class="is-active">Photo</a>
+          </li>
+          <li>
+            <a href="#Video" class="is-active">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </div>
+
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -10,41 +10,41 @@
     <link rel="stylesheet" href="https://fonts.google.com/specimen/Roboto">
   </head>
   <body>
-    <header class="navigation">
-      <div class="nav-logo">
+    <header class="header">
+      <div class="header__logo">
         <a href="#home">
           <img src="images/logo.png" alt="Logo_Moyo" >
         </a>
       </div>
 
-      <nav class="box">
-        <ul class="nav-list">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,49 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class="navigation">
+      <div class="nav-logo">
+        <a href="#home">
+          <img src="images/logo.png" alt="Logo_Moyo" >
+        </a>
+      </div>
+
+      <nav class="box">
```
asakevych, line 20: it's better to make nav as separate block, so your nav element can have .nav class and then your header could have .header class

src/index.html
```diff
@@ -7,8 +7,49 @@
...
+        <ul class="nav-list">
+          <li class="link2">
```
asakevych, line 22: every li element should have the same class- e.g .nav_item

src/index.html
```diff
@@ -7,8 +7,49 @@
...
+            <a href="#Apple" class="is-active link left">Apple</a>
```
asakevych, line 23: every a element should have the same class- e.g .nav_link

src/style.css
```diff
@@ -1,3 +1,76 @@
+/* #region Reset */
+
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
```
asakevych, line 7: it's better to set font-weight only for your text elements, for example links `<a>`

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  flex-direction: row;
+  align-items: center;
+  justify-content: space-between;
+  width: auto;
+  padding: 0 50px;
+}
+
+.nav-logo {
+  display: block;
+  margin-top: 5px;
```
asakevych, line 26: unnecessary margin here

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+}
+
+.nav-list {
+  list-style-type: none;
```
asakevych, line 30: it's not needed, just use `list-style: none`

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  margin-right: 10px;
+}
+
+.left {
+  margin-right: 10px;
+}
+
+.right {
+  margin-left: 10px;
+}
```
asakevych, line 59: please use pseudo classes here e.g
```suggestion
.link:not(:last-of-type){
  margin-left: 10px;
}
```

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+
+.is-active::after {
+  content: "";
+  display: block;
+  height: 4px;
+  width: 100%;
+  border-radius: 2px;
+  background: #00acdc;
+  position: absolute;
+  bottom: -1px;
```
asakevych, line 74: you should have bottom 0, something is wrong if you have to use -1px 😉

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  padding: 0;
+}
+
+.link {
+  color: black;
+  text-decoration: none;
+  display: block;
+  position: relative;
+  padding-top: 22.5px;
+  padding-bottom: 22.5px;
```
asakevych, line 45: it's bad practice to set paddings like 22.5px, better set height for link - 60px,  it would work properly if you  change display to flex and set align-items to center

src/index.html
```diff
@@ -7,8 +7,49 @@
...
+            <a href="#Gadgets" class="link center">Gadgets</a>
+          </li>
+          <li>
+            <a href="#Tablets" class="link center">Tablets</a>
+          </li>
+          <li>
+            <a href="#Photo" class="link center">Photo</a>
+          </li>
+          <li>
+            <a href="#Video" class="link right">Video</a>
```
asakevych, line 47: center, right, left classes are unnecessary

asakevych (changes requested):
Great job! just some small changes needed 😉

src/style.css
Noini111: Thank you, but in task says, that we shouldn't fix height, and also everything is on wrong position, I tryed,  but I prefer not to fight against this test, I prefer to move on next tasks, and there I will try remembering this rule. :)

src/style.css
Noini111: Hello, I have tryed to replace it with :firstchilde :lastchild, but otherwise it's not working... :C

Noini111:
I made adjustments, which I could and didn't break the test. :)

choeqq (approved):
Looks good to me 😄 
GJ 😉

mvjl000 (approved):

PR #3481, opened 2023-07-04
add task solution

Valentyn97:


src/style.css
```diff
@@ -1,3 +1,88 @@
...
+.logo__image {
+  height: 100%;
+  width: 100%;
+}
+
+.nav {
+  display: block;
+}
+
+li:first-child {
```
DarkMistyRoom, line 36: use classes instead of tags

src/style.css
```diff
@@ -1,3 +1,88 @@
...
+  font-size: 12px;
+  font-weight: 500;
+  white-space: nowrap;
+  text-decoration: none;
+  text-transform: uppercase;
+  color: #000;
+  transition: color 0.3s;
+}
+
+.nav__link.is-active {
```
DarkMistyRoom, line 65:
```suggestion
.is-active {
```

DarkMistyRoom: no need to increase specificity here

DarkMistyRoom (approved):
Good! Just check comments

PR #3485, opened 2023-07-05
add task solution

PawelP2001:


PawelP2001:
[DEMO LINK](https://PawelP2001.github.io/layout_moyo-header)
[TEST REPORT LINK](https://PawelP2001.github.io/layout_moyo-header/report/html_report/)

src/style.css
```diff
@@ -1,3 +1,63 @@
...
+}
+
+.is-active {
+  color: black;
+  text-decoration: none;
+  display: block;
+  padding: 10px;
+  flex-shrink: 0;
+}
+
```
mvjl000, line 48: You should also add the blue line under the first link - apple.
After applying changes related to class naming I mentioned before try to do something linke this:
create ".nav__link--active" selector in which you are going to set the color to blue and position relative (we need this position in order to place the blue line correctly)

Below this selector block create an after pseudo element like this: `.nav__link--active::after` - this is going to be the blue line. Inside set all necessary properties like content: ""; width: 100% etc. and remember about position: absolute; 

;)

mvjl000 (changes requested):

src/style.css
```diff
@@ -1,3 +1,73 @@
...
+.navigation {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
```
choeqq, line 20: those selectors are both targeting the same element - header with class `navigation`, you can remove element selector 😉

src/index.html
```diff
@@ -7,8 +7,44 @@
...
+            <a href="#gadgets" class="nav__link">gadgets</a>
+          </li>
+          <li>
+            <a href="#tablets" class="nav__link">tablets</a>
+          </li>
+          <li>
+            <a href="#photo" class="nav__link">photo</a>
+          </li>
+          <li>
+            <a href="#video" class="nav__link nav__link--last">video</a>
```
choeqq, line 44: I can't really see the usege of that second class - please remove it, if not necessary 😉

choeqq (approved):
All test are passing, so I approved your changes 🚀 

Please keep in my our comments 😉

mvjl000 (approved):

PR #3488, opened 2023-07-06
moyo header

AndrzejPec:
- [DEMO LINK](https://AndrzejPec.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://AndrzejPec.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,12 +3,59 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+          content="width=device-width, user-scalable=no, initial-scale=1.0,
+          maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    >
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="MoyoHeader">
+      <div class="MoyoHeader__logo">
+        <a href="#home" class="MoyoLogo__Link">
+          <img
+            src="./images/logo.png"
+            alt="Moyo Logo"
+            class="MoyoLogo__Image"
+          >
+        </a>
+      </div>
+      <nav class="MoyoHeader__nav">
+        <nav class="MoyoHeader__nav__links">
+          <ul>
+            <li class="MoyoHeader__nav__elem">
+              <a class=" is-active">Apple</a>
+            </li>
+            <li class="MoyoHeader__nav__elem">
+              <a>Samsung</a>
+            </li>
+            <li class="MoyoHeader__nav__elem">
+              <a>Smartphones</a>
+            </li>
+            <li class="MoyoHeader__nav__elem">
+              <a data-qa="hover">Laptops & Computers</a>
+            </li>
+            <li class="MoyoHeader__nav__elem">
+              <a>Gadgets</a>
+            </li>
+            <li class="MoyoHeader__nav__elem">
+              <a>Tablets</a>
+            </li>
+            <li class="MoyoHeader__nav__elem">
+              <a>Photo</a>
+            </li>
+            <li class="MoyoHeader__nav__elem">
+              <a>Video</a>
+            </li>
+          </ul>
+        </nav>
+      </nav>
+    </header>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,124 @@
+/* Stylowanie wszystkich divów */
+/* div,
+nav {
+  border: 2px solid;
+} */
+
```
choeqq, line 6: please remove unnecessary comments

src/style.css
```diff
@@ -1,3 +1,124 @@
...
 body {
+  font-family: "Roboto", sans-serif;
   margin: 0;
+  padding: 0;
+}
+
+ul {
+  margin: 0;
+  padding: 0;
+}
```
choeqq, line 16: uss class selectors instead of elements please

src/index.html
```diff
@@ -3,12 +3,59 @@
...
+          <img
+            src="./images/logo.png"
+            alt="Moyo Logo"
+            class="MoyoLogo__Image"
+          >
+        </a>
+      </div>
+      <nav class="MoyoHeader__nav">
+        <nav class="MoyoHeader__nav__links">
+          <ul>
```
choeqq, line 31: please provide class on `<ul>` element

src/index.html
```diff
@@ -3,12 +3,59 @@
...
+        <a href="#home" class="MoyoLogo__Link">
+          <img
+            src="./images/logo.png"
+            alt="Moyo Logo"
+            class="MoyoLogo__Image"
+          >
+        </a>
+      </div>
+      <nav class="MoyoHeader__nav">
+        <nav class="MoyoHeader__nav__links">
```
choeqq, line 30: we should use only 1 nav element

src/style.css
```diff
@@ -1,3 +1,124 @@
...
+  align-items: center;
+}
+
+.MoyoHeader__nav__links a {
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: normal;
+  text-transform: uppercase;
+}
```
choeqq, line 33: it would be better to add another class on `<a>` tag, instead of nesting selectors

src/style.css
```diff
@@ -1,3 +1,124 @@
...
+
+.MoyoHeader__nav__elem a:hover {
+  text-decoration: none;
+  color: #00acdc;
+}
+
+.MoyoHeader__nav__elem img {
+  width: 20px;
+  height: 20px;
+}
```
choeqq, line 110: you can also add class on ``, it will be easier to target

src/style.css
```diff
@@ -1,3 +1,124 @@
...
+
+.MoyoHeader__nav__elem:last-child {
+  margin-right: 0;
+}
+
+[data-qa="hover"] {
+  cursor: pointer;
+  text-decoration: none;
+}
```
choeqq, line 119: you should add hover class on every element not only on one with `data-qa`, you did it above

src/style.css
```diff
@@ -1,3 +1,124 @@
...
+
+a.is-active {
```
choeqq, line 121: you can leave `is-active` class without a element

choeqq (changes requested):
small adjustments needed 👍

src/style.css
```diff
@@ -1,3 +1,98 @@
...
+}
+
+.MoyoHeader__nav__list {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  display: flex;
+  height: 100%;
+  flex-grow: 1;
+  align-items: stretch;
```
choeqq, line 14: I don't think this property is necessary

src/style.css
```diff
@@ -1,3 +1,98 @@
...
+  width: 40px;
+  height: 40px;
+  vertical-align: bottom;
+}
+
+.MoyoHeader__nav__elem:last-child {
+  margin-right: 0;
+}
+
+[data-qa="hover"]:hover,
```
choeqq, line 92: please remove this selector - hover is already set below

src/style.css
```diff
@@ -1,3 +1,98 @@
...
+a:hover,
```
choeqq, line 93: please change `a` selector to `nav__link` class

src/style.css
```diff
@@ -1,3 +1,98 @@
...
+  left: 0;
+  width: 100%;
+  height: 4px;
+  background-color: #00acdc;
+}
+
+.MoyoLogo__Image {
+  width: 40px;
+  height: 40px;
+  vertical-align: bottom;
```
choeqq, line 85: I don't think this property is necessary

src/style.css
```diff
@@ -1,3 +1,98 @@
...
+  align-content: center;
+  display: flex;
+  flex-wrap: wrap;
+  text-decoration: none;
+}
+
+.MoyoHeader__logo {
+  margin-left: 50px;
+  border-color: blue;
+  font-size: 0;
```
choeqq, line 44: There is no text - you don't have to change `font-size`

choeqq (changes requested):
All tests are passing, but small improvements are needed 😄

src/style.css
AndrzejPec: unfortunately it is necessary because of some strange img property - it was described in the introductory video

choeqq (approved):
All good, well done 👍

PR #3521, opened 2023-07-12
Created header  according to figma design

davinnchii:


davinnchii:
Локально тести проходять 
Screenshot 2023-07-17 at 16 40 26

src/css/style.css
```diff
@@ -0,0 +1,64 @@
...
+  justify-content: space-between;
+  align-items: center;
+  padding-right: 50px;
+  padding-left: 50px;
+  max-width: 1200px;
+}
+
+.nav {
+  min-width: 635.54px;
+}
```
BudnikOleksii, line 22: You don't need to hardcode width for the current design.

src/css/style.css
```diff
@@ -0,0 +1,64 @@
...
+  font-style: normal;
+  line-height: 100%;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding-right: 50px;
+  padding-left: 50px;
```
BudnikOleksii, line 16: It's better to use the shorthand property "padding".

BudnikOleksii (changes requested):
Good start👍🏻
But there is still some work that needs to be done.

src/index.html
```diff
@@ -6,9 +6,82 @@
...
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
+        <a href="#home">
+          <img
+          src="images/logo.png"
+          alt="logo"
+          width="40px"
+          height="40px"
+          >
```
BudnikOleksii, line 21:
```suggestion
          <img
            src="images/logo.png"
            alt="logo"
            class="logo__image"
          >
```
- Don't use inline styles, use classes for styling instead.
- Pay attention to the checklist: https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md
[CODE STYLE] - If the HTML-element has long attribute values or number of attributes is more than 2 - start each one, including the first, on the new line with 2-space indentation related to tag. Tag’s closing bracket should be on the same level as opening one.

BudnikOleksii (changes requested):
Let's fix code style a bit and get rid of inline styles

dffUqp (approved):
LGTM:fire:

PR #3580, opened 2023-07-18
add task solution

SmartSergiZ:
- [DEMO LINK](https://SmartSergiZ.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://SmartSergiZ.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,8 +1,12 @@
+:root {
+  --black: #000;
+  --active: #00acdc;
+}
+
 nav,
 body {
   margin: 0;
   padding: 0;
-  box-sizing: border-box;
   font-family: Roboto, sans-serif;
 }
 
@@ -36,14 +40,14 @@ body {
   display: flex;
   height: 60px;
   text-decoration: none;
-  color: #000;
+  color: var(--black);
   font-size: 12px;
   font-weight: 500;
   align-items: center;
 }
 
 .is-active {
-  color: #00acdc;
+  color: var(--active);
 }
 
 .logo__img {
@@ -52,16 +56,16 @@ body {
 }
 
 .nav__link:hover {
-  color: #00acdc;
+  color: var(--active);
 }
 
 .nav__link.is-active::after {
   display: block;
   height: 4px;
   width: 100%;
-  background-color: #00acdc;
+  background-color: var(--active);
   position: absolute;
   bottom: 0;
-  border-radius: 8px;
+  border-radius: 4px;
   content: "";
 }
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,7 @@
+* {
+  box-sizing: border-box;
+}
+
 :root {
   --text-color: #000;
   --link-color: #00acdc;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,67 @@
...
 body {
   margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: Roboto, sans-serif;
+}
+
+.header {
+  margin: 0;
+  box-sizing: border-box;
```
alexander-ignatow, line 11: Setting `box-sizing` individually for each class is not a great idea - as there are few scenarios when one needs to vary it.

In most cases, including yours, it is better to set a single box-sizing value for all elements and vary it only when needed

Try to fix this issue - it looks minor in this task but may become painful in large projects.

Here is a good article on box-sizing: https://css-tricks.com/box-sizing/

And here is a good way to set box-sizing: https://css-tricks.com/box-sizing/#aa-good-better-and-probably-best-box-sizing-reset-methods

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+  color: #00acdc;
+}
+
+.logo__img {
+  display: flex;
+  align-items: center;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
alexander-ignatow, line 55: It is better to set colors via CSS variables - to be able to control them.  This even more important if the color is reused.
Check this: https://www.w3schools.com/css/css3_variables.asp

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+}
+
+.nav__link.is-active::after {
+  display: block;
+  height: 4px;
+  width: 100%;
+  background-color: #00acdc;
+  position: absolute;
+  bottom: 0;
+  border-radius: 8px;
```
alexander-ignatow, line 65: The border-radius is bigger than the height - an obvious typo here :)
It does not have any effect in this case, still please fix it.

alexander-ignatow (changes requested):
Good solution! Visually everything is ok, but let's improve the code quality a bit.

SmartSergiZ:
tnx)
all changes have already been added

roman-mirzoian (approved):
Well done 👍

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+  --text-color: #000;
+  --link-color: #00acdc;
+}
+
+nav,
 body {
   margin: 0;
+  padding: 0;
+  font-family: Roboto, sans-serif;
+  box-sizing: border-box;
```
alexander-ignatow, line 11: box-sizing is not inherited - you should set it globally via `*`

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+}
+
+.header {
+  margin: 0;
```
alexander-ignatow, line 15: No margins on BEM blocks

alexander-ignatow (changes requested):
Very good - some things are fixed, still some remain

Also pleas make sure to deploy your latest solution ☝️

BudnikOleksii (approved):
Good job👍

PR #3588, opened 2023-07-19
my task solution

cherniavskaaa:


src/style.css
```diff
@@ -1,3 +1,81 @@
...
+}
+
+.container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 1200px;
+  height: 60px;
+  background-color: hsl(0, 0%, 100%);
+  /* padding: 0 50px; */
```
alexander-ignatow, line 20: No need to keep these old, commented out code (here and below)

src/style.css
```diff
@@ -1,3 +1,81 @@
...
+}
+
+.nav__link {
+  text-decoration: none;
+  padding: 23px 0;
+  color: hsl(0, 0%, 0%);
+}
+
+.is-active {
+  color: hsl(193, 100%, 43%);
```
alexander-ignatow, line 63: Here and below - use CSS variables
https://www.w3schools.com/css/css3_variables.asp

src/style.css
```diff
@@ -1,3 +1,81 @@
...
+  position: relative;
+
+}
+
+.is-active::after {
+  display: block;
+  content: "";
+  width: 37px;
```
alexander-ignatow, line 71: Do not hardcode width. Try setting `is-active` style to other links and you will see why.

instead set the width to `100%` so it will fit any element.

src/style.css
```diff
@@ -1,3 +1,81 @@
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  box-sizing: border-box;
+  padding: 0 50px;
```
alexander-ignatow, line 5: Don't set paddings to `body` - it is better to style some internal containers instead (in your case it would be `header`

src/style.css
```diff
@@ -1,3 +1,81 @@
...
+  display: flex;
+  justify-content: center;
+}
+
+.container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 1200px;
+  height: 60px;
```
alexander-ignatow, line 18: per checklist: `Don't set height for header explicitly. Let the content (links) dictate it.`

src/style.css
```diff
@@ -1,3 +1,81 @@
...
+.header {
+  display: flex;
+  justify-content: center;
+}
+
+.container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 1200px;
```
alexander-ignatow, line 17: Try to avoid setting width explicitly for top-level containers as so.

src/index.html
```diff
@@ -6,9 +6,67 @@
...
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
```
alexander-ignatow, line 16: Most probably you can remove `<div class="container" ...` and move it's styles to the header

That will make your life easier when styling the header

alexander-ignatow (changes requested):
Strong attempt 💪 

Now let's make this code better!

src/index.html
cherniavskaaa: i tried to fix it but it didn't work this way

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+.nav__link {
+  text-decoration: none;
+  padding: 23px 0;
+  color: var(--black);
+}
+
+.is-active {
+  color: var(--blue);
+  position: relative;
+
```
dffUqp, line 66:
```suggestion
```

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  justify-content: center;
+  padding: 0 50px;
+}
+
+.container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+  background-color: hsl(0, 0%, 100%);
```
dffUqp, line 23: let's declare this color inside `:root`

dffUqp (changes requested):
LGTM:fire: but needs a few minor fixes:smiley:

cherniavskaaa:
I guess I fixed all changes.

src/style.css
```diff
@@ -1,3 +1,82 @@
+:root {
+  --blue: hsl(193, 100%, 43%);
+  --black: hsl(0, 0%, 0%);
+  --white: hsl(0, 0%, 100%);
```
roman-mirzoian, line 4: When naming variables, you should use the name according to the purpose, and not according to the color, because the color may change to another during the development process. Then you have to update it throughout the code, not just where this variable is created.

src/index.html
roman-mirzoian: In addition to repositioning the classes and removing the redundant tag, you also need to update the `css` styles to make everything work properly.

roman-mirzoian (changes requested):
Good, but needs some fixing.

src/index.html
cherniavskaaa: I've fixed it now, i had to delete width in the header to make it works, thanks.

src/style.css
```diff
@@ -1,3 +1,86 @@
...
+  background-color: var(--background);
+}
+
+/* .container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+  background-color: var(--background);
+} */
```
tiserett, line 29: get rid of all the comments

tiserett (approved):
Well done, but make sure not to leave any comments in your PR

src/style.css
cherniavskaaa: thank you, I did it)

PR #3593, opened 2023-07-19
Added header layout

Aserg1992:
- [DEMO LINK](https://Aserg1992.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Aserg1992.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -16,71 +16,55 @@
 
   <body>
     <header class="header">
-      <div class="container">
-        <div class="header_wraper">
-          <div class="header__logo">
-            <a href="/">
-              <img src="images/logo.png" alt="MOYO logo"
-              width="40" height="40">
+      <a href="/" class="header__logo">
+        <img src="images/logo.png" alt="MOYO logo"
+        width="40px" height="40px">
+      </a>
+      <nav class="nav">
+        <ul class="nav-list">
+          <li class="nav-list__item">
+            <a href="#apple" class="nav-list__link is-active">
+              Apple
             </a>
-
-          </div>
-          <div class="header__nav">
-            <nav class="nav">
-              <ul class="nav-list">
-                <li class="nav-list__item">
-                  <a href="#apple" class="nav-list__link is-active">
-                    Apple
-                  </a>
-                </li>
-
-                <li class="nav-list__item">
-                  <a href="#samsung" class="nav-list__link">
-                    Samsung
-                  </a>
-                </li>
-
-                <li class="nav-list__item">
-                  <a href="#smartphones" class="nav-list__link">
-                    Smartphones
-                  </a>
-                </li>
-
-                <li class="nav-list__item">
-                  <a href="#laptops" class="nav-list__link"
-                    data-qa="hover">
-                    Laptops & Computers
-                  </a>
-                </li>
-
-                <li class="nav-list__item">
-                  <a href="#gadgets" class="nav-list__link">
-                    Gadgets
-                  </a>
-                </li>
-
-                <li class="nav-list__item">
-                  <a href="#tablets" class="nav-list__link">
-                    Tablets
-                  </a>
-                </li>
-
-                <li class="nav-list__item">
-                  <a href="#photo" class="nav-list__link">
-                    Photo
-                  </a>
-                </li>
-
-                <li class="nav-list__item">
-                  <a href="#video" class="nav-list__link">
-                    Video
-                  </a>
-                </li>
-              </ul>
-            </nav>
-          </div>
-        </div>
-      </div>
+          </li>
+          <li class="nav-list__item">
+            <a href="#samsung" class="nav-list__link">
+              Samsung
+            </a>
+          </li>
+          <li class="nav-list__item">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,95 @@
...
+:root {
+  --select-color: #00acdc;
+  --main-color: #000;
+  --font-family: roboto, sans-serif;
+  --main-text-size: 16px;
+}
+
+html,
+body {
+  height: 100%;
```
roman-mirzoian, line 12: Do not set unnecessary styles, use them only when necessary. You have a lot of extra `height: 100%;` now.

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+}
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
```
roman-mirzoian, line 17: You created the `--font-family` variable, so you should use it here.

src/index.html
```diff
@@ -1,14 +1,70 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="/" class="header__logo">
+        <img src="images/logo.png" alt="MOYO logo"
+        width="40px" height="40px">
```
roman-mirzoian, line 21: It is better not to set inline styles, but to do styling through `css` files.

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  --font-family: roboto, sans-serif;
+  --main-text-size: 16px;
+}
+
+html,
+body {
+  height: 100%;
+}
+
 body {
```
roman-mirzoian, line 15: Try not to refer to tags directly, but do it through classes.

roman-mirzoian (changes requested):
The task is completed and I see that all the tests have passed, but be careful when writing the code: do not use unnecessary styles where they are not needed, if you have created a variable - use it for its intended purpose.
The code itself needs to be reworked a bit to make it correct.

src/style.css
Aserg1992: Problem is fixed

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+}
+
+.nav-list__item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav-list__link.is-active {
+  position: relative;
+  color: var(--select-color);
+}
```
alexander-ignatow, line 64: A better idea is not to alter `position` based on the modifier

a better idea would be moving `position: relative` to `nav-list__link`

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  left: 0;
+  height: 4px;
+  width: 100%;
+  border-radius: 8px;
+  background-color: var(--select-color);
+}
+
+.header {
+  align-items: center;
+  height: 60px;
```
alexander-ignatow, line 79: Per checklist:
`Don't set height for header explicitly. Let the content (links) dictate it.`

Hint: if you will set height (explicitly or by padding) for one of the inner elements, you will not need those ugly `height: 100%` in so many places

alexander-ignatow (changes requested):
Good 👍 but still some issues to fix

BTW - it is a good habit to answer review comments - that helps you to track them and it is polite.
You are also free to use emotions to react to them 👍

src/index.html
```diff
@@ -1,14 +1,70 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Moyo header</h1>
+
+  <body class="body">
+    <header class="header">
+      <a href="/" class="header__logo">
+        <img src="images/logo.png" alt="MOYO logo"
+        class="header__logo__image">
```
tiserett, line 21: fix code style

src/index.html
```diff
@@ -1,14 +1,70 @@
...
+            </a>
+          </li>
+          <li class="nav-list__item">
+            <a href="#smartphones" class="nav-list__link">
+              Smartphones
+            </a>
+          </li>
+          <li class="nav-list__item">
+            <a href="#laptops" class="nav-list__link"
+              data-qa="hover">
```
tiserett, line 42: code style

tiserett (changes requested):
good job, but make sure to fix all the coments ;)

src/index.html
Aserg1992: image

What exactly am I supposed to fix?

roman-mirzoian: Don't forget to check the `checklist`, there are lots of useful tips and examples of good code style.
image

roman-mirzoian (changes requested):
There is quite a bit left and it will be perfect 👍

src/index.html
Aserg1992: Thank you! Done!

alexander-ignatow (approved):
LGTM! 👍

PR #3597, opened 2023-07-20
add task solution

MarekSax:
https://mareksax.github.io/layout_moyo-header/
https://mareksax.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,73 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
+  background: #fff;
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,77 @@
...
   margin: 0;
+  font-family: Roboto, sans-serif;
+  color: #000;
+}
+
+div,
+li,
+p,
+ul {
+  box-sizing: border-box;
```
darokrk, line 15: stil style property should be use for all elements in the project so should be used like this 
```suggestion
* {
    box-sizing: border-box;
  }
```

darokrk (approved):
Good job 👍🏽  left one mention to fix, please do that before move next 😃

darokrk (approved):
Good job 👍🏽

PR #3599, opened 2023-07-21
add task solution

dshukalovych:
- [DEMO LINK](https://dshukalovych.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://dshukalovych.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,10 +5,45 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,300;0,400;1,300&display=swap" rel="stylesheet">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img src="images/logo.png" alt="MAYO logo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="link is-active apple">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="link ">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="link ">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Laptops&Computers"
+             class="link is-active"
+             data-qa="hover">Laptops & Computers</a>
+          </li>
+          <li class="nav__item">
+            <a href="#gadgets" class="link">Gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#tablets" class="link">Tablets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#photo" class="link ">Photo</a>
+          </li>
+          <li class="nav__item">
+            <a href="#video" class="link ">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,52 @@
 body {
   margin: 0;
+  display: block;
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+.link {
+  text-decoration: none;
+  color: black;
+  cursor: pointer;
+  padding: 20px 0 ;
+}
+
+.nav__list {
+  list-style: none;
+  display: flex;
+  padding: 0;
+  margin: 0;
+}
+
+.logo {
+  margin-top: 10px;
+  height: 40px;
+  width: 40px;
+}
+
+.nav__item {
+  margin-left: 20px;
+  font-family: Roboto, Arial, sans-serif;
+  font-size: 12px;
+  text-transform: uppercase;
+  display: flex;
+}
+
+.is-active:hover {
+  color: #00acdc;
+}
+
+.apple {
+  text-decoration: underline;
+  text-decoration-color: #00acdc;
+  text-underline-offset: 18.5px;
+  text-decoration-thickness: 4px;
+  color: #00acdc;
 }
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Moyo header
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_moyo-header/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_moyo-header/report/html_report/)
+- [DEMO LINK](https://dshukalovych.github.io/layout_moyo-header/)
+- [TEST REPORT LINK](https://dshukalovych.github.io/layout_moyo-header/report/html_report/)
 
 > Follow [these instructions](https://mate-academy.github.io/layout_task-guideline/#how-to-solve-the-layout-tasks-on-github)
 


src/index.html
```diff
@@ -5,10 +5,45 @@
...
+          <li class="nav__item">
+            <a href="#samsung" class="link ">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="link ">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a href="#Laptops&Computers"
+             class="link is-active"
+             data-qa="hover">Laptops & Computers</a>
```
Esceype, line 31: Checklist point number 13

src/style.css
```diff
@@ -1,3 +1,52 @@
 body {
   margin: 0;
+  display: block;
```
Esceype, line 3: Remove it here here you need just reset browser default margins

src/style.css
```diff
@@ -1,3 +1,52 @@
...
+  margin-left: 20px;
+  font-family: Roboto, Arial, sans-serif;
+  font-size: 12px;
+  text-transform: uppercase;
+  display: flex;
+}
+
+.is-active:hover {
+  color: #00acdc;
+}
```
Esceype, line 44: hover should be at all links, not only for this selector

src/style.css
```diff
@@ -1,3 +1,52 @@
...
+
+.apple {
+  text-decoration: underline;
+  text-decoration-color: #00acdc;
+  text-underline-offset: 18.5px;
+  text-decoration-thickness: 4px;
+  color: #00acdc;
 }
```
Esceype, line 52: First of all that bad name for class.
Second it should be adaptive and responsive, so if you use this class for another link it should be 100% width at all situations and have position at end of block

src/style.css
```diff
@@ -1,3 +1,52 @@
...
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+.link {
+  text-decoration: none;
+  color: black;
+  cursor: pointer;
+  padding: 20px 0 ;
```
Esceype, line 18: You don`t need it here

src/style.css
```diff
@@ -1,3 +1,52 @@
...
+}
+
+.logo {
+  margin-top: 10px;
+  height: 40px;
+  width: 40px;
+}
+
+.nav__item {
+  margin-left: 20px;
```
Esceype, line 35: Are you sure that the first child needs this margin ?

src/index.html
```diff
@@ -5,10 +5,45 @@
...
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="link is-active apple">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="link ">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="link ">Smartphones</a>
```
Esceype, line 26: The name of this class does not meet the requirements of BEM

src/index.html
```diff
@@ -5,10 +5,45 @@
...
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,300;0,400;1,300&display=swap" rel="stylesheet">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
```
Esceype, line 14: If this is separate block and not an element of header BEM block then you can`t you margin here, think how to fix it

src/style.css
```diff
@@ -1,3 +1,52 @@
...
+
+.header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+.link {
```
Esceype, line 14: What is the line height of this block ?

src/style.css
```diff
@@ -1,3 +1,52 @@
 body {
   margin: 0;
+  display: block;
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
```
Esceype, line 8: You don`t need you set flex-direction: row because this is default direction of flex box

src/style.css
```diff
@@ -1,3 +1,52 @@
...
+
+.logo {
+  margin-top: 10px;
+  height: 40px;
+  width: 40px;
+}
+
+.nav__item {
+  margin-left: 20px;
+  font-family: Roboto, Arial, sans-serif;
```
Esceype, line 36: Add one more font as fallback font and move this css prop(font-family) to the very top of your block tree.
The font is always defined at the topmost classes, usually in body or html

Esceype (changes requested):
Adding to all the comments below.
Also, the problem is that your entire header is not the height it should be, on the mockup the header occupies exactly 60 pixels, think about how to do it without hardcoding, but allowing the block to expand it by itself.
I believe that you will quickly deal with these comments and we will be able to continue

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+  padding: 0 50px;
+  /* max-height: 60px; */
+  margin: auto;
+}
+
+.nav__link {
+  text-decoration: none;
+  color: black;
+  cursor: pointer;
+  padding: 30px 0 ;
```
Esceype, line 19: Wrong paddings, so this is the reason why your tests do not pass. Take padding from mockup and and round them up

Esceype (changes requested):
GJ, but still one more thing

src/index.html
```diff
@@ -5,10 +5,50 @@
...
+        </a>
+      </div>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link ">Samsung</a>
+          </li>
```
denys-danyliuk, line 27:
```suggestion
        <ul class="nav__list">
          <li class="nav__item">
            <a href="#apple" class="nav__link is-active">Apple</a>
          </li>
          
          <li class="nav__item">
            <a href="#samsung" class="nav__link ">Samsung</a>
          </li>
```
add spaces between sibling multiline blocks

src/index.html
```diff
@@ -5,10 +5,50 @@
...
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="nav__logo">
+        <a href="#home">
+          <img src="images/logo.png" alt="MAYO logo">
```
denys-danyliuk, line 17:
```suggestion
          <img src="images/logo.png" alt="MOYO logo">
```

src/index.html
```diff
@@ -5,10 +5,50 @@
...
+            <a href="#samsung" class="nav__link ">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link ">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+             href="#Laptops&Computers"
+             class="nav__link"
+             data-qa="hover">Laptops & Computers
```
denys-danyliuk, line 35: 13'th point in checklist

denys-danyliuk (changes requested):

src/index.html
```diff
@@ -5,10 +5,57 @@
...
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link ">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#Laptops&Computers"
+              class="nav__link"
+              data-qa="hover">Laptops & Computers
+            </a>
```
denys-danyliuk, line 39: 13'th point in checklist

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+  align-items: center;
+  padding: 0 50px;
+  margin: auto;
+}
+
+.nav__link {
+  position: relative;
+  display: flex;
+  align-items: center;
+  height: 60px;
```
denys-danyliuk, line 19: don't hardcode exact height, use paddings

denys-danyliuk (changes requested):

xiimiik (approved):
Fine work!

ps dont forget about my comment!

PR #3602, opened 2023-07-22
Develop

Exess94:


src/style.css
```diff
@@ -1,3 +1,65 @@
...
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
+}
+
+.is-active.nav__link::after {
```
volodymyr-soltys97, line 51:
```suggestion
.is-active::after {
```

src/style.css
```diff
@@ -1,3 +1,65 @@
...
+  text-decoration: none;
+  text-transform: uppercase;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.is-active {
+  color: #00acdc;
```
volodymyr-soltys97, line 48: Create a variable for this color and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

VitaliyBondarenko1982 (approved):

PR #3625, opened 2023-07-26
the moyo header has been created

noraaudonnet:
https://noraaudonnet.github.io/layout_moyo-header/
https://noraaudonnet.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,55 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
+        <a href="#home">
+          <img
+            src="./images/logo.png"
+            alt="Moyo logo"
+            class="logo"
+          >
+        </a>
+        <nav class="nav">
+          <ul class="nav__list">
+            <li class="nav__item">
+              <a href="#apple" class="nav__link is-active">Apple</a>
+            </li>
+            <li class="nav__item">
+              <a href="#samsung" class="nav__link">Samsung</a>
+            </li>
+            <li class="nav__item">
+              <a href="#smartphones" class="nav__link">Smartphones</a>
+            </li>
+            <li class="nav__item">
+              <a
+                href="#laptops&computers"
+                class="nav__link"
+                data-qa="hover"
+              >
+              Laptops & Computers
+              </a>
+            </li>
+            <li class="nav__item">
+              <a href="#gadgets" class="nav__link">Gadgets</a>
+            </li>
+            <li class="nav__item">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,55 @@
...
+    <header class="header">
+      <div class="container">
+        <a href="#home">
+          <img
+            src="./images/logo.png"
+            alt="Moyo logo"
+            class="logo"
+          >
+        </a>
+        <nav class="nav">
```
Anastasiia-Svintsova, line 24: Don't forget to add empty lines between multiline sibling blocks of HTML. Fix it everywhere 
```suggestion
        </a>
        
        <nav class="nav">
```

src/style.css
```diff
@@ -1,3 +1,59 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
+  max-width: 100%;
```
Anastasiia-Svintsova, line 9: max-width seems redundant here

src/style.css
```diff
@@ -1,3 +1,59 @@
...
+  color: #000;
+  font-size: 12px;
+  height: 60px;
+  line-height: 60px;
+  text-transform: uppercase;
+  text-decoration: none;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
Anastasiia-Svintsova, line 42: Use variables for repeatable colors

Anastasiia-Svintsova (changes requested):
Good start! Let's add a few fixes 

- The menu shouldn't shrink and shrink the logo
image

src/index.html
noraaudonnet: thanks! 
but I don't really understand how to resolve the issue with the shrinking menu and logo. I've tried to see it as you did, but for me it has been shown non-shrinked logo, so a normal one. at the same time I did see a problem with the second line appear in the menu and don't know how to fix it :( I've tried flex-wrap/flex-shrink and it still doesn't work. so far I've added flex-shrink: 0 property to the logo but can't see any difference. I've also tried to add flex-wrap: now-rap to the container with the menu -> it didn't work. then I tried to add the same property to the nav__list -> it didn't work. the same situation with flex-shrink there. so, what should I do at this point then? I'm also a bit confused with empty lines between multiline sibling block, which means I'm not sure what the multiline sibling blocks are. I added empty lines between <li> in the navigation, but not sure if you meant exactly those ones. and yes, added that empty line between logo and nav

src/index.html
```diff
@@ -7,8 +7,63 @@
...
+            </li>
+
+            <li class="nav__item">
+              <a
+                href="#laptops&computers"
+                class="nav__link"
+                data-qa="hover"
+              >
+              Laptops & Computers
+              </a>
```
witflash, line 46:
```suggestion
              <a
                href="#laptops&computers"
                class="nav__link"
                data-qa="hover"
              >
                Laptops & Computers
              </a>
```

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+}
+
+.container {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+nav {
```
witflash, line 23: Use class selector instead of tag selector:
```
.nav {
  ...
```

witflash (approved):
Good job!
Pay attention to my comments for your future work.

PR #3680, opened 2023-08-09
MOYO

DudeMood:
https://DudeMood.github.io/layout_moyo-header/
https://DudeMood.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,79 @@
 body {
   margin: 0;
+  max-width: 1200px;
+  min-width: 1024px;
+}
+
+.menu {
+  display: flex;
+  align-items: center;
+  padding: 0 50px;
+  flex-direction: row;
+  justify-content: space-between;
+  height: 60px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,79 @@
 body {
   margin: 0;
+  max-width: 1200px;
```
polosanya, line 3: It would be great not to limit the max width, so the header looks okay on large screens too
image

src/style.css
```diff
@@ -1,3 +1,79 @@
...
+  justify-content: space-between;
+  height: 60px;
+}
+
+.img {
+  display: absolute;
+  width: 40px;
+  height: 40px;
+  left: 50px;
+  top: 10px;
```
polosanya, line 21: Pay attention to the CSS properties order

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #3688, opened 2023-08-11
add task solution

Patrykmclaren14:
https://Patrykmclaren14.github.io/layout_moyo-header/
https://Patrykmclaren14.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,21 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img src="./images/logo.png" alt="logo"
+      class="photo">
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item"><a href="#apple" class="is-active">apple</a></li>
+          <li class="nav__item"><a href="#samsung">samsung</a></li>
+          <li class="nav__item"><a href="#smartphones">smartphones</a></li>
+          <li class="nav__item"><a href="#Laptops & Computers" data-qa="hover">Laptops & Computers</a></li>
+          <li class="nav__item"><a href="#Gadgets">Gadgets</a></li>
+          <li class="nav__item"><a href="#Tablets">Tablets</a></li>
+          <li class="nav__item"><a href="#Photo">Photo</a></li>
+          <li class="nav__item"><a href="#Video">Video</a></li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,64 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap");
+
 body {
   margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+  background-color: black;
+  display: flex;
+  justify-content: center;
+}
+
+.header {
+  width: 1200px;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  background-color: white;
+  padding: 0 50px;
+}
+
+.photo {
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
+  list-style: none;
+  display: flex;
+  align-items: center;
+  padding: 0;
+}
+
+.nav__item {
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: normal;
+  text-transform: uppercase;
+  margin-left: 20px;
+  padding: 7px 0;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
+
+a {
+  text-decoration: none;
+  color: #000;
+  font-size: 12px;
+  box-sizing: border-box;
+  padding: 23px 0;
+}
+
+a.is-active {
+  color: #00acdc;
+  border-bottom: 4px solid #00acdc;
+  border-radius: 2px;
+  padding-bottom: 19px;
+}
+
+[data-qa="hover"]:hover {
+  color: #00acdc;
 }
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -9,6 +9,21 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img src="./images/logo.png" alt="logo"
+      class="photo">
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item"><a href="#apple" class="is-active">apple</a></li>
```
VitaliyBodnar, line 17: please add the spaces between the `<li>` blocks, as it will make a code cleaner

src/style.css
```diff
@@ -1,3 +1,64 @@
...
+
+a.is-active {
+  color: #00acdc;
+  border-bottom: 4px solid #00acdc;
+  border-radius: 2px;
+  padding-bottom: 19px;
+}
+
+[data-qa="hover"]:hover {
+  color: #00acdc;
```
VitaliyBodnar, line 63: #00acdc color is duplicated, so please move all duplicated styles to the variables

VitaliyBodnar (changes requested):

src/style.css
```diff
@@ -1,3 +1,64 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap");
+
 body {
   margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+  background-color: black;
+  display: flex;
+  justify-content: center;
```
mvjl000, line 9: Flex here is unnecessary, we can leave the default display.

src/style.css
```diff
@@ -1,3 +1,64 @@
...
+}
+
+.header {
+  width: 1200px;
```
mvjl000, line 13: My previous comment is related to this one. Removing flex from the body will make setting width unnecessary.
This approach with fixed width value is not ideal for [RWD](https://pl.wikipedia.org/wiki/Responsive_web_design)

src/style.css
```diff
@@ -1,3 +1,64 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  background-color: white;
+  padding: 0 50px;
+}
+
+.photo {
+  width: 40px;
+  height: 40px;
```
mvjl000, line 23: Checklist - 4th point:

I'd love to see a `logo dimensions` CSS variable for that.

src/style.css
```diff
@@ -1,3 +1,64 @@
...
+  padding: 0;
+}
+
+.nav__item {
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: normal;
+  text-transform: uppercase;
+  margin-left: 20px;
```
mvjl000, line 39: This `margin-left` won't be needed ;) Check my next comment

src/style.css
```diff
@@ -1,3 +1,64 @@
...
+  padding: 7px 0;
+}
+
+.nav__item:first-child {
+  margin-left: 0;
+}
```
mvjl000, line 45: This works however it can be done in a better, more elegant way I'd say.

Take a look at the last point of the checklist:
_Selector :not(:first-child) has been used while setting indents in navigation_

Don't set the maring in the `nav__item`. Instead do it in a new selecor with `:not(:first-child)` pseudo-classes

src/style.css
```diff
@@ -1,3 +1,64 @@
...
+  box-sizing: border-box;
+  padding: 23px 0;
+}
+
+a.is-active {
+  color: #00acdc;
+  border-bottom: 4px solid #00acdc;
+  border-radius: 2px;
+  padding-bottom: 19px;
+}
```
mvjl000, line 60: The blue line under selected tab should be done with a pseudo element (like `after`). Please try it that way ;)

mvjl000: Btw I don't think that border-radius is necessary

mvjl000 (changes requested):
Just a couple of changes needed 😁

mbruhler (changes requested):
please add changes mentioned by others, this was just develop to master merge

src/style.css
```diff
@@ -1,3 +1,75 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap");
```
JoyCoffeeAddict, line 1: typically, we import fonts in html files, also choose only the weights that you are using in the project.

src/style.css
```diff
@@ -1,3 +1,75 @@
...
+:root {
+  --color: #00acdc;
+  --logo-size: 40px;
+}
+
 body {
   margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+  background-color: black;
```
JoyCoffeeAddict, line 12: Typically we use `#000` hex notation for colors.

src/style.css
```diff
@@ -1,3 +1,75 @@
...
+
+a.is-active {
+  color: var(--color);
+  position: relative;
+}
+
+a.is-active::after {
+  content: "";
+  display: flex;
+  border-bottom: 4px solid #00acdc;
```
JoyCoffeeAddict, line 64: Use a variable here.

JoyCoffeeAddict (changes requested):

src/style.css
```diff
@@ -1,3 +1,75 @@
...
+  padding: 7px 0;
+}
+
+.nav__item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav__item a {
+  text-decoration: none;
+  color: #000;
```
mbruhler, line 46: put this color also to the variable, because it's used more than once in the code

src/style.css
```diff
@@ -1,3 +1,75 @@
+:root {
+  --color-primary: #00acdc;
+  --logo-size: 40px;
+}
+
 body {
   margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+  background-color: #000;
```
mbruhler, line 10: move to css var

mbruhler (changes requested):
Almost there!

Please move one color to the css variable and we're good to go!

mbruhler (approved):
Accepted :)

PR #3693, opened 2023-08-11
add task solution

markiuan:


markiuan (commented):
'add a normal solution'

lerastarynets (approved):
You're showing incredible skills, keep it going!

PR #3752, opened 2023-08-23
moyo-header

Krzychu92:


darokrk (changes requested):
Test still not passing. Please compare Figma's designs to your solution:

image

You have differences with font size, weight, color, active link does not blue underline by default, active link underline looks different compared to designs.
image (2)


image (1)

PR #3753, opened 2023-08-23
add task solution

denzzlle171:
- [DEMO LINK](https://denzzlle171.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://denzzlle171.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,51 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700;800&family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+        class="header__logo"
+        src="./images/logo.png"
+        alt="logo-moyo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#apple" class="nav__link">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#apple" class="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item" >
+            <a
+            href="#apple"
+            class="nav__link"
+            data-qa="hover"
+            >Laptops & Computers</a>
+          </li>
+          <li class="nav__item">
+            <a href="#apple" class="nav__link">Gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#apple" class="nav__link">Tablets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#apple" class="nav__link">Photo</a>
+          </li>
+          <li class="nav__item">
+            <a href="#apple" class="nav__link">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,88 @@
+
 body {
   margin: 0;
+  padding: 0;
+  background-color: #ebebe4;
+  font-family: Roboto, Inter, Helvetica, sans-serif;
+  box-sizing: border-box;
+  --color-selected: #00acdc;
+}
+
+.header {
+  display: flex;
+  margin: auto;
+  justify-content: space-between;
+  align-items: center;
+  width: 1200px;
+  padding: 0 50px;
+  background-color: #fff;
+  box-sizing: border-box;
+}
+
+.header__logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+}
+
+.logo {
+  display: inline-block;
+  text-decoration: none;
+  background-color: red;
+}
+
+.nav__list {
+  margin: 0;
+  padding: 0;
+  display: flex;
+}
+
+.nav__list :first-child {
+  margin: 0;
+}
+
+.nav__item {
+  position: relative;
+  list-style: none;
+  margin-left: 20px;
+
+}
+
+.nav__link {
+  display: flex;
+  flex-direction: column;
+  line-height: 60px;
+  text-decoration: none;
+  text-transform: uppercase;
+  cursor: pointer;
+  font-size: 12px;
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,88 @@
...
+
+.nav__list :first-child {
+  margin: 0;
+}
+
+.nav__item {
+  position: relative;
+  list-style: none;
+  margin-left: 20px;
+
```
StasSokolov, line 48: remove spaces please

src/style.css
```diff
@@ -1,3 +1,88 @@
...
+  box-sizing: border-box;
+  --color-selected: #00acdc;
+}
+
+.header {
+  display: flex;
+  margin: auto;
+  justify-content: space-between;
+  align-items: center;
+  width: 1200px;
```
StasSokolov, line 16: don't hardcode exact width

StasSokolov (changes requested):
Great job, left to fix only few comments!)

src/style.css
```diff
@@ -1,3 +1,86 @@
...
+  box-sizing: border-box;
+  --color-selected: #00acdc;
+}
+
+.header {
+  display: flex;
+  margin: auto;
+  justify-content: space-between;
+  align-items: center;
+  min-width: 1200px;
```
Oleksandr-kopaevich, line 16: Use a combination of width: 100% to force the element to take all parent widths but also add max-width: 1200px to limit it. Look at the visual bug for your current approach.
image

src/style.css
```diff
@@ -1,3 +1,86 @@
+
 body {
   margin: 0;
+  padding: 0;
+  background-color: #ebebe4;
+  font-family: Roboto, Inter, Helvetica, sans-serif;
+  box-sizing: border-box;
```
Oleksandr-kopaevich, line 7: Usually we apply border box property only once for all elements on a page.
```
*, *::before, *::after {
  box-sizing: border-box;
}
```
See details here https://css-tricks.com/almanac/properties/b/box-sizing/

src/style.css
```diff
@@ -1,3 +1,86 @@
...
+}
+
+.header__logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+}
+
+.logo {
+  display: inline-block;
```
Oleksandr-kopaevich, line 29: Why do you need an inline-block here? As for me it doesn't change anything

src/style.css
```diff
@@ -1,3 +1,86 @@
...
+  width: 100%;
+  border-radius: 8px;
+  background-color: var(--color-selected);
+}
+
+.nav__link:hover {
+  color: var(--color-selected);
+}
+
+@media (max-width: 1024px) {
```
Oleksandr-kopaevich, line 82: This rule says that for screen sizes that smaller than 1024px we apply css rule for the element with class name header that makes header to be at least 1024px. You are trying to insert bigger element into smaller screen. Rewrite this or remove after fixing the previous comment related to width: 100% + max-width

Oleksandr-kopaevich (changes requested):
@denzzlle171 Good job, but there is one bug with header width and some moments to improve, see comments for more details.

src/style.css
denzzlle171: if you narrow the screen any further, it breaks the layout, so I think that 

@media (max-width: 1024px) {
  .header {
    width: 1024px;
  }
}

still need as minimum size 
changed min-width: to width: 1024px;
or delete this @media completely after all

src/index.html
```diff
@@ -6,9 +6,58 @@
...
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+        class="header__logo"
+        src="./images/logo.png"
+        alt="logo-moyo">
```
superpooperxxx, line 20: The correct formatting should look like this: 
```
<img
  class="header__logo"
  src="./images/logo.png"
  alt="logo-moyo"
>
```

src/index.html
```diff
@@ -6,9 +6,58 @@
...
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+
+          <li class="nav__item" >
+            <a
+            href="#laptops-&-computers"
+            class="nav__link"
+            data-qa="hover"
+            >Laptops & Computers</a>
```
superpooperxxx, line 41: Also formatting:
```
<a
  href="#laptops-&-computers"
  class="nav__link"
  data-qa="hover"
>
  Laptops & Computers
</a>
```

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+  --color-selected: #00acdc;
+}
+
+.header {
+  display: flex;
+  margin: auto;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+  max-width: 1200px;
```
superpooperxxx, line 19: Not sure why you need to limit the width and center in. Let it take whole width

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+}
+
+.header__logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+}
+
+.logo {
+  text-decoration: none;
```
superpooperxxx, line 31: You do the same for other links, I suggest resetting it for all anchor elements

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+  background-color: red;
```
superpooperxxx, line 32: Not sure that you need this

src/style.css
superpooperxxx: This rule is an overkill for this task

superpooperxxx (changes requested):
Good job🔥 Pay close attention to the styles, cause your code should always be clean

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  cursor: pointer;
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  color: #000;
+}
+
+.is-active {
+  color: var(--color-selected);
+}
```
superpooperxxx, line 59: Combine those styles with `nav__link:hover`

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  background-color: #ebebe4;
+  font-family: Roboto, Inter, Helvetica, sans-serif;
+  --color-selected: #00acdc;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
```
superpooperxxx, line 17: Don't need to specify it

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  display: flex;
+  flex-direction: column;
+  line-height: 60px;
+  text-decoration: none;
+  text-transform: uppercase;
+  cursor: pointer;
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  color: #000;
```
superpooperxxx, line 54: Redundant styles:
1. `flex-direction: column`
2. `cursor: pointer;`
3. `font-style: normal;`

superpooperxxx (approved):
Well done on fixing all mentor's comments🔥 I approve the home task, take my notes as an improvement

PR #3756, opened 2023-08-24
add task solution moyo header

Kuper18:
Hi! Here is my solution!

demo link https://kuper18.github.io/layout_moyo-header/


StasSokolov (approved):
Good job👍 i'm approving it)

PR #3782, opened 2023-09-02
add task solution

OlegPopovych:
https://OlegPopovych.github.io/layout_moyo-header/
https://OlegPopovych.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,12 +3,57 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin = "anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+          src="./images/logo.png"
+          alt="logo"
+          class="logo__link">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#laptopsandcomputers"
+              class="nav__link"
+              data-qa="hover"
+            >Laptops & Computers</a>
+          </li>
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__link">Gadgets</a>
+          </li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -3,12 +3,57 @@
...
+      crossorigin = "anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+          src="./images/logo.png"
+          alt="logo"
```
superpooperxxx, line 22: I would suggest getting used to write some more descriptive `alt`, what do you think about adding the name of the company logo represents ?

src/index.html
```diff
@@ -3,12 +3,57 @@
...
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#laptopsandcomputers"
+              class="nav__link"
+              data-qa="hover"
+            >Laptops & Computers</a>
```
superpooperxxx, line 41: Good job, on keeping your code tight, but small fix to this is needed:
```
<a
  href="#laptopsandcomputers"
  class="nav__link"
  data-qa="hover"
>
  Laptops & Computers
</a>
```

src/style.css
```diff
@@ -1,3 +1,63 @@
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  --color-blue: #00acdc;
+  --gap: 20px;
```
superpooperxxx, line 5: Usually it is better to define CSS variables inside `:root` or `html` selectors, it is not a mistake though, just a recommendation. The same for `font-family`

src/style.css
```diff
@@ -1,3 +1,63 @@
...
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
```
superpooperxxx, line 16: Setting `width` and `height` on inline element is redundant. Try turning off those styles in browser dev tools.

src/style.css
```diff
@@ -1,3 +1,63 @@
...
+  margin: auto 0;
+}
+
+.nav__list {
+  padding: 0;
+  margin: 0;
+  list-style: none;
+  display: flex;
+  align-items: stretch;
```
superpooperxxx, line 25: This is a default value for `align-items` property, no need to specify it explicitly.

src/style.css
```diff
@@ -1,3 +1,63 @@
...
+}
+
+.nav__item {
+  display: flex;
```
superpooperxxx, line 29: Are you sure you need this line ?

superpooperxxx (changes requested):
Good job, on making this task. Overall looks really good🔥 Added some comments to improve.
Also general suggestion is moving all 'reseting styles' to the top, e.g. reseting of padding or margins, etc.

src/style.css
OlegPopovych: Yeah, it is not necessary

src/style.css
OlegPopovych: Yeah, it is not necessary

src/index.html
OlegPopovych: I've had some doubts about this. Thanks

src/index.html
OlegPopovych: Yes. It looks much better

src/style.css
OlegPopovych: Thanks, it's better practice

src/style.css
OlegPopovych: Without this parameter area above and below text isn't clickable

src/style.css
OlegPopovych: Interesting thing,   "margin: 0;" don't works into :root

vasylmiskiv: Cause :root doesn't have any wrappers for margins

vasylmiskiv (approved):
GJ

PR #3786, opened 2023-09-03
add task solution

koros-rk:
- [DEMO LINK](https://koros-rk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://koros-rk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,55 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="container">
+
+      <a href="#home" class="logo">
+        <img src="images/logo.png" alt="MOYO">
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#" class="nav__link is-active">Apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">Samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link"
+            data-qa="hover">Laptops & Computers</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">Gadgets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">Tablets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">Photo</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">Video</a>
+          </li>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,55 @@
...
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#" class="nav__link"
+            data-qa="hover">Laptops & Computers</a>
+          </li>
```
superpooperxxx, line 38: The formatting should look like this:
```
<li class="nav__item">
  <a
    href="#"
    class="nav__link"
    data-qa="hover"
  >
    Laptops & Computers
  </a>
</li>
```

src/index.html
```diff
@@ -7,8 +7,55 @@
...
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="container">
+
```
superpooperxxx, line 16: I like that you add spaces, but it is common practice to add them only between siblings elements not between parent and child

src/style.css
```diff
@@ -1,3 +1,71 @@
-body {
+html body {
   margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+  --text-active: hsl(193, 100%, 43%);
+}
```
superpooperxxx, line 6: 1. Not sure why you refer to body selector like this `html body`, it can be just body
2. CSS variables and global font styles are usually written inside `:root` or `html` selectors (this is just a common practice not an obligation).

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+
+.container {
+  display: flex;
+  flex-direction: row;
```
superpooperxxx, line 10: This is a default value for `flex-direction`, no need to specify it explicitly.

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+  justify-content: space-between;
+  align-items: center;
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
```
superpooperxxx, line 17: <a> is an inline element by default, so specifying `width` and `height` on it doesn't do anything. Try turning off those styles in browser dev tools.

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+  margin-left: 50px;
+}
+
+.nav {
+  margin-right: 50px;
+}
```
superpooperxxx, line 23: I suggest adding padding to `header` class.

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+
+.nav__item {
+  display: block;
```
superpooperxxx, line 26: Redundant styles.

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+  margin-left: 20px;
+}
+
+.nav__list {
+  margin: 0;
+  padding: 0;
+  display: flex;
+  flex-direction: row;
+  list-style: none;
```
superpooperxxx, line 35: I suggest separating the `reset` styles from other. Try writing it above as:
```
ul {
  padding: 0;
  margin: 0;
  list-style: none;
}
```
Do the same for every reset styles

superpooperxxx (changes requested):
Overall good job🔥, just some improvements to write cleaner code🤗

33Orange (approved):

33Orange (approved):
Good job

PR #3818, opened 2023-09-04
moyo header Leonid Pijavskiy

HardCodWorker:
[Demo](https://HardCodWorker.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://HardCodWorker.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" id ="Home">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
@@ -7,8 +7,61 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin = "anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class = "header">
+      <a href="#Home" class ="logo">
+        <img
+          src ="images/logo.png"
+          alt ="Moyo"
+          class ="logo__img"
+        >
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="Apple"
+              class ="nav__link is-active"
+              >Apple
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="Samsung" class ="nav__link">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="Smartphones" class ="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="Laptops & Computers"
+              class ="nav__link"
+              data-qa ="hover">Laptops & Computers
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="Gadgets" class ="nav__link">Gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a href="Tablets" class ="nav__link">Tablets</a>
+          </li>
+          <li class="nav__item">
+            <a href="Photo" class ="nav__link">Photo</a>
+          </li>
+          <li class="nav__item">
+            <a href="Video" class ="nav__link">Video</a>
+          </li>
+        </ul>
+
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,77 @@
 body {
   margin: 0;
+
+}
+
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -1,14 +1,67 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class = "header">
+      <a href="#Home" class ="logo">
+        <img
+          src ="images/logo.png"
+          alt ="Moyo"
+          class ="logo__img"
+        >
+      </a>
```
superpooperxxx, line 26: Don't add those spaces when you define attributes
```
<img
  src="images/logo.png"
  alt="Moyo"
  class="logo__img"
>
```

src/index.html
```diff
@@ -1,14 +1,67 @@
...
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="Apple"
+              class ="nav__link is-active"
+              >Apple
+            </a>
```
superpooperxxx, line 35: 1. Checklist p.6
2. [HTML formatting](https://mate-academy.github.io/style-guides/htmlcss.html#html-formatting)

src/index.html
```diff
@@ -1,14 +1,67 @@
...
+          </li>
+          <li class="nav__item">
+            <a href="Smartphones" class ="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="Laptops & Computers"
+              class ="nav__link"
+              data-qa ="hover">Laptops & Computers
+            </a>
```
superpooperxxx, line 48: Checklist p.6

src/style.css
```diff
@@ -1,3 +1,77 @@
 body {
   margin: 0;
+
+}
+
+.logo__img {
+  padding: 10px 0;
```
superpooperxxx, line 7: Redundant, cause `align-items: center` on parent element will anyway center it vertically

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+  width: 40px;
+}
+
+.nav {
+  font-size: 12px;
+  font-weight: 500;
+  font-family: "Roboto", sans-serif;
+  display: block;
+
+}
```
superpooperxxx, line 19: 1. Move font related styles to `html` or `:root` selector, cause it is a good practice and those styles are actually 'global'
2. `display: block;` is redundant

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+
+.nav__list {
+  margin: 0;
+  list-style: none;
+  padding: 0;
```
superpooperxxx, line 24: Separate `reset type` styles from the rest by moving them to the top of the file. Later such styles will lay in separate file.

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+
+}
+
+.nav__link {
+  line-height: 60px;
+  height: 60px;
+  display: block;
+  text-decoration: none;
+  color: #000;
+  white-space: nowrap;
```
superpooperxxx, line 35: 1. Not sure why you need it
2. `position: relative;` here is also redundant

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+  justify-content: space-between;
+
+}
+
+.nav__item:not(:first-child) {
+
+  display: block;
+
+  margin-left: 20px;
+}
```
superpooperxxx, line 55: 1. I like this solution🔥
2. Remove those extra spaces
3. `display: block;` is redundant

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+
+.is-active {
+  color: #00acdc;
+  position: relative;
+}
```
superpooperxxx, line 60: Combine with `.nav__link:hover` (color property)

superpooperxxx (changes requested):
Hey, overall good job🔥 I added some recommendations, but pay attention to checklist, cause this is always gonna be a reason for reject

src/style.css
HardCodWorker: If I remove:
Display block - than the blue undeline jumps straight under the "Äpple" text, whilst I need the line to be in the bottom of the element

Text Decoration - removes the undelines under the links

Color: is not inherited by the parent elements (for the links)

White-space - does not let the long names to wrap - as it affects when the screen size is reduced

Position relative - that was just used for testing so I am removing it =)

HardCodWorker:
@superpooperxxx please check commit with "final changes" as I have run into some issues during deploying which caused several commits...

src/style.css
superpooperxxx: oh, I meant only `white-space` property, this is a little overkill here, but well done for knowing the reason for usage

superpooperxxx (changes requested):
In `final changes` I can see only `index.html` file. And still this `href="#Laptops & Computers"` goes against Checklist p.6

superpooperxxx (changes requested):
.

HardCodWorker:
Could you please check now if all files are commited? Anyway will attend QnA session to spcify the issue...

superpooperxxx (approved):
Yep, now I see🔥 Good job

PR #3819, opened 2023-09-04
MOYO _header

BetterSol:
https://BetterSol.github.io/layout_moyo-header/
https://BetterSol.github.io/layout_moyo-header/report/html_report/


src/index.html
```diff
@@ -7,8 +7,41 @@
...
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="moyo__header">
```
33Orange, line 19: We use 2 underlines for elements inside some blocks.

For example:

`.header`- block
`.header__logo` - element inside this block
`.header__logo--green` - modifier for this element which changes the default behavior

33Orange (changes requested):
Almost done.

Just one moment with the BEM styles system

src/index.html
BetterSol: Is it correct that I have to change the class name from "moyo__header" to "header__moyo" instead?

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+}
+
+.nav__block {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
```
superpooperxxx, line 34: 1. `flex-direction: row` is the default value, no need to specify it.
2. I recommend moving reseting styles to the top and apply them for all `<ul>` (Usually you don't want those padding and bullets in every list)
```
ul {
  list-style: none;
   padding: 0;
   margin: 0;
}
```

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  display: flex;
+  text-decoration: none;
+  text-transform: uppercase;
+  color: var(--text-color);
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 14px;
+  position: relative;
+  padding: 23px 0;
+}
```
superpooperxxx, line 50: No need in `position: relative` here, it should be only in `is-active`

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+
+.nav__link:hover {
+  color: var(--primary_color);
+  cursor: pointer;
+}
+
+.is-active {
+  color: var(--primary_color);
+  position: relative;
+}
```
superpooperxxx, line 60: 1. Links by default have `cursor: pointer`. Redundant.
2. I suggest combining color setting
```
.nav__link:hover,
.is-active:active,
.is-active {
  color: var(--primary_color);
}

.is-active {
  position: relative;
}
```

src/style.css
```diff
@@ -1,3 +1,74 @@
+html {
+  font-family: Roboto, serif;
+}
```
superpooperxxx, line 3: `Roboto` is a `sans-serif` font

superpooperxxx (approved):
Well done🔥 I added recommendations which will improve your code, study them and try to implement in future tasks

PR #3825, opened 2023-09-05
My solution

GTXtab:


src/style.css
```diff
@@ -1,3 +1,73 @@
...
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.nav__link:active {
+  color: #00acdc;
+  box-shadow: 0 4px 0 #00acdc;
+}
```
TarasHoliuk, line 57: It's not required in the task, so you can just remove it

but if you want to keep it, it should look the same as the line in `is-active` class:
image

src/style.css
```diff
@@ -1,3 +1,73 @@
...
+
+.is-active {
+  position: relative;
+  color: #00acdc;
```
TarasHoliuk, line 61: move repeatable colors to variables

src/style.css
```diff
@@ -1,3 +1,73 @@
+
+@import url("https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap");
```
TarasHoliuk, line 2: import only font with weights you use here

TarasHoliuk (changes requested):
Add correct links to Demo and Test Report in the readme and in the PR description

etojeDenys (approved):
lgtm

PR #3826, opened 2023-09-06
make moyo header

YePolishchuk:
- [DEMO LINK](https://YePolishchuk.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://YePolishchuk.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,48 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+      crossorigin = "anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img
+        src="images/logo.png"
+        alt="MOYO"
+        class="logo"
+        >
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a href="#laptops-&-computers"
+              class="nav__link"
+              data-qa="hover">Laptops & Computers</a>
+          </li>
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__link">Gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#tablets" class="nav__link">Tablets</a>
+          </li>
+          <li class="nav__item">
+            <a href="#photo" class="nav__link">Photo</a>
+          </li>
+          <li class="nav__item">
+            <a href="#video" class="nav__link">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,8 +7,48 @@
...
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a href="#laptops-&-computers"
+              class="nav__link"
+              data-qa="hover">Laptops & Computers</a>
```
vasylmiskiv, line 36: Move Laptops & Computers on the new line change tag style like so

src/index.html
```diff
@@ -7,8 +7,48 @@
...
     <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+      crossorigin = "anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <img
```
vasylmiskiv, line 17: Wrap up img inside `<a>` tag

src/style.css
```diff
@@ -1,3 +1,73 @@
...
+
+.header {
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+  text-align: center;
+}
+
+.logo {
+  padding: 10px 0;
```
vasylmiskiv, line 16: If you wrap up your image inside  `<a>` tag  you dont need to use paddings in this case

src/style.css
```diff
@@ -1,3 +1,73 @@
...
+  display: block;
+  height: 40px;
```
vasylmiskiv, line 18: The image alraedy has correct height and width so you should remove height and width

src/style.css
```diff
@@ -1,3 +1,73 @@
...
+  text-decoration: none;
+  position: relative;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+/*  :active*/
+.is-active {
+  text-transform: uppercase;
```
vasylmiskiv, line 54: This element already has these properties:
```
  display: block;
  height: 60px;
  line-height: 60px;
  text-decoration: none;
  position: relative;
```
Because of `nav__link`. So you should remove the properties above

After that you need to merge styles `color: #00acdc;`. After that, as you can see `is-active` and `nav__link:hover` have the same color. Let's merge these styles

src/style.css
```diff
@@ -1,3 +1,73 @@
...
+/*  :active*/
+.is-active::after {
+  content: "";
+  position: absolute;
+  display: block;
+  background-color: #00acdc;
+  height: 4px;
+  width: 100%;
+  bottom: 0;
+  left: 0;
```
vasylmiskiv, line 71: In case with `display:block` you don't need to use `left:0`, so remove it

src/style.css
```diff
@@ -1,3 +1,73 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
```
vasylmiskiv, line 5: Add variable for your color `#00acdc;` here because the color often repeat in the code. And use variable when you need

vasylmiskiv (changes requested):
Need to add some changes

src/style.css
```diff
@@ -1,3 +1,67 @@
+html {
+  --colour: #00acdc;
+}
```
superpooperxxx, line 3: `--colour` name is not descriptive enough.

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+  height: 60px;
+  line-height: 60px;
+  color: #000;
+  text-decoration: none;
+  position: relative;
+}
+
+.nav__link:hover {
+  color: var(--colour);
+}
```
superpooperxxx, line 51: You can combine it with `.is-active`

superpooperxxx (approved):
GJ🔥

PR #3842, opened 2023-09-11
add task solution

VKdrvtsv:
Have some problem with a test "link whith class is-active". On the site all is ok, but tests are failed.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,89 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+html {
+  --header-height: 60px;
+}
+
 body {
   margin: 0;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+  flex-shrink: 0;
+  margin-left: 50px;
+}
+
+.nav {
+  margin-right: 50px;
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  position: relative;
+}
+
+.nav__item {
+  position: relative;
+  cursor: pointer;
+}
+
+.nav__item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav__link {
+  line-height: var(--header-height);
+  text-decoration: none;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,89 @@
...
+
+.nav__item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav__link {
+  line-height: var(--header-height);
+  text-decoration: none;
+  color: #000;
+  font-family: "Roboto", sans-serif;
```
mykhalenych, line 50: move to html selector

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+  height: 4px;
+  border-radius: 8px;
+  background: #00acdc;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.nav__link:active::after {
```
mykhalenych, line 75: why do you create two lines?
delete it use only is-active::after

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+  right: 0;
+  left: 0;
+  content: "";
+  height: 4px;
+  flex-shrink: 0;
+  border-radius: 8px;
+  background: #00acdc;
+}
+
+.nav__link:visited {
```
mykhalenych, line 87: redundant

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+  position: absolute;
+  right: 0;
+  left: 0;
+  height: 4px;
+  border-radius: 8px;
+  background: #00acdc;
+}
+
+.nav__link:hover {
+  color: #00acdc;
```
mykhalenych, line 72: create a variable with color

mykhalenych (changes requested):

src/style.css
```diff
@@ -1,3 +1,75 @@
+* {
+  margin: 0;
+  padding: 0;
```
anastasiiavorobiova, line 3:
```suggestion
  padding: 0;
  box-sizing: border-box;
```

anastasiiavorobiova (approved):
Great work!

PR #3883, opened 2023-09-25
develop header

mttony:
- [DEMO LINK](https://mttony.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://mttony.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,69 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Roboto, sans-serif;
+  scroll-behavior: smooth;
+
+  --blue-text-color: #00acdc;
+}
+
 body {
   margin: 0;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  margin-left: 20px;
+}
+
+.nav__link:hover {
+  color: var(--blue-text-color);
+}
+
+.is-active {
+  color: var(--blue-text-color);
+}
```
etojeDenys, line 57: we can join it with a comma

etojeDenys (changes requested):

src/style.css
mttony: fixed!
thanks!

etojeDenys (approved):
awesome

PR #3887, opened 2023-09-26
add task solution

AlinaMyronenko:
in the nav tag, fails the test
please explain how to write correctly

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,89 @@
-body {
+html {
   margin: 0;
+}
+
+body {
+  margin: 0 ;
+}
+
+.header {
+  display: flex;
+  flex-direction: row-reverse;
+  align-items: center;
+  max-width: 1200px;
+  background: rgba(255, 255, 255, 1);
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+  padding: 0 40px;
+}
+
+.logo {
+  order: 1;
+  position: absolute;
+  width: 40px;
+  height: 40px;
+  top: 10px;
+  left: 50px;
+
+}
+
+.nav {
+  margin: 0;
+}
+
+.nav-list {
+  height: 58px;
+  list-style: none;
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  margin: 0;
+  padding: 0;
+}
+
+.link {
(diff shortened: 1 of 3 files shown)


anastasiia-tilikina (changes requested):
please provide demo link and report link

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+  align-items: center;
+  max-width: 1200px;
+  background: rgba(255, 255, 255, 1);
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+  padding: 0 40px;
+}
+
+.logo {
+  order: 1;
+  position: absolute;
```
OleziO, line 21: You can create logo without absolute position.

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+
+.nav {
+  margin: 0;
+}
+
+.nav-list {
+  height: 58px;
+  list-style: none;
+  display: flex;
+  justify-content: flex-start;
```
OleziO, line 37:
```suggestion
```

It is default value

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+}
+
+.link {
+  padding: 23px 0;
+  margin: 0 10px;
+  flex-shrink: 0;
+  align-self: center;
+  text-decoration: none;
+  position: relative;
+  font-family: "Roboto", sans-serif;
```
OleziO, line 50: Move to the body or html selector and remove it from all other places.

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+  font-size: 12px;
+  font-style: normal;
```
OleziO, line 52:
```suggestion
```

Default value

src/style.css
```diff
@@ -1,3 +1,89 @@
...
+    left: 509px;
+  }
+}
+
+.link:hover {
+  color: rgba(0, 172, 220, 1);
+}
+
+.is-active {
+  color: rgba(0, 172, 220, 1);
```
OleziO, line 76: Move all repeatable colors to the variables.

OleziO (changes requested):
Ask the question with test in the `fe_chat`

PR #3933, opened 2023-10-10
Solution 4

aureliagorzna:


mvjl000 (changes requested):
Please add deploy 🙏

mbruhler (changes requested):
Good work but some changes are needed.
image
As you can see, the bottom line is too short. It should cover the whole width of nav element.

Also clear the comments for the final version.

PR #3937, opened 2023-10-10
task

adrian200002:
- [DEMO LINK](https://adrian200002.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://adrian200002.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,25 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <nav class="nav">
+        <div class="container">
+          <div class="logo"><img src="images/logo.png" alt="logo"></div>
+          <ul class="list">
+            <li class="elements active"><a class="nav__link apple" href="apple.html">Apple</a></li>
+            <li class="elements"><a class="nav__link" href="samsung.html">Samsung</a></li>
+            <li class="elements"><a class="nav__link" href="smartphones.html">Smartphones</a></li>
+            <li class="elements"><a class="nav__link laptops" href="latops.html">Laptops & Computers</a></li>
+            <li class="elements"><a class="nav__link" href="gagets.html">Gadgets</a></li>
+            <li class="elements"><a class="nav__link" href="tablets.html">Tablets</a></li>
+            <li class="elements"><a class="nav__link" href="photo.html">Photo</a></li>
+            <li class="elements"><a class="nav__link" href="video.html">Video</a></li>
+          </ul>
+        </div>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,69 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 14px;
+  letter-spacing: 0;
+}
+
+nav > .container {
+  display: flex;
+  width: 100%;
+  justify-content: space-between;
+}
+
+a {
+  text-decoration: none;
+}
+
+.nav__link {
+  color: #000;
+}
+
+.apple {
+  color: #00acdc;
+}
+
+.laptops {
+  color: #00acdc;
+}
+
+.list {
+  display: flex;
+  align-items: center;
+  list-style: none;
+  text-transform: uppercase;
+}
+
+.logo {
+  margin-top: 10px;
+  margin-left: 50px;
+}
+
+.elements {
+  margin-right: 20px;
+  color: black;
+  position: relative;
+}
+
+ul:not(:first-child) {
+  margin-right: 50px;
+}
+
+.active::after {
+  content: "";
+  display: block;
+  position: absolute;
+  width: 37px;
+  height: 4px;
+  top: 26.5px;
+  border-radius: 8px;
+  background: #00acdc;
+}
+
+a:hover {
+  text-decoration: underline;
+  text-decoration-color: #00acdc;
+  text-decoration-thickness: 2px;
 }
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Moyo header
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_moyo-header/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_moyo-header/report/html_report/)
+- [DEMO LINK](https://adrian200002.github.io/layout_moyo-header/)
+- [TEST REPORT LINK](https://adrian200002.github.io/layout_moyo-header/report/html_report/)
 
 > Follow [these instructions](https://mate-academy.github.io/layout_task-guideline/#how-to-solve-the-layout-tasks-on-github)
 


src/style.css
```diff
@@ -1,3 +1,69 @@
...
+a {
+  text-decoration: none;
+}
+
+.nav__link {
+  color: #000;
+}
+
+.apple {
+  color: #00acdc;
```
mvjl000, line 25: This color could definitely be in a variable

mvjl000 (changes requested):
Heey, your deploy seems not to be working, I get a 404 error. Could you repeat the process according to the [guideline](https://mate-academy.github.io/layout_task-guideline/)?

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  top: 26.5px;
+  border-radius: 8px;
+  background: #00acdc;
+}
+
+a:hover {
+  text-decoration: underline;
+  text-decoration-color: #00acdc;
+  text-decoration-thickness: 2px;
 }
```
JoyCoffeeAddict, line 69: you should not add a underline for `:hover` styles, it's not behaving like this in figma
Also, you hardly ever want to style all of your `<a>` tags on your page, you should use more specific selector.

src/index.html
```diff
@@ -7,8 +7,25 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <nav class="nav">
+        <div class="container">
+          <div class="logo"><img src="images/logo.png" alt="logo"></div>
+          <ul class="list">
+            <li class="elements active"><a class="nav__link apple" href="apple.html">Apple</a></li>
```
JoyCoffeeAddict, line 18: `apple` should be `is-active`

src/index.html
```diff
@@ -7,8 +7,25 @@
...
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <nav class="nav">
+        <div class="container">
+          <div class="logo"><img src="images/logo.png" alt="logo"></div>
```
JoyCoffeeAddict, line 16: Logo should not be a part of `<nav>` according to readme Tips and Hints. remember that it will also require CSS changes.

src/style.css
```diff
@@ -1,3 +1,69 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 14px;
+  letter-spacing: 0;
```
JoyCoffeeAddict, line 7: You almost never want to change all of the font related styles for all of the elements on the page, select only the elements that need it.

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+}
+
+nav > .container {
+  display: flex;
+  width: 100%;
+  justify-content: space-between;
+}
```
JoyCoffeeAddict, line 14: You can just use `.container` here

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+
+.active::after {
+  content: "";
+  display: block;
+  position: absolute;
+  width: 37px;
+  height: 4px;
+  top: 26.5px;
+  border-radius: 8px;
+  background: #00acdc;
```
JoyCoffeeAddict, line 62: Create a variables for repeated values.

JoyCoffeeAddict, line 62: These styles should be on active link (`<a>` tag), in your HTML `.active` is a class on `<li>`

JoyCoffeeAddict (changes requested):
Remember to resolve all of the previous issues.

mvjl000 (changes requested):
You haven't apply previous mentors' suggestions. Please do so or in case of further difficulties join the QnA

src/style.css
```diff
@@ -1,3 +1,65 @@
+:root {
+  --color1: #00acdc;
+  --color2: #000;
+  --blockheight: 4px;
+  --blocktopgap: 26.5px;
+  --blockwidth: 37px;
+  --border1: 8px;
```
mvjl000, line 7: Remember that you should create CSS variables only for values that appear more than once!

Also, color1 and color2 are not the most descriptive names

src/style.css
```diff
@@ -1,3 +1,65 @@
...
+}
+
+.logo {
+  margin-top: 10px;
+  margin-left: 50px;
+}
+
+.nav__link {
+  text-decoration: none;
+  padding: 5px 0;
```
mvjl000, line 30: link should have more clickable space - increase the vertical padding

src/style.css
```diff
@@ -1,3 +1,65 @@
...
+}
+
+.list {
+  display: flex;
+  align-items: center;
+  list-style: none;
+  text-transform: uppercase;
+}
+
+.elements {
```
mvjl000, line 46: This is not a good class name for this element.
It is plurar and non descriptive at all. I suggest `nav__item`

src/style.css
```diff
@@ -1,3 +1,65 @@
...
+.nav__link {
+  text-decoration: none;
+  padding: 5px 0;
+  color: var(--color2);
+}
+
+.apple,
+.laptops {
+  color: var(--color1);
+}
```
mvjl000, line 37: Blue color for apple link should be set by is-active class. And laptops should be blue all the time. Links should be blue on hover.
Btw you don't need those two classes

src/index.html
```diff
@@ -7,8 +7,30 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <div class="container">
```
mvjl000, line 14: You already have a header. You can use it as a 'container'

src/style.css
```diff
@@ -1,3 +1,65 @@
...
+
+.list {
```
mvjl000, line 39: Use BEM - `nav__list`

Btw, your `ul` has some padding on the left side. Try to get rid of that.

src/style.css
```diff
@@ -1,3 +1,65 @@
...
+  text-transform: uppercase;
+}
+
+.elements {
+  margin-right: 20px;
+  color: var(--color2);
+  position: relative;
+}
+
+ul:last-child {
```
mvjl000, line 52: Please use classes for selectors

mvjl000 (changes requested):

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  color: var(--primarycolor);
+  position: relative;
+}
+
+[data-qa="hover"].nav__link {
+  color: var(--secondarycolor);
+}
+
+.nav__list:last-child {
+  margin-right: 50px;
```
mvjl000, line 53: Get rid of this. Same for margin left for logo. Instead give padding of 50px to the parent element of both logo and nav

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+}
+
+.logo {
+  margin-top: 10px;
+  margin-left: 50px;
+}
+
+.nav__link {
+  text-decoration: none;
+  padding: 10px 0;
```
mvjl000, line 26: Give the links more padding!

mvjl000 (changes requested):
Please compare your solution with figma mock carefully

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  font-size: 12px;
+}
+
+.container {
+  display: flex;
+  box-sizing: border-box;
+  width: 100%;
+  justify-content: space-between;
+  align-items: center;
+  padding: 10px 50px;
```
mvjl000, line 18: We need to simplify all your margins and paddings. I think we should stars with padding here. We can get rid of padding top & bottom here, so it doesn't determine the height of the entire header
```suggestion
  padding: 0 50px;
```

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+}
+
+.nav__link {
+  text-decoration: none;
+  padding: 15px 0;
```
mvjl000, line 23: You should play with padding values here. Compare it with figma and finally you will find the correct value. (It should definitely by greater than 15)

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  color: var(--primarycolor);
+}
+
+.nav__link:hover {
+  color: var(--secondarycolor);
+}
+
+.nav__list {
```
mvjl000, line 31: Please reset margin for the list because is your solution there is some margin top & bottom.

mvjl000 (changes requested):
You are very close!

choeqq (changes requested):
Hi, my name is Kacper, I am a mentor in other group, and I'll be helping mentors team with checking homework, until they'll catch up with your group's work 😄 

Please, always make sure to address all the comments in your code, it seems like your solution lacks a little bit a positioning, styles looks good but navigation items aren't aligned as in figma design provided, please try to fix vertical positioning, and you'll be good to go ✅

mbruhler (approved):
Hi,

Finally as expected! Your view looks good now, well done!

Approved.

PR #3939, opened 2023-10-10
add task solution

K4rol4:


src/style.css
```diff
@@ -1,3 +1,122 @@
...
+  justify-content: flex-end;
+  list-style: none;
+  margin-top: 0;
+  padding-top: 0;
+}
+
+.nav__item {
+  margin-right: 20px;
+  margin-top: 12.5px;
+  padding-top: 0;
```
mvjl000, line 49: All of this is too complicated and it's making a bit of mess

src/style.css
```diff
@@ -1,3 +1,122 @@
...
+}
+
+.nav__link {
```
mvjl000, line 52: You should add some vertical padding to the nav link so it has proper height

src/style.css
```diff
@@ -1,3 +1,122 @@
...
+.is-special-colored {
+  color: var(--a-hover-color);
+}
+
+nav a:hover {
+  color: var(--a-hover-color);
+  /*text-decoration: underline;*/
+  /*text-decoration-thickness: 5px;*/
+  /*text-underline-offset: 0.75em;*/
+  /*text-underline-position: under;*/
```
mvjl000, line 97: Remove comments

src/style.css
```diff
@@ -1,3 +1,122 @@
...
+  color: var(--a-hover-color);
+  text-decoration: none;
+  margin-left: 10px;
+  pointer-events: none;
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: -18.5px;
```
mvjl000, line 81: After adding padding to the `a` tag, you can set bottom to 0

src/style.css
```diff
@@ -1,3 +1,122 @@
...
+
+.nav__link:hover::after {
+  content: "";
+  display: block;
+  position: absolute;
+  width: 100%;
+  height: 4px;
+  bottom: 0;
+  border-radius: 8px;
+}
```
mvjl000, line 67: You should only change the color on hover. No additional pseudo-elements needed!

src/style.css
```diff
@@ -1,3 +1,122 @@
...
+  bottom: -18.5px;
+  left: 0;
+  width: 100%;
+  height: 3px;
+  background-color: var(--a-hover-color);
+}
+
+.is-special-colored {
+  color: var(--a-hover-color);
+}
```
mvjl000, line 90: Get rid of unused selectors

mvjl000 (changes requested):
Your solution needs some changes & improvements. Start with things listed by me and I'm sure that you will be able to push it forward 😄

src/style.css
```diff
@@ -1,3 +1,92 @@
...
 body {
+  font-family: Roboto, sans-serif;
   margin: 0;
+  padding: 0;
+  color: black;
+}
+
+a {
+  padding: 22.5px;
+}
```
JoyCoffeeAddict, line 25: you want to add padding to `nav__link` that is a specific `<a>` tag, you don't want to add the padding to all of the `<a>` tags on your page. Also, move these styles to other `nav__link` styles.

src/style.css
```diff
@@ -1,3 +1,92 @@
...
+@font-face {
+  font-family: Roboto;
+  src: url("Roboto/Roboto-Medium.ttf");
+}
+
 body {
+  font-family: Roboto, sans-serif;
   margin: 0;
+  padding: 0;
+  color: black;
```
JoyCoffeeAddict, line 20: you already created variable for this color.

src/style.css
```diff
@@ -1,3 +1,92 @@
+:root {
+  font-family: Roboto, Helvetica, sans-serif;
+  font-size: 12px;
+  --bg-color: #fff;
+  --a-color: #000;
+  --a-hover-color: #00a0e3;
+  margin: 0;
+  padding: 0;
+}
```
JoyCoffeeAddict, line 9: we don't style the `:root`, we only set variables here.

JoyCoffeeAddict (changes requested):
I had some suggestions, but it's crucial to resolve all of the previous issues  before requesting review again.

src/style.css
```diff
@@ -1,3 +1,87 @@
+:root {
+  font-family: Roboto, Helvetica, sans-serif;
+  font-size: 12px;
+  --bg-color: #fff;
+  --a-color: #000;
+  --a-hover-color: #00a0e3;
+}
```
mvjl000, line 7: You should only keep variables here

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+.nav__item {
+  margin-right: 20px;
+  margin-top: 12.5px;
+  padding-top: 0;
+}
+
+.nav__link {
+  display: flex;
+  position: relative;
+  padding: 0 20px;
```
mvjl000, line 53: You need padding top & bottom, not left and right.

More like this:
```suggestion
  padding: 20px 0;
```

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+  color: var(--a-color);
+  text-decoration: none;
+}
+
+.nav__link:hover::after {
+  color: var(--a-hover-color);
+}
```
mvjl000, line 60: I think that's not what you want to do.
You are setting after's text color to blue on hover.

mvjl000 (changes requested):
Your styles on this PR and on the deploy are  different. Have you done tests and the deploy command?

Also please fix the links, they shoud look like this:
https://k4rol4.github.io/layout_moyo-header/
https://k4rol4.github.io/layout_moyo-header/report/html_report/

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+  pointer-events: none;
+}
+
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  width: 100%;
+  height: 3px;
```
mbruhler, line 77: Should be 4px. Refer to figma for dimensions

mbruhler (changes requested):
Please add changes from other mentors.

Refer to figma for dimensions, because you are not using the proper ones.

mbruhler (changes requested):
No changes have been made, but review was requested. Please add changes from comments. Also add links to your solution.

PR #3942, opened 2023-10-10
add solution

dawidpizon:


mbruhler (changes requested):
Student resigned from course.

PR #3977, opened 2023-10-16
add header-menu

olegnovikov10:
[DEMO LINK](https://olegnovikov10.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://olegnovikov10.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -6,9 +6,52 @@
...
+            </a>
+          </div>
+          <nav class="nav">
+            <ul class="nav-list">
+              <li class="nav-list__item">
+                <a href="#" class="nav-list__link is-active">APPLE</a>
+              </li>
+              <li class="nav-list__item">
+                <a href="#" class="nav-list__link ">SAMSUNG</a>
+              </li>
```
BudnikOleksii, line 31:
```suggestion
              <li class="nav-list__item">
                <a href="#" class="nav-list__link is-active">APPLE</a>
              </li>
              
              <li class="nav-list__item">
                <a href="#" class="nav-list__link ">SAMSUNG</a>
              </li>
```
It's better to add a blank line between multiple siblings.

src/style.css
```diff
@@ -1,3 +1,69 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+html {
+  --main-color: #00acdc;
+}
```
BudnikOleksii, line 8: You can add main variables to `:root` selector

src/style.css
```diff
@@ -1,3 +1,69 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
BudnikOleksii, line 4: Don't reset margin/paddings for all elements, only for elements that you actually need

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  margin-left: 20px;
+}
+
+.nav-list__item:first-child {
+  margin-left: 0;
+}
+
+.nav-list__link {
+  display: inline-block;
+  height: 60px;
```
BudnikOleksii, line 44: It's okay in this case, but it's better to avoid hardcoding width/height, you can define your container size with paddings or line-height in your solution.

BudnikOleksii (changes requested):
Good job 👍 
Let's improve your solution slightly

olegnovikov10:
fixed all problems

BudnikOleksii (approved):
🚀 
Approved, don't forget to "Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it."
image

PR #4004, opened 2023-10-20
complete task solution

Dmytro-Kretov:
[DEMO LINK] (https://Dmytro-Kretov.github.io/layout_moyo-header/)
[TEST REPORT LINK] (https://Dmytro-Kretov.github.io/layout_moyo-header/report/html_report/)

- [x] var(--blue-color) is used for a blue color

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,66 @@
-body {
+.body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  --blue-color: #00acdc;
+}
+
+.header {
+  width: auto;
+  display: flex;
+  justify-content: space-between;
+  align-content: center;
+  padding: 0 50px;
+}
+
+.logo__img {
+  margin: 10px 0;
+  width: 40px;
+  height: 40px;
+}
+
+.nav__list {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.nav__item {
+  display: flex;
+  margin-right: 20px;
+}
+
+.nav__item:last-child {
+  margin-right: 0;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,66 @@
-body {
+.body {
   margin: 0;
+  font-family: Roboto, sans-serif;
```
VitaliyBondarenko1982, line 3: good practice to set font-family for html selector.

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+}
+
+.nav__item:last-child {
+  margin-right: 0;
+}
+
+.nav__link {
+  font-size: 12px;
+  font-weight: 500;
+  display: inline;
```
VitaliyBondarenko1982, line 40:
```suggestion
```
looks like redundant . it is default value. better to set some other not inlive display (inline-block, flex, block)

src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  color: var(--blue-color);
+}
+
+.is-active::after {
+  position: relative;
+  content: "";
+  display: block;
+  top: 19px;
+  bottom: 0;
+  /* width: 100%; */
```
VitaliyBondarenko1982, line 62:
```suggestion
  position: absolute;
  content: "";
  left: 0;
  bottom: 0;
  width: 100%;
```
pseudo element should be with positon absolute and be positioned to closest non static element.
we can avoid to hardcode top property.
set for is-active position: relative.

VitaliyBondarenko1982 (approved):
Good job.
Please check comments

PR #4048, opened 2023-10-31
add task solution

hannasavka:


etojeDenys (changes requested):
looks good, but try to fix the tests or feel free to ask for some help in the chat

hannasavka:
tests passed

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+.nav__item {
+  padding: 0;
+  position: relative;
+  display: flex;
+  height: 60px;
+  margin-left: 20px;
+  align-items: center;
+  color: #000;
+  text-decoration: none;
+  background-color: #fff;
```
loralevitska, line 46: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

loralevitska (approved):
Good job!

src/style.css
hannasavka: okay, thank you

PR #4062, opened 2023-11-05
moyo header

DavaiBezDava1:


src/style.css
```diff
@@ -1,3 +1,82 @@
...
 body {
   margin: 0;
+  background-color: rgb(120, 120, 255);
+  padding: 0;
+  box-sizing: border-box;
+  font-family: "Roboto", sans-serif;
+
+}
+
+nav {
```
etojeDenys, line 12:
```suggestion
.nav {
```
better to use class selectors

etojeDenys (changes requested):
add links from readme to the PR description

src/style.css
DavaiBezDava1: - [DEMO LINK](https://DavaiBezDava1.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://DavaiBezDava1.github.io/layout_moyo-header/report/html_report/)

PR #4065, opened 2023-11-06
Done

KlymIzbash:
 Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_moyo- header/blob/master/checklist.md)


VitaliyBodnar (changes requested):
Demo link is not valid, try to redeploy:
Screenshot 2023-11-06 at 14 01 31

PR #4067, opened 2023-11-06
add task solution

KlymIzbash:
 Code follows all the [Code Style Rules ❗️](https://github.com/KlymIzbash/layout_moyo-header/blob/master/checklist.md)


denys-danyliuk (changes requested):
your test link doesn't work, probably you forgot to run tests before deploy, also pay attention to the codestyle, I can tell that you forgot to do that :)
add links to the pr description so it will be easier to check, if you have any questions ask them in fe_chat
Знімок екрана 2023-11-06 о 15 53 37

PR #4082, opened 2023-11-09
add task solution

KlymIzbash:
 Code follows all the [Code Style Rules ❗️](https://github.com/KlymIzbash/layout_moyo-header/blob/master/checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,79 @@
 body {
   margin: 0;
 }
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  padding: 0 50px;
+
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo {
+  display: block;
+  height: 40px;
+  width: 40px;
+}
+
+.logo__image {
+  height: 100%;
+  width: 100%;
+}
+
+.nav__list {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,79 @@
...
+
+.nav__link {
+  position: relative;
+
+  display: flex;
+  align-items: center;
+
+  height: 60px;
+
+  font-family: "Roboto", sans-serif;
```
sTorba24, line 50: set it for html or body tag

sTorba24 (changes requested):
Well done! Little thing to improve

src/style.css
KlymIzbash: done

KlymIzbash: если я сую heighy в body то тесты не проходят

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+
+.nav__link:hover {
+  color: #00acdc;
+}
+
+.nav__link.is-active {
+  color: #00acdc;
+}
+
+.nav__link.is-active::after {
```
Andriy-Shymkiv, line 61:
```suggestion
.is-active {
  color: #00acdc;
}

.is-active::after {
```

Andriy-Shymkiv (approved):
good job! just remove the unnecessary selector specificity

PR #4111, opened 2023-11-14
add task solution

nabokok:
https://nabokok.github.io/layout_moyo-header/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,51 @@
 <!doctype html>
 <html lang="en">
+
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Moyo header</h1>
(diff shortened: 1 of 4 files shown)


artemforsoff:
I don’t see the [checklist](https://github.com/mate-academy/layout_moyo-header#checklist)

src/style.css
```diff
@@ -1,3 +1,69 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap");
+
 body {
+  font-family: "Roboto", sans-serif;
   margin: 0;
+  box-sizing: border-box;
```
artemforsoff, line 6: `box-sizing` is not inherited from the parent element
so it's better to specify for all elements

nabokok:
[x] Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md)

src/index.html
```diff
@@ -1,14 +1,51 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a class="logo" href="#">
+        <img src="images/logo.png" alt="logo">
+      </a>
+      <nav class="menu">
+        <ul class="menu-list">
+          <li>
+            <a href="#" class="list-item is-active">Apple</a>
```
artemforsoff, line 21: I think a class `menu-link` would be better because it’s not clear what kind of list this is

artemforsoff (changes requested):
overall very good job 🔥
I left a couple of minor comments, look at them

artemforsoff (approved):
Good job 🔥
better set `box-sizing` like this `*, *::before, *::after { box-sizing: border-box }`

PR #4124, opened 2023-11-18
Moyo header

OMedvid:
https://OMedvid.github.io/layout_moyo-header/
https://OMedvid.github.io/layout_moyo-header/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,63 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  text-transform: uppercase;
+  font-size: 12px;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background-color: white;
+  padding: 0 50px;
+  width: 100% - 100px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,63 @@
 body {
   margin: 0;
+  font-family: "Roboto", sans-serif;
+  text-transform: uppercase;
+  font-size: 12px;
```
VitaliyBondarenko1982, line 5: general styles for fonts better to set in html selector

src/style.css
```diff
@@ -1,3 +1,63 @@
...
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background-color: white;
+  padding: 0 50px;
+  width: 100% - 100px;
```
VitaliyBondarenko1982, line 14:
```suggestion
```
why do you need such styles for widht?  header width will be on full width of body as it is block element.

VitaliyBondarenko1982 (approved):
Good job.
Comments just for future tasks.

PR #4126, opened 2023-11-19
add task solution

artem8746:
- [DEMO LINK](https://artem8746.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://artem8746.github.io/layout_moyo-header/report/html_report/)

[x] Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,10 +5,71 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <title>Moyo header</title>
+
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
+
+    <title>Moyo header</title>
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="moyo-logo-link">
+        <img
+          class="moyo-logo"
+          src="/images/logo.png"
+          alt="moyo logotype"
+        >
+      </a>
+
+      <nav class="nav">
+        <ul class="nav-list">
+          <li>
+            <a href="#apple" class="is-active nav-link">Apple</a>
+          </li>
+
+          <li>
+            <a href="#samsung" class="nav-link">Samsung</a>
+          </li>
+
+          <li>
+            <a href="#smartphones" class="nav-link">Smartphones</a>
+          </li>
+
+          <li>
+            <a
+              href="#laptops-and-computers"
+              class="nav-link"
+              data-qa="hover"
+            >
+              Laptops & Computers
+            </a>
+          </li>
+
+          <li>
+            <a href="#gadgets" class="nav-link">Gadgets</a>
+          </li>
+
+          <li>
+            <a href="#tablets" class="nav-link">Tablets</a>
+          </li>
+
+          <li>
+            <a href="#photo" class="nav-link">Photo</a>
+          </li>
+
+          <li>
+            <a href="#video" class="nav-link">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <main></main>
+    <footer></footer>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,71 @@
 body {
+  --primary-color: #00acdc;
+
   margin: 0;
+}
+
+.header {
+  box-sizing: border-box;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -5,10 +5,70 @@
...
+      <a href="#" class="moyo-logo-link">
+        <img class="moyo-logo"
+        src="/images/logo.png"
+        alt="moyo logotype"
+        >
+      </a>
+
+      <nav class="header-nav">
+        <ul class="nav-list">
+          <li class="nav-list-item">
```
artemforsoff, line 31: `nav-list-item` can not be block
because a list element without a list does not exist
`nav-list-item` is an element of the block `nav-list`
so [Follow naming convention `block__element--modifier`](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules#bem)

src/style.css
```diff
@@ -1,3 +1,76 @@
 body {
+  --hover-link-color: #00acdc;
   margin: 0;
 }
+
+.header {
+  box-sizing: border-box;
```
artemforsoff, line 7: the rule `box-sizing: border-box` must be specified for all elements

src/style.css
```diff
@@ -1,3 +1,76 @@
 body {
+  --hover-link-color: #00acdc;
```
artemforsoff, line 2: by the name of the variable, I assume that the color is used only on hover,
but in the code, this variable is also used for the active state
better to rename it to a more abstract name, for example `--accent-color`, `--primary-color`

src/style.css
```diff
@@ -1,3 +1,76 @@
...
   margin: 0;
 }
+
+.header {
+  box-sizing: border-box;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  position: absolute;
```
artemforsoff, line 11: what is this rule for?
works without it
try adding content to the tag `<main>` and see the result

src/index.html
```diff
@@ -5,10 +5,70 @@
...
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="moyo-logo-link">
+        <img class="moyo-logo"
+        src="/images/logo.png"
+        alt="moyo logotype"
+        >
+      </a>
+
+      <nav class="header-nav">
```
artemforsoff, line 29: such a menu can also be used in the footer, so in this case it is better not to bind the class name to the place where this element is used
`<nav class="nav">` will be better

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  text-align: center;
+  text-wrap: nowrap;
+  text-transform: uppercase;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 60px;
+  text-decoration: none;
+  height: 60px;
```
artemforsoff, line 45: if you set `height`, you can remove `line-height: 60px;`

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  margin-left: 20px;
+  color: #000;
+}
+
+.nav-link:hover {
+  color: var(--hover-link-color);
+}
+
+.nav-list :first-child > .nav-link {
```
artemforsoff, line 54: [Increasing an Element Specificity](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules#typical-bem-mistakes-in-css)

artemforsoff:
Copy this [Checklist](https://github.com/mate-academy/layout_moyo-header#checklist) to the Pull Request description after links, and put - [x] before each point after you checked it.

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+.nav-link {
+  display: inline-block;
+  text-align: center;
+  text-wrap: nowrap;
+  text-transform: uppercase;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 60px;
```
artemforsoff, line 43: Don't just copy all styles from Figma. Think, which of them are relevant. Uneven sizes (e.g. line-height: 14.6px) are definitely useless.

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+.nav-link:hover {
+  color: var(--hover-link-color);
+}
+
+.nav-list :first-child > .nav-link {
+  margin-left: 0;
+}
+
+.moyo-logo-link {
+  height: 40px;
```
artemforsoff, line 59: It’s better to set the image size so that the link size is formed from the internal content

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+  font-weight: 500;
+  line-height: 60px;
+  text-decoration: none;
+  height: 60px;
+  margin-left: 20px;
+  color: #000;
+}
+
+.nav-link:hover {
+  color: var(--hover-link-color);
```
artemforsoff, line 51: to improve UX add a transition

artemforsoff (changes requested):
overall good work 🔥
there are some minor comments, please pay attention to them

src/style.css
artem8746: but in the checklist is written that I am not allowed to use * selector and I should set styles only for elements that require them, what should I do?

PR #4137, opened 2023-11-22
Develop

Zubkovskiy:
[DEMO LINK](https://Zubkovskiy.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://Zubkovskiy.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -2,13 +2,47 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="index.html" class="header-img">
+        <img src="images/logo.png" alt="logo">
+      </a>
+      <nav class="header-nav">
```
Esceype, line 18:
```suggestion
      </a>
      
      <nav class="header-nav">
```

Always add one empty line between children of the same parent. Only in one case you might not add a blank line if each child occupies only one line from the opening tag to the closing tag. Same for all the similar cases below.

src/style.css
```diff
@@ -1,3 +1,78 @@
+:root {
+  --main-color: #00ACDC;
+}
+
 body {
+  font-family: Roboto, sans-serif;
```
Esceype, line 6: Add one more font as fallback font

src/index.html
Esceype, line 18: Navigation should be a separate BEM block

src/index.html
```diff
@@ -2,13 +2,47 @@
...
+        <ul class="header-items">
```
Esceype, line 19:
```suggestion
        <ul class="nav__items">
```
After you create nav as a separate block them rename all the elements and use the double underscore for naming

src/style.css
```diff
@@ -1,3 +1,78 @@
...
+  gap: 20px;
+}
+
+.header-link {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: relative;
+  height: 60px;
+  font-style: normal;
```
Esceype, line 56: This is default value so you can delete it

Esceype (changes requested):
Great job!
Read the following comments

src/index.html
Zubkovskiy: I agree, but this error occurs:
34:1  ✖  Expected class selector ".nav__img" to be kebab-case    selector-class-pattern

SanyaBratashchuk (approved):

PR #4279, opened 2023-12-19
Develop

alionazherdetska:
Copy this Checklist to the Pull Request description after links, and put - [x] before each point after you checked it.

- [ ] Header height is set in 1 place (for the links)
- [ ] Content is vertically centered (for any header height)
- [ ] CSS is used to show all letters in Uppercase (don't type them in HTML)
- [ ] Logo is an image wrapped with a link
- [ ] CSS Variable is used for a blue color
- [ ] Pseudo-element is used for a blue line below the active link
- [ ] Code follows all the Code Style Rules ❗️

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,63 @@
 body {
   margin: 0;
 }
+
+html {
+  font-family: Roboto, sans-serif;
+  margin: 0;
+  font-size: 12px;
+  font-weight: 500;
+
+  --main-color: #00ACDC;
+}
+
+header {
+  display: flex;
+  width: 100%;
+  align-items: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,63 @@
...
+
+a {
+  text-decoration: none;
+  margin-right: 4rem;
+  color: black;
+  line-height: 24px;
+}
+
+a:hover {
+  color: var(--main-color);
```
etojeDenys, line 47: image

etojeDenys (changes requested):
add the links from readme.md to the pull request description

PR #4376, opened 2024-01-11
Develop

YarynaPuhach:
- [DEMO LINK](https://YarynaPuhach.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://YarynaPuhach.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,76 @@
+* {
+  box-sizing: border-box;
+}
+
+:root {
+  --main-color: #00acdc;
+  --background-color: #fff;
+  --text-color: #000;
+}
+
+html {
+  font-family: Roboto, sans-serif;
+}
+
 body {
   margin: 0;
+  font-style: normal;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  background-color: var(--background-color);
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1,3 +1,76 @@
...
+}
+
+.nav__item {
+  position: relative;
+  margin-right: 20px;
+}
+
+.nav__item:last-child {
+  margin-right: 0;
+}
```
asakevych, line 49: 1. instead of rewriting your margin only for last-child you can use `:not(:first-child)` and set margin only for not last elements
2. i think you don't need position relative here, it's not used

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+
+.nav__link {
+  display: block;
+  position: relative;
+  padding: 23px 0;
```
asakevych, line 54: it's bad practice to calculate every pixel: 
   for example if someday you decided to have your line-height bigger you should calculate paddings again 
I suggest it's better to set height to 60px and make the element flex to center  the content inside (your text)

asakevych (changes requested):
Good job! just some changes are needed :wink:

choeqq (approved):
LGTM ✅

PR #4440, opened 2024-01-25
Develop

Sywuxy:
- [DEMO LINK](https://Sywuxy.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Sywuxy.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,23 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <div class="container">
+      <header class="header">
+        <a href="#" class="header__logo"><img src="./images/logo.png" alt="logo"></a>
+
+        <nav class="header__nav">
+          <ul class="header__list">
+            <li class="header__list-item"><a href="#" class="is-active">Apple</a></li>
+            <li class="header__list-item"><a href="#">Samsung</a></li>
+            <li class="header__list-item"><a href="#">Smartphones</a></li>
+            <li class="header__list-item"><a href="#" data-qa="hover">Laptops & Computers</a></li>
+            <li class="header__list-item"><a href="#">Gadgets</a></li>
+            <li class="header__list-item"><a href="#">Tablets</a></li>
+            <li class="header__list-item"><a href="#">Photo</a></li>
+            <li class="header__list-item"><a href="#">Video</a></li>
+          </ul>
+        </nav>
+      </header>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,71 @@
...
+
+* {
+  margin: 0;
+}
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+
+  --header-height: 60px;
```
etojeDenys, line 11: move all vars to `::root` selector

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+  color: #000;
+  text-align: center;
+  text-transform: uppercase;
+  text-decoration: none;
+  font-weight: 500;
+  font-size: 12px;
+  padding: 0;
+}
+
+a:hover, a.is-active{
```
etojeDenys, line 57: better to add the class to `a` tags and use class selectors

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+  color: #00ACDC;
```
etojeDenys, line 58: move all repeatable colors to var

etojeDenys (changes requested):

src/style.css
Sywuxy: Denys, I cant use classes, because, my extensions say`s that in tag <a> can be only 2 attributes, as you know tag <a> has href and some task say to add 'data-qa="hover"' attribute, i can`t add one more class then, so I decided to style tags, wihout class selectors.

anastasiia-tilikina: yet we need to use class for CSS styles) if you have any problems feel free to ask them in chat) I believe we will find the issue

anastasiia-tilikina (changes requested):
One more thing to fix)

src/style.css
Sywuxy: Okay, I will ask then)

src/index.html
```diff
@@ -9,6 +9,25 @@
...
+
+      <nav class="header__nav">
+        <ul class="header__list">
+          <li class="header__list-item"><a href="#" class="is-active header__link">Apple</a></li>
+          <li class="header__list-item"><a href="#" class="header__link">Samsung</a></li>
+          <li class="header__list-item"><a href="#" class="header__link">Smartphones</a></li>
+          <li class="header__list-item">
+            <a href="#"
+            class="header__link"
+            data-qa="hover">Laptops & Computers</a>
```
etojeDenys, line 23: Pasted Graphic

etojeDenys (approved):
good job

PR #4443, opened 2024-01-26
add task solution

dionechik:
- [DEMO LINK](https://dionechik.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://dionechik.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -3,7 +3,7 @@
   padding: 0;
 }
 
-html {
+:root {
   --active-color: #00ACDC;
 }
 
@@ -50,10 +50,7 @@ header {
   text-decoration: none;
 }
 
-.nav-item :hover {
-  color: var(--active-color);
-}
-
+.nav-item :hover,
 .is-active {
   color: var(--active-color);
 }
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  text-decoration: none;
+}
+
+.nav-item :hover {
+  color: var(--active-color);
+}
+
+.is-active {
+  color: var(--active-color);
+}
```
etojeDenys, line 59:
```suggestion
.nav-item :hover,
.is-active {
  color: var(--active-color);
}
```

src/style.css
```diff
@@ -1,3 +1,69 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+html {
+  --active-color: #00ACDC;
```
etojeDenys, line 7: better to move all vars in `::root`

etojeDenys (approved):
lgtm

src/style.css
dionechik: Thanks for the feedback. Should I create a new commit on github or just fix the code myself?

PR #4461, opened 2024-01-29
add task solution

Davwidos:
- [DEMO LINK](https://davwidos.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://davwidos.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,30 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Moyo header</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+          crossorigin="anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a class="logo_link"><img src="images/logo.png" alt="logo"></a>
+      <nav class="nav">
+        <ul>
+          <li><a class="nav_link is-active">apple</a></li>
+          <li><a class="nav_link">samsung</a></li>
+          <li><a class="nav_link">smartphones</a></li>
+          <li>
+            <a class="nav_link"
+            data-qa="hover">laptops & computers</a>
+          </li>
+          <li><a class="nav_link">gadgets</a></li>
+          <li><a class="nav_link">tablets</a></li>
+          <li><a class="nav_link">photo</a></li>
+          <li><a class="nav_link">video</a></li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,83 @@
 body {
   margin: 0;
+  padding: 0;
+  font-family: Roboto, sans-serif;
+  box-sizing: border-box;
+
+  --blue-color: #00ACDC;
+}
+
+
+.header {
+  box-sizing: border-box;
+  width: 100vw;
+  padding: 0 50px;
+  background-color: #fff;
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+}
+
+.nav {
+  display: flex;
+  justify-content: flex-end;
+}
+
+.nav_link {
+  font-size: 12px;
+  font-weight: 500;
+  text-decoration: none;
+  color: #000;
+  text-transform: uppercase;
+
+  margin: 0 10px;
+  height: 60px;
+
+  display: flex;
+  flex-direction: column;
+  text-align: center;
+  justify-content: center;
+  align-items: stretch;
+}
+
+.nav_link:hover {
+  color: var(--blue-color);
+}
+
+
+.logo_link {
+  display: flex;
+  align-items: center;
+  flex-grow: 1;
+}
+
+.is-active {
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -6,9 +6,30 @@
...
+    <link rel="stylesheet" href="style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+          crossorigin="anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a class="logo_link"><img src="images/logo.png" alt="logo"></a>
```
asakevych, line 17: please start image tag from new line
```
<a ...>
  <img ..../>
<a/>
```

src/index.html
```diff
@@ -6,9 +6,30 @@
...
+      <nav class="nav">
+        <ul>
+          <li><a class="nav_link is-active">apple</a></li>
```
asakevych, line 20: start every tag from new line:
```suggestion
 <li>
   <a class="nav_link is-active">apple</a>
 </li>
```

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+  padding: 0;
+}
+
+ul li:first-of-type .nav_link {
+  margin-left: 0;
+}
+
+ul li:last-of-type .nav_link {
+  margin-right: 0;
 }
```
asakevych, line 83: please write all styles using classes, not tags

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+  height: 4px;
+  background-color: var(--blue-color);
+}
+
+.nav > ul {
+  list-style: none;
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
```
asakevych, line 75: you can use class to style element,here you can use `header__list` because due to BEM you have block header and list is an element inside

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+.is-active::before {
+  content: '';
+  height: 4px;
+}
+
+.is-active::after {
+  content: '';
+  height: 4px;
+  background-color: var(--blue-color);
+}
```
asakevych, line 68: do you need both before and after? it seems one of them is unused

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+.logo_link {
+  display: flex;
+  align-items: center;
+  flex-grow: 1;
+}
+
+.is-active {
+  color: var(--blue-color);
+  justify-content: space-between;
+}
```
asakevych, line 57: it's better to add styles more specific
```suggestion
.nav_link.is-active {
  color: var(--blue-color);
}
```
you don't need justify-content -> you doesn't use flex here and it's not needed (without `display: flex` this property do nothing :wink: )

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+  background-color: #fff;
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+}
+
+.nav {
+  display: flex;
+  justify-content: flex-end;
+}
```
asakevych, line 24: it seems those styles do nothing, you can remove them

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+
+
+.header {
+  box-sizing: border-box;
+  width: 100vw;
+  padding: 0 50px;
+  background-color: #fff;
+  display: flex;
+  justify-content: flex-end;
```
asakevych, line 17: flex-end position everything to the end, you need to use space-between
see: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+.nav {
+  display: flex;
+  justify-content: flex-end;
+}
+
+.nav_link {
+  font-size: 12px;
+  font-weight: 500;
+  text-decoration: none;
+  color: #000;
```
asakevych, line 30: you can create variables for all of the colors

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+  text-transform: uppercase;
+
+  margin: 0 10px;
+  height: 60px;
+
+  display: flex;
+  flex-direction: column;
```
asakevych, line 37: you don't need to change direction, its unnecessary here, if you remove it you won't see any changes

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+  text-align: center;
+  justify-content: center;
+  align-items: stretch;
```
asakevych, line 40: `align-items: stretch;` and `text-align: center;` are also not used

src/style.css
asakevych, line 68: you need to make it position absolute and pose this element 0 px from the bottom 
don't forget to add width to make element visible

src/index.html
asakevych, line 17: you can use block `header` and change class to `heder__link`, because you don't have any element to style inside, it's unnecessary to create new block

asakevych (changes requested):
So far so good!! just some changes are needed :rocket:

src/style.css
Davwidos: In class `is-active` i use `justify-content: space-between;` to make after appear at the bottom of link. Without before text will be placed at the top of link.

src/style.css
Davwidos: That means i should also add `display: flex` in `.nav_link.is-active`? `.nav_link` alreadey has `display: flex`

src/style.css
Davwidos: I changed position for element "after" so i no longer need element "before".

src/style.css
Davwidos: After the changes i dont't need justify-content here

choeqq (approved):
LGTM ✅

PR #4479, opened 2024-01-29
add task solution

DmitryKot96:
- [DEMO LINK](https://DmitryKot96.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://DmitryKot96.github.io/layout_moyo-header/report/html_report/)

- [ x] Header height is set in 1 place (for the links)
- [x ] Content is vertically centered (for any header height)
- [ x] CSS is used to show all letters in Uppercase (don't type them in HTML)
- [x ] Logo is an image wrapped with a link
- [x ] **CSS Variable** is used for a blue color
- [x ] Pseudo-element is used for a blue line below the active link
- [x ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,80 @@
+
+
+html {
+  margin: 0;
+  padding: 0;
+}
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+
+  --main-color: #000;
+  --blue-color: #00ACDC;
+}
+
+.header__content {
+  display: flex;
+  margin: 0;
+  justify-content:space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+.header__logoBox {
+  height: 40px;
+  width: 40px;
+}
+
+.header__nav{
+  margin: 0;
+}
+
+.header__navList {
+  margin: 0;
+  display: flex;
+  list-style-type: none;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  margin: 0;
+  padding: 0;
+}
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+
+  --main-color: #000;
+  --blue-color: #00ACDC;
```
BudnikOleksii, line 13: `--blue-color` - is actually the main color, the variable names should show their purpose

BudnikOleksii (changes requested):
Good work 👍 
Just a few changes are required, check the comments

src/style.css
```diff
@@ -1,3 +1,72 @@
...
+}
+
+.header__logoBox {
+  height: 40px;
+  width: 40px;
+}
+
+.header__nav{
+  margin: 0;
+}
```
stassokolov1, line 23: there is no need to set `margin: 0`, works without this style as expected (not critical)

stassokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #4485, opened 2024-01-30
add task solution

Bohdan-antonyk:
- [X ] Header height is set in 1 place (for the links)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,90 @@
+* {
+  box-sizing: border-box;
+
+  --main-color: #3498db;
+}
+
 body {
   margin: 0;
+  font-size: 12px;
+  font-weight: 500;
+  font-family: ' Roboto ', sans-serif;
+  color: #000;
+  background-color: #fff;
+}
+
+a {
+  color: #000;
+  text-decoration: none;
+  font-size: 12px;
+  display: block;
+  text-transform: uppercase;
+  font-weight: 500;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,90 @@
...
+  margin-left: 0;
+}
+
+nav ul li:last-child {
+  margin-right: 0;
+}
+
+/* nav ul li:nth-child(4) a {
+  color: #00ACDC;
+} */
```
anastasiiavorobiova, line 74:
```suggestion

```

src/style.css
```diff
@@ -1,3 +1,90 @@
...
+  justify-content: space-between;
+  padding: 0 50px;
+}
+
+.left-side__header {
+  width: 40px;
+  height: 40px;
+}
+
+nav > ul {
```
anastasiiavorobiova, line 51: Consider using classes for styling

anastasiiavorobiova (changes requested):
Great work! Keep doing: 
Screenshot 2024-01-30 at 15 05 54
The link for the image is bigger than the image 
Screenshot 2024-01-30 at 15 05 43
The height of navigation and navigation links should be the same with the header 
Screenshot 2024-01-30 at 15 05 32
Screenshot 2024-01-30 at 15 05 37
There are default styles here. The link height also should be 60 pixels

PR #4487, opened 2024-01-30
Develop

pawelkloc:
- [DEMO LINK](https://pawelkloc.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://pawelkloc.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,26 +8,52 @@
     <title>Moyo header</title>
     <link rel="preconnect" href="https://fonts.googleapis.com">
     <link rel="preconnect" href="https://fonts.gstatic.com">
-    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;700&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <header>
+    <header class="header">
       <div class="logo">
-        <a href="#">
-          <img src="./img/logo.png" alt="logo">
+        <a href="#home">
+          <img src="./images/logo.png" alt="Moyo logo">
         </a>
       </div>
       <nav class="nav">
-        <ul>
-          <li><a href="#" class="nav__link">apple</a></li>
-          <li><a href="#" class="nav__link">samsung</a></li>
-          <li><a href="#" class="nav__link">smartphones</a></li>
-          <li><a href="#" class="nav__link">laptops & computers</a></li>
-          <li><a href="#" class="nav__link">gadgets</a></li>
-          <li><a href="#" class="nav__link">tablets</a></li>
-          <li><a href="#" class="nav__link">photo</a></li>
-          <li><a href="#" class="nav__link">video</a></li>
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">samsung</a>
+          </li>
+
+          <li class="nav__item">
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -6,9 +6,56 @@
...
     <title>Moyo header</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
```
asakevych, line 16: 1.this div is redundant, you can use `<a>` to give some styles
2. also it's better to use class `header__logo`, because it's a part of `header` block
3. you should add width and height for `<a>` element to make it contain image with the size that you have on Figma

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+
+  --color: #00ACDC;
 }
+
+
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  padding: 0 50px;
```
asakevych, line 13: add `align-items: center` to make give logo appropriate spaces on top and bottom instead of using margin on the logo element

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+}
+
+.logo {
+  width: 40px;
+  height: 40px;
+  margin: 10px 0;
+
```
asakevych, line 20: unnecessary, better to use `align-items: center` in `header`

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+}
+
+.nav {
+  position: sticky;
+  top: 0;
```
asakevych, line 25: 1. unnecessary, if you want to make your header sticky please add position sticky to the whole header
2. your element is on top, you don't need `top: 0`, it does nothing here

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+ }
+
+.nav__list {
+  list-style-type: none;
```
asakevych, line 29: `list-style: none` is better solution here because it removes all list styles which can be different in different browsers and `list-style-type` removes only left dot

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+  display: flex;
+  flex-direction: row;
```
asakevych, line 31: it's a default value, you don't need to set it once again

src/style.css
```diff
@@ -1,3 +1,67 @@
...
+  margin-right: 20px;
+}
+
+.nav__link {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  text-transform: uppercase;
+  text-decoration: none;
+  align-items: center;
```
asakevych, line 46: it works only for display flex, but here you have inline, so it do nothing here

asakevych (changes requested):
Good job! :rocket:  just some small changes needed :wink:

src/style.css
```diff
@@ -1,3 +1,56 @@
...
+}
+
+.nav__link {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  text-transform: uppercase;
+  text-decoration: none;
+  color: #000;
+  line-height: 60px;
```
choeqq, line 39: I am not sure if `line-height` is proper way to change element's height 🤔

choeqq (changes requested):
Almost there, please make sure all tests, are passing 🚀

choeqq (changes requested):
Everything looks good, except for logo, and underline on apple, please try few different approaches to change the height of those elements

asakevych (approved):
Looks good to me, nice work :fire:

PR #4515, opened 2024-02-04
layout_moyo-header

damian2403:


damian2403:
[DEMO LINK](https://damian2403.github.io/layout_moyo-header/)

[TEST REPORT LINK](https://damian2403.github.io/layout_moyo-header/)

src/index.html
```diff
@@ -1,14 +1,52 @@
...
+      rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#" class="logo">
+        <img src="/moyo.png" 
+        class="imglogo"
+        alt="logo">
+      </a>
```
asakevych, line 20: 1. you should add classes due to bem, so here if you want to have a separate block `logo` so your image should have class `logo__img`, but to my mind there is no need to create a separate class, it could be an element of header block ->  `header__logo`
2. something is wrong with formatting, every attribute should be from new line  
```suggestion
      <a href="#" class="logo">
        <img 
          src="/moyo.png" 
          class="imglogo"
          alt="logo"
        >
      </a>
```

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 50px;
+  height: 60px;
+}
+
+.nav {
+  display: inline-block;
+}
```
asakevych, line 30: this style do nothing here, you can remove it and also remove class from `<nav>` element if it is not used

src/index.html
```diff
@@ -1,14 +1,52 @@
...
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="nav__link is-active link__1" href="#">Apple</a>
```
asakevych, line 24: 1. if you want to make a modification it should look this wa: `nav__link--1`
2. no need to add modification here, every link is the same

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
+.logo img {
+  width: 40px;
+  height: 40px;
+  vertical-align: bottom;
+}
```
asakevych, line 18: 1. please use classes to style elements
2. you don't need to style img, it has a proper styles, you can just style`<a>` element and use class `header__logo-link` or just `header__logo` to do it

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 50px;
+  height: 60px;
```
asakevych, line 25: it the checklist you have a requirement not to define height of the container, please make links height to define the height of header

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+}
+
+.nav {
```
asakevych, line 28: if you have block nav and block header please create a separate file for every bem block and import them to style.css

src/style.css
```diff
@@ -1,3 +1,80 @@
...
 }
+
+.nav__item {
+  margin-right: 20px;
+}
+
+.nav__item:last-child
+{
+  margin-right: 0;
+}
```
asakevych, line 46: instead you can use 
```suggestion
.nav__item:not(:last-child) {
  margin-right: 20px;
}
```

src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  width: 100%;
+  height: 4px;
+  position: absolute;
+  bottom: 0;
+  left: 0;
+}
+
+.link__1 {
+  color: var(--main-color);
+}
```
asakevych, line 80: we want to link, which is active to have such color, please use selector `.nav__link.is-active` to add this style and also you can do it in the line 59 not to duplikate styles, so it would look like: 
```suggestion
.nav__link:hover,
.nav__link.is-active {
  color: var(--main-color);
}
```

asakevych (changes requested):
Good job!! just some changes are required :wink: 
please also fix your tests raport link in pr, because it's the same as demo link

src/style.css
choeqq: @damian2403 one last change left: there is requirement not to hardcode the height on header, please fix if and you're good to go

choeqq (changes requested):
One little change needed @damian2403 all tests are passing 🚀

choeqq (approved):

PR #4538, opened 2024-02-08
moyo header task solve

mikestoic:
- [DEMO LINK](https://mikestoic.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://mikestoic.github.io/layout_moyo-header/report/html_report/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to fix the failed tests, you need to fix font family fo the links
image

The height should be 60px
image

PR #4554, opened 2024-02-14
added header with navigation

RostyslavKuzmych:


sarakhmen (approved):
Good job but let's follow BEM best practices next time ;) Pay attention to the block, elements and modifiers :) https://en.bem.info/methodology/quick-start/

PR #4609, opened 2024-02-24
add task solution

ArtemDemeniev:
- [DEMO LINK](https://ArtemDemeniev.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ArtemDemeniev.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,67 @@
-body {
+.body {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
   margin: 0;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1,3 +1,67 @@
...
+  height: 4px;
+  border-radius: 2px;
+  background-color: #00acdc;
+}
+
+.menu__link.is-active {
+  color: rgb(0, 172, 220);
+}
+
+.menu__link:hover {
```
polosanya, line 65: You can combine the selectors with same styles

polosanya (changes requested):
Almost done! Let's improve your code a bit)

PR #4612, opened 2024-02-24
Develop

JohnyJDX:

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,66 @@
+html {
+  margin: 0;
+}
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,70 @@
...
+}
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+}
+
+ul {
```
polosanya, line 16: Consider using class selectors instead of the tag

polosanya (changes requested):
Please don't forget to add demo and test links to the PR description.
Also, check yourself with the design. Make sure that header has the box-shadow property
image

DanilWeda (approved):
Great job! 
Thanks!

PR #4617, opened 2024-02-26
Develop

HigorPedroso:
[DEMO LINK](https://HigorPedroso.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://HigorPedroso.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,100 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="#">
+        <img
+          alt="Logo Moyo"
+          src="images/logo.png"
+          class="logo-header"
+        />
+      </a>
+      <nav class="menu">
+        <ul class="list-menu">
+          <li class="item-menu">
+            <a
+              href="#"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,12 +11,101 @@
...
+    <header class="header">
+      <a href="#">
+        <img
+          alt="Logo Moyo"
+          src="images/logo.png"
+          class="logo-header"
+        />
+      </a>
+      <nav class="menu">
+        <ul class="list-menu">
```
pedro-ruas, line 41: Wrong BEM naming.

I see you have declared all your elements in the format "element-block" when it should actually be "block__element".
You should check the declaration for all your classes from here onwards. Here is an example:
```
<nav class="header__menu menu">
    <ul class="menu__list">
        <li class="menu__item">
            <a class="menu__link">
                Link
            </a>
        </li>
    </ul>
</nav>
```

pedro-ruas (commented):
Some BEM naming issues to be fixed

src/style.css
pedro-ruas, line 1: After checking your BEM naming in the above file, remember to change this styles accordingly

pedro-ruas (changes requested):

src/style.css
```diff
@@ -1,3 +1,78 @@
...
+  color: var(--color-primary);
+  width: 90%;
+  bottom: 0;
+  border: 2px solid var(--color-primary);
+  border-radius: 10px;
+}
+
+.menu .menu__list .menu__item .menu__link:hover {
+  color: var(--color-primary);
 }
```
EdPirro, line 78: Define each element as a single CSS rule, applying only the necessary styles to each one!
Every element should be unique and not depend on its position on the page (that is why the naming convention follows the block__element--modifier rule)

Using
```css
.menu .menu__list {
/*...styles*/
} 
```
Is not only unnecessary but discouraged! It creates a highly coupled and needlessly convoluted styles;

Use:
```css
.menu {
/*...menu styles*/
} 

.menu__list {
/* ...menu__list styles */
}
```

EdPirro (changes requested):
Good effort, @HigorPedroso, you're almost there! Please carefully review the change I requested, it is very important! If you have any questions about it, please ask them in the qna chat!

pedro-ruas (approved):
Good job!

PR #4640, opened 2024-02-28
Moyo develop

WGoliasz:
- [DEMO LINK] (https://WGoliasz.github.io/layout_moyo-header/)
- [TEST REPORT LINK] (https://WGoliasz.github.io/layout_moyo-header/report/html_report/)
- [ x ] Header height is set in 1 place (for the links)
-  [x ] Content is vertically centered (for any header height)
- [ x ] CSS is used to show all letters in Uppercase (don't type them in HTML)
- [ x ] Logo is an image wrapped with a link
- [ x ] **CSS Variable** is used for a blue color
- [ x ] Pseudo-element is used for a blue line below the active link
- [ x ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,16 @@
+html {
+  margin: 0;
+}
+
 body {
   margin: 0;
+}
+
+.ul {
+  display: flex;
+  list-style: none;
+}
+
+li {
+  padding: 10px;
 }
--- a/src/style.css
+++ b/src/style.css
@@ -1,5 +1,5 @@
 :root {
-  --blue-color: #00ACDC;
+  --blue-color: #00acdc;
 }
 
 html {
@@ -31,7 +31,11 @@ body {
 }
 
 .nav_item {
-  margin: 12px;
+  margin: 11px 10px;
+}
+
+.nav_list > .nav_item:last-child {
+  margin-right: 0;
 }
 
 .nav_list {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  height: 40px;
+}
+
+.nav {
+  height: 60px;
+}
+
+.nav_item {
+  position: relative;
+  margin: 0 20px 0 0;
```
yevhenii-pyl, line 35: You are using `display:flex` for parent container. 
This makes usage of margin-right redundant, you can(and should) just use `gap` for parent. 

As a rule of thumb: wherever possible, parent should be responsible for positioning it's children within itself, NOT children with margins || paddings || etc.

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+}
+
+.nav_list > .nav_item:last-child {
```
yevhenii-pyl, line 38: With gap you can avoid this.

yevhenii-pyl (approved):
Good job 👍 
I would also probably add `cursor:pointer` for hover effect, but this is not the in the checklist, so we can just move forward.

WGoliasz:
I don't used display:gap becouse i thought i couldn't. Isn't that what it
was about? Details in the attachment.

czw., 29 lut 2024, 15:25 użytkownik Yevhenii Pylpychuk <
***@***.***> napisał:

PR #4641, opened 2024-02-28
add task solution

lmmaarriiaa:
- [DEMO LINK](https://lmmaarriiaa.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://lmmaarriiaa.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,8 +15,108 @@
       rel="stylesheet"
       href="./style.css"
     />
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500&display=swap"
+      rel="stylesheet"
+    />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a
+        class="logo"
+        href="index.html"
+      >
+        <img
+          class="logo__img"
+          src="/src/images/logo.png"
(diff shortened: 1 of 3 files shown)


DanilWeda (changes requested):
Good! 
image

Check the build version and demo link.
Thanks!

src/index.html
```diff
@@ -10,13 +10,116 @@
...
+        class="logo"
+        href="index.html"
+      >
+        <img
+          class="logo__img"
+          src="/src/images/logo.png"
+          alt="Moyo logo"
+        />
+      </a>
+      <nav class="nav">
```
undeletable, line 47: Multiline sibling elements in HTML should have an empty line between them.

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+  margin-left: 20px;
+}
+
+.nav__item:first-child {
+  margin: 0;
+}
+
+.nav__link {
+  display: inline-block;
+  height: 60px;
```
undeletable, line 48: You don't actually need both `height` and `line-height` in this case. One of them should be sufficient, just make sure that text is vertically centered.

undeletable (approved):
Well done! There's just a small discrepancy with the guideline, so I'm approving.

src/style.css
lmmaarriiaa: Hi! This is my way to make text centered and I have no idea how to do it another way in this case. Could you help?

undeletable: `line-height` sets the height of text line to 60px. So if there are no additional margins/paddings/borders etc., the parent element would take 60px height. This means that if you remove `height` style in this case, nothing would change.

lmmaarriiaa: Thanks!

PR #4650, opened 2024-03-01
add task solution

eduvinagre:
- [DEMO LINK](https://eduvinagre.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://eduvinagre.github.io/layout_moyo-header/report/html_report/)

- [X] Header height is set in 1 place (for the links)
- [X] Pseudo-element is used for a blue line below the active link


src/index.html
```diff
@@ -15,8 +15,108 @@
...
+        class="header__logo"
+      >
+        <img
+          src="./images/logo.png"
+          alt="logo"
+        />
+      </a>
+
+      <nav class="header__nav">
+        <ul class="header__nav__list">
```
jv-aquino, line 45: You are using the BEM nesting in a wrong way here, try using Mix Block or just the normal nesting conventions

https://en.bem.info/methodology/quick-start/#nesting-1

src/index.html
```diff
@@ -15,8 +15,108 @@
...
+          <li class="header__nav__item-1">
```
jv-aquino, line 46: You didn't need to add numbers for each of them, they have (except for the first one) the same attributes

src/style.css
```diff
@@ -1,3 +1,103 @@
...
+
+.header__nav__list {
+  padding: 0;
+  margin: 0;
+  display: flex;
+  list-style: none;
+}
+
+.header__nav__item-1 {
+  padding-right: 20px;
```
jv-aquino, line 37: With this use of classes your code is more prone to errors and difficult to change, so group them more

src/style.css
```diff
@@ -1,3 +1,103 @@
...
+  height: 60px;
+  padding: 0;
+  color: #000;
+  font-size: 12px;
+  font-weight: 500;
+  white-space: nowrap;
+  transition: color 0.3s;
+}
+
+.header__nav__item-4 .header__nav__link:hover {
```
jv-aquino, line 85: Every link should have this property

jv-aquino (changes requested):
You need to change some things in the implementations to follow the best practices, but your code organization is good

src/index.html
eduvinagre: Hi João, I'm having a hard time understanding this. What would be the right way? 
```
<nav class="header nav">
  <ul class="header nav list">
``` 
Would that be the right way of doindo this?

eduvinagre: nesting fixed

src/style.css
eduvinagre: fixed

src/index.html
eduvinagre: fixed

src/style.css
eduvinagre: fixed

PR #4675, opened 2024-03-05
add task solution

DaniilManakov:
x Header height is set in 1 place (for the links)
 x Content is vertically centered (for any header height)
 x CSS is used to show all letters in Uppercase (don't type them in HTML)
 x Logo is an image wrapped with a link
 x CSS Variable is used for a blue color
 x Pseudo-element is used for a blue line below the active link
 x Code follows all the [Code Style Rules ❗️](https://github.com/DaniilManakov/layout_moyo-header/blob/master/checklist.md)


l4st1m0za (changes requested):
Please, provide the DEMO and TEST links

PR #4689, opened 2024-03-06
Add moyo header

Follder:
Unfortunally, I created pull request to the master branch in my account, and then merged these branches.
May I create new pull request to the master branch ma from my master?


DanilWeda (changes requested):
Fine.
Add a hover effect to all navigation elements. (now only one navigation element has it )
You can create pull from any branch.
Thanks!

alexander-ignatow (changes requested):
1 As for your question - as I see it, yes, it should work

2. On your code - please add the hover effect, as @DanilWeda mentioned
Per task: `link color should be changed on :hover`

Follder:
> 1 As for your question - as I see it, yes, it should work
> 
> 2. On your code - please add the hover effect, as @DanilWeda mentioned
>    Per task: `link color should be changed on :hover`

But I added hover effect in the last commit, as
[data-qa]:hover,
.nav__link:hover {
  transition: .3s;
  color: var(--main-color)
}

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+}
+
+.nav__list {
+  font-size: 12px;
+  font-weight: 500;
+  display: flex;
+  text-transform: uppercase;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
```
tiserett, line 36: please, order your css properties in a [better way](https://9elements.com/css-rule-order/)

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  border-radius: 4px;
+  bottom: 0;
+  left: 0;
+  right: 0;
+}
+
+.nav__link:hover {
+  transition: 0.3s;
+  color: var(--main-color);
 }
```
tiserett, line 74: move it closer to the `.nav__link` selector

tiserett (approved):

PR #4704, opened 2024-03-07
Develop

Olek-91:
- [DEMO LINK](https://Olek-91.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Olek-91.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,101 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
+        <a href="#">
+          <img
+            src="images/logo.png"
+            alt="logo"
+          />
+        </a>
+      </div>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item item-1">
+            <a
+              class="is-active nav__link"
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,95 @@
-body {
+body,
```
undeletable, line 1: These styles are not needed. We only need to reset margin for `body`.
Also, according to CSS coding guideline, we should not use tags in CSS selectors, except `html` and `body`.

src/index.html
```diff
@@ -11,12 +11,103 @@
...
+      <div class="logo">
+        <a href="#">
+          <img
+            src="images/logo.png"
+            alt="logo"
+            class="logo__img"
+          />
+        </a>
+      </div>
+      <nav class="nav">
```
undeletable, line 42: According to coding guideline, please add empty lines between multi-line sibling HTML elements (`<div class="logo">`, `nav`, `<li>`)

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  margin-right: 50px;
+}
+
+.nav__item {
+  list-style: none;
+  font-size: 12px;
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+
+  /* margin: 0 10px; */
```
undeletable, line 44: Please remove the commented code. If it's not used, it might look confusing, especially when it becomes outdated.

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  margin-left: 50px;
+}
+
+/* .logo__img {
+  width: 100%;
+  height: 100%;
+} */
+
+.nav__link:hover {
+  color: #00acdc;
```
undeletable, line 76: Please define the blue color using CSS variable, according to the task requirement.

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+.nav__item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav__link {
+  display: inline-block;
+  text-decoration: none;
+  color: #000;
+  text-transform: uppercase;
+  height: 60px;
```
undeletable, line 60: Both `height` and `line-height` are not needed here. One of them would be fine, just make sure that the text is vertically centered.

undeletable (changes requested):
The solution looks fine. You only need to make it compliant to the guidelines and requirements.

PR #4708, opened 2024-03-08
add task solution

ArtemDemeniev:
- [DEMO LINK](https://ArtemDemeniev.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ArtemDemeniev.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,62 @@
-body {
+.body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 50px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,62 @@
-body {
+.body {
```
undeletable, line 1: No need to add `body` class. You can use `body` tag in CSS, as well as `html`. The other tags should not be used in selectors.

src/style.css
```diff
@@ -1,3 +1,62 @@
...
+
+.menu__link.is-active::after {
+  content: '';
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: 4px;
+  border-radius: 2px;
+  background-color: #00acdc;
```
undeletable, line 56: Blue color should be defined as CSS variable

mrkovaliv (approved):
Good work 👍

PR #4711, opened 2024-03-08
create moyo header

iAlfreX:
- [DEMO LINK](https://iAlfreX.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://iAlfreX.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/normalize.css
+++ b/src/normalize.css
@@ -1,0 +1,351 @@
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
+
+/* Document
+   ========================================================================== */
+
+/**
+ * 1. Correct the line height in all browsers.
+ * 2. Prevent adjustments of font size after orientation changes in iOS.
+ */
+
+html {
+  line-height: 1.15; /* 1 */
+  -webkit-text-size-adjust: 100%; /* 2 */
+}
+
+/* Sections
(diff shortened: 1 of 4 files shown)


src/normalize.css
```diff
@@ -0,0 +1,351 @@
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
```
undeletable, line 1: Using normalize.css is usually a good solution for large projects, but for this small task it's redundant. Most of tags mentioned in this file don't even exist in your HRML. And we only need to reset margin for `body`.

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  display: flex;
+  align-items: center;
+  height: 60px;
+  color: #000;
+  text-decoration: none;
+  text-transform: uppercase;
+  position: relative;
+}
+
+.nav-link:hover {
```
undeletable, line 51: Both this and the next selector has the same style. So you can combine this style and the next one into a single expression, listing selectors as comma-separated list.

undeletable (changes requested):
Well done. Just some changes needed in order to follow the guideline.

iAlfreX:
Done

undeletable (approved):
Well done

PR #4726, opened 2024-03-11
Develop

dyakym:
- [DEMO LINK](https://dyakym.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://dyakym.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,99 @@
+html {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+  font-size: 12px;
+}
+
+.roboto-medium {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+}
+
 body {
   margin: 0;
+}
+
+.header {
+  background-color: rgba(255, 255, 255, 1);
+  align-items: center;
+
+  /* padding: 0 50px; */
+  display: flex;
+  justify-content: space-between;
+  width: 100%;
+}
+
+.logo {
+  margin: 10px 0 10px 50px;
+
+  /* margin-left: 50px; */
+
+  /* padding: 10px 0; */
+  height: 40px;
+}
+
+.nav {
+  height: 60px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,99 @@
...
+  font-weight: 500;
+  font-style: normal;
+  font-size: 12px;
+}
+
+.roboto-medium {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+}
```
alexander-ignatow, line 12: Do you need this class?
If not - please remove it. Leaving unused code is a bad practice

src/style.css
```diff
@@ -1,3 +1,99 @@
...
+
 body {
   margin: 0;
 }
+
+.header {
+  background-color: rgba(255, 255, 255, 1);
+  align-items: center;
+
+  /* padding: 0 50px; */
```
alexander-ignatow, line 22: Leaving commented-out code like this is also a bad practice
Can you please remove it here and below

alexander-ignatow (changes requested):
Very good! 💪 

Just make a small cleanup to make the task look perfect!

src/style.css
```diff
@@ -1,3 +1,85 @@
...
+  padding: 0;
+}
+
+.nav__item {
+  list-style: none;
+  margin: 0 10px;
+  padding: 0;
+  justify-content: center;
+  display: flex;
+  align-items: center;
```
tiserett, line 46: please, order your css properties in a [better way](https://9elements.com/css-rule-order/)

src/style.css
```diff
@@ -1,3 +1,85 @@
...
+  margin-left: 0;
+}
+
+.nav__list :last-child {
+  margin-right: 0;
+}
+
+.nav__link:hover {
+  color: #00acdc;
+}
```
tiserett, line 69: I suggest moving those closer to their parent selectors

tiserett (approved):

PR #4728, opened 2024-03-11
add task solution

OTsiolkovskiy:
- [DEMO LINK](https://OTsiolkovskiy.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://OTsiolkovskiy.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,12 +11,110 @@
       content="ie=edge"
     />
     <title>Moyo header</title>
+
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500&display=swap"
+      rel="stylesheet"
+    />
+
     <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo">
+        <a
+          href="#"
+          class="logo_link"
+        >
+          <img
+            src="./images/logo.png"
+            alt="logo"
+            class="logo_moyo"
+          />
+        </a>
+      </div>
+      <div>
+        <nav class="nav">
+          <ul class="nav_list">
+            <li class="nav_item">
+              <a
+                href="#Apple"
+                class="nav_link is-active"
+              >
+                Apple
+              </a>
+            </li>
+            <li class="nav_item">
+              <a
+                href="#Samsung"
+                class="nav_link"
+              >
+                Samsung
+              </a>
+            </li>
+            <li class="nav_item">
+              <a
+                href="#Smartphones"
+                class="nav_link"
+              >
+                Smartphones
+              </a>
+            </li>
+            <li class="nav_item">
+              <a
+                href="#Laptops & Computers"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,12 +11,110 @@
...
+          href="#"
+          class="logo_link"
+        >
+          <img
+            src="./images/logo.png"
+            alt="logo"
+            class="logo_moyo"
+          />
+        </a>
+      </div>
```
mshepel, line 46: Please add empty lines between multiline sibling elements - here, in `<ul>` and in `<head>`.

src/index.html
```diff
@@ -11,12 +11,110 @@
...
+      <div>
```
undeletable, line 47: Please add empty lines between multiline sibling elements - here, in `<ul>` and in `<head>`.

src/style.css
```diff
@@ -1,3 +1,85 @@
...
+.nav_item:not(:first-child) {
+  margin-left: 20px;
+}
+
+.nav_item :hover {
+  color: var(--second-color);
+  text-decoration: solid 2px black;
+}
+
+.nav__link:hover {
```
undeletable, line 66: `.nav__link` is not used in your HTML, so this style can be removed

src/style.css
```diff
@@ -1,3 +1,85 @@
...
+  padding: 0;
+  display: flex;
+  list-style: none;
+}
+
+.nav_link {
+  text-decoration: none;
+  color: #000;
+  display: inline-block;
+  height: 60px;
```
undeletable, line 52: Both `height` and `line-height` are not needed here. With `line-height` you set the height of text to 60px, so if the parent item contains single line of text and has no vertical margins/paddings/borders, it will already be 60px high. So `height` is redundant here.

undeletable (changes requested):
The solution looks nice. It just needs some improvements, mostly related to guidelines. Please check the comments.

src/index.html
OTsiolkovskiy: fixed, added empty lines

src/index.html
```diff
@@ -11,12 +11,110 @@
...
+        <nav class="nav">
+          <ul class="nav_list">
```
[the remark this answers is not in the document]
OTsiolkovskiy: fixed, added BEM  syntax

src/style.css
OTsiolkovskiy: removed

src/style.css
OTsiolkovskiy: removed

src/style.css
```diff
@@ -1,3 +1,79 @@
...
+
+.header {
+  display: flex;
+  background-color: #fff;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.logo {
+  margin-left: 50px;
```
alexander-ignatow, line 21: > The block shouldn't influence its environment, meaning you shouldn't set the external geometry (margin) or positioning for the block.

One of the techniques to avoid such problem is using mixes
https://en.bem.info/methodology/css/#external-geometry-and-positioning

alexander-ignatow (approved):
For now, let's go with it, however, there are two issues, that are to be fixed in the next tasks:

1. Make sure to submit only the files, related to the task. For instance `dist` folder definitely should not be here
2. See my comment for BEM blocks below

Otherwise, the task looks good 👍

PR #4729, opened 2024-03-11
Develop

luyx0:
- [DEMO LINK](https://luyx0.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://luyx0.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,19 +15,6 @@
       rel="stylesheet"
       href="./style.css"
     />
-    <link
-      rel="preconnect"
-      href="https://fonts.googleapis.com"
-    />
-    <link
-      rel="preconnect"
-      href="https://fonts.gstatic.com"
-      crossorigin="anonymous"
-    />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
-      rel="stylesheet"
-    />
   </head>
   <body>
     <header>
@@ -38,7 +25,7 @@
         />
       </div>
       <nav class="nav">
-        <ul class="list">
+        <ul class="nav__list">
           <li class="link">
             <a
               href="#"
--- a/src/index.html
+++ b/src/index.html
@@ -15,8 +15,101 @@
       rel="stylesheet"
       href="./style.css"
     />
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,91 @@
+@import 'https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&displ0ay=swap';
```
undeletable, line 1: The same font is added twice in the different places: here and in `<head>` of HTML. We don't need this, so please remove font addition from one of the places.

src/style.css
```diff
@@ -1,3 +1,91 @@
...
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 14px;
+  display: flex;
+}
+
+.element::after {
+  content: '';
+  position: absolute;
+  background-color: #00acdc;
```
undeletable, line 56: Please define blue color as CSS variable

src/style.css
```diff
@@ -1,3 +1,91 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+header {
+  background-color: #fff;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 60px;
```
undeletable, line 19: According to the requirements, height for header should not be set. It should only be set for link items, so that header's height would be calculated automatically.

src/style.css
```diff
@@ -1,3 +1,91 @@
...
+  border-radius: 8px;
+  height: 4px;
+  width: 0;
+  left: 0;
+  bottom: 0;
+}
+
+.element:active {
+  color: #00acdc;
+  box-shadow: 0 4px 0 #00acdc;
```
undeletable, line 66: When I click on menu item, this style adds the blue line under it. This is not needed.

src/index.html
```diff
@@ -15,8 +15,101 @@
...
-    <h1>Moyo header</h1>
+    <header>
+      <div class="logo">
+        <img
+          src="images/logo.png"
+          alt="logo"
+        />
+      </div>
+      <nav class="nav">
+        <ul class="list">
```
undeletable, line 41: Please use BEM syntax for classes (e.g., `nav__list` instead of `list`, where `nav` refers to the parent block).

undeletable (changes requested):
The solution looks fine, but needs improvements in order to meet guidelines and requirements. Please check the comments to the code.

luyx0:
- [DEMO LINK](https://luyx0.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://luyx0.github.io/layout_moyo-header/report/html_report/)

PR #4732, opened 2024-03-11
add task solution

SlawikH:
- [DEMO LINK](https://SlawikH.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://SlawikH.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,106 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
+    <link
+      rel="stylesheet"
+      href="style.css"
+    />
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,12 +11,107 @@
...
+      <a
+        href="#"
+        class="header_logo"
+      >
+        <img
+          src="/src/images/logo.png"
+          alt="MOYO logo"
+        />
+      </a>
+      <nav>
```
undeletable, line 46: Please add empty lines between multiline siblings across all the HTML file.

src/index.html
```diff
@@ -11,12 +11,107 @@
...
+        <ul class="nav_list">
```
undeletable, line 47: In terms of BEM, `nav` is a block and `list`, `item` are elements. Please fix two things for compliance with this syntax:
- Define `nav` block for parent item
- According to BEM syntax, block and element names should be separated with double underscore in the class name - and in your code they're separated with single underscore. Please correct that.

src/style.css
```diff
@@ -1,3 +1,71 @@
-body {
+:root {
+  --blue-color: #00acdc;
+}
+
+* {
```
undeletable, line 5: Please don't use `*` selectors. This might have a negative impact to performance, since browser engine has to apply such styles to every single item in HTML. That's why it's prohibited in our code style guide.
Reset of margin and padding would be sufficient for `body` only.

undeletable (changes requested):
The solution looks fine. Just please improve your code to meet the requirements and code style guide. See the comments.

src/index.html
SlawikH: for some reason ,when i run npm test, my code structure is getting broken
it might be pritier

tiserett (approved):

PR #4800, opened 2024-03-26
add task solution

Stasandrushchak:
- [DEMO LINK](https://Stasandrushchak.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Stasandrushchak.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -10,13 +10,118 @@
...
+        <ul class="u-list">
+          <li class="nav_list_item">
+            <a
+              href="./index.html"
+              class="nav_link is-active"
+            >
+              apple
+              <img
+                src="./images/Rectangle.png"
+                alt="is_active"
```
polosanya, line 62: It is not correct to add underline as an image. What if another link becomes active? It should have underline too.
Consider implementing it with after/before pseudoelement
https://www.w3schools.com/css/css_pseudo_elements.asp

src/style.css
```diff
@@ -1,3 +1,112 @@
...
+  padding: 0;
+}
+
+.roboto-regular {
+  font-family: Roboto, sans-serif;
+  font-weight: 400;
+  font-style: normal;
+}
+
+.roboto-bold {
```
polosanya, line 21: There are no such selectors(classes) as .roboto-regular, .roboto-bold etc..

src/style.css
```diff
@@ -1,3 +1,112 @@
...
+  text-align: left;
+}
+
+.is-active::after {
+  position: relative;
+  background-image: url(./images/Rectangle.png);
+  bottom: 4;
+}
+
+.video {
```
polosanya, line 110: Don't hardcode the name of the class according to the link name. What if it changes?

polosanya (changes requested):
Nice start! Let's try to pass all tests
Pay attention to the design. Header should have box-shadow property
image

src/index.html
Stasandrushchak: Thank you very much for advice, and link)

src/style.css
Stasandrushchak: Can inspect my new code?

PR #4815, opened 2024-04-01
add header first v

SofiSolf:
- [DEMO LINK](https://SofiSolf.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://SofiSolf.github.io/layout_moyo-header/report/html_report/)

p.s. : я ще не виправила помилки в коді до псевдоелемента під першою лінкою)


src/index.html
```diff
@@ -16,7 +16,94 @@
...
+          <li class="nav__item">
+            <a
+              class="nav__link"
+              href="index.html"
+            >
+              Photo
+            </a>
+          </li>
+
+          <li class="nav__item nav__item-8">
```
nazarmatsevych, line 97:
```suggestion
          <li class="nav__item nav__item">
```

Don't add extra classes, you can use :last-child rule to remove margin-right from the last element

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+}
+
+.nav__item-1.is-active::after {
+  content: ' ';
+  height: 4px;
+  width: 37px;
+  background-color: var(--blue-color);
+  position: absolute;
+  bottom: 1px;
+  left: 722px;
```
nazarmatsevych, line 54:
```suggestion
  bottom: 0;
  left: 0;
```

and you need to add posititon relative to `.is-active` class

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+}
+
+.nav__link.nav__link-4 {
+  color: var(--blue-color);
+}
+
+.nav__item-1.is-active::after {
+  content: ' ';
+  height: 4px;
+  width: 37px;
```
nazarmatsevych, line 50:
```suggestion
  width: 100%;
```

src/index.html
```diff
@@ -16,7 +16,94 @@
       href="./style.css"
     />
   </head>
+
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo-wrapper">
```
nazarmatsevych, line 22: I think you don't need a wrapper here

nazarmatsevych (changes requested):
Good start but need to make some improvements

1. You should import Roboto font in `head `tag
image


If you have any questions - feel free to ask them in fe_chat

src/style.css
SofiSolf: Hi! Thanks for your recommendations. I made some changes. Although I used Roboto font for header, it doesn't look similar to original. Also, I added position relative to .is-active class instead of position absolute, but the pseudo-element doesn't show up :( I would highly appreciate your reply and some advice!

VitaliyBondarenko1982 (approved):
Looks like demo links was cached and not works properly (can help remove .parcel-cache folder and deploy again)
But see that tests passed, so approve.

SofiSolf:
> Looks like demo links was cached and not works properly (can help remove .parcel-cache folder and deploy again) But see that tests passed, so approve.

Thanks) I've updated the links

PR #4826, opened 2024-04-04
creat header

RuslanHnatyshyn:
- [DEMO LINK](https://RuslanHnatyshyn.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://RuslanHnatyshyn.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,66 @@
 body {
   margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+:root {
+  --black: #000;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,66 @@
...
+  display: block;
+}
+
+.nav-link:hover {
+  color: var(--blue);
+}
+
+.is-active {
+  position: relative;
+  color: var(--blue);
```
polosanya, line 54: You can combine the selectors with same properties

polosanya (changes requested):
Almost done!
The header should have box-shadow property. Pay attention to the design
image

polosanya (approved):
Well done!

PR #4830, opened 2024-04-08
add task solution

Stasandrushchak:
- [DEMO LINK](https://Stasandrushchak.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Stasandrushchak.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -24,9 +24,7 @@ body {
 .logo {
   height: 40px;
   width: 40px;
-  margin-top: 10px;
-  margin-bottom: 10px;
-  margin-left: 50px;
+  margin: 10px 0 10px 50px;
 }
 
 .logo_link {
@@ -36,7 +34,7 @@ body {
 }
 
 .nav {
-  margin-right: 50px;
+  margin: 0 50px 0 0;
   padding: 0;
 }
 
@@ -53,7 +51,7 @@ body {
 }
 
 .nav_link {
-  margin: 0 10px;
+  margin: 0 0 0 20px;
   line-height: 60px;
   height: 60px;
   text-decoration: none;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1,3 +1,87 @@
...
+  justify-content: space-between;
+  box-shadow: 0 2px 4px 0 #0000000d;
+}
+
+.logo {
+  height: 40px;
+  width: 40px;
+  margin-top: 10px;
+  margin-bottom: 10px;
+  margin-left: 50px;
```
polosanya, line 29: You can use shorthand property instead
https://developer.mozilla.org/en-US/docs/Web/CSS/margin#syntax

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+  font-family: Roboto, Arial, Helvetica, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+}
+
+.nav_link:hover {
+  color: var(--blue);
+}
+
+a.is-active {
```
polosanya, line 72: There is no need to combine the selectors. Just use class selector instead

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+}
+
+a.is-active::after {
+  position: absolute;
+  content: url(./images/Rectangle.png);
+  bottom: -26px;
+  left: 0;
+}
+
+.last-link {
```
polosanya, line 85: Consider using pseudoclass to select the last child

src/style.css
```diff
@@ -1,3 +1,87 @@
...
+
+a.is-active {
+  color: var(--blue);
+  margin-left: 0;
+  position: relative;
+}
+
+a.is-active::after {
+  position: absolute;
+  content: url(./images/Rectangle.png);
```
polosanya, line 80: Don't add underline as a photo. What if you need to add it for another link?

polosanya (changes requested):
Nice! Let's improve your solution a bit)

src/style.css
Stasandrushchak: I added some changes due to Your recommendations

polosanya (approved):
Well done! 👍

PR #4860, opened 2024-04-18
add task solution

dkolinska:


lerastarynets (changes requested):
Pls deploy test reports as now they're unavailable via the link

PR #4899, opened 2024-04-29
add header

MariCodec:
- [DEMO LINK](https://MariCodec.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://MariCodec.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,102 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    />
+
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="header_logo">
+        <a href="#">
+          <img
+            src="./images/logo.png"
+            alt="image_logo"
+          />
+        </a>
+      </div>
+      <nav class="header_nav">
+        <ul class="nav_list">
+          <li class="nav_list_link">
+            <a
+              href="#"
+              class="link is-active"
+            >
+              Apple
+            </a>
+          </li>
+          <li class="nav_list_link">
+            <a
+              href="#"
+              class="link"
+            >
+              Samsung
+            </a>
+          </li>
+          <li class="nav_list_link">
+            <a
+              href="#"
+              class="link"
+            >
+              Smartphones
+            </a>
+          </li>
+          <li class="nav_list_link">
+            <a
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -1,3 +1,58 @@
...
+}
+
+html {
+  --text-color: #00acdc;
+}
+
+.header {
+  font-family: Roboto, sans-serif;
+  width: 100%;
+  height: 60px;
```
Oleksandr-kopaevich, line 13: Please see the checklist this point:
the height should be set for nav links (not the header), take it from the design.
Remove height from here, it'll be calculated by height of header's children

Oleksandr-kopaevich:
From requirements
image
- There should not be margins before the first and after the last list items

src/index.html
```diff
@@ -11,12 +11,103 @@
...
     />
     <title>Moyo header</title>
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
```
Oleksandr-kopaevich, line 21: why do you need these two links?

Oleksandr-kopaevich (changes requested):
Hey, please fix some comments to finish the checklist and make tests to pass (id you don't know how the come to any QnA to ask)

src/style.css
```diff
@@ -1,3 +1,69 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  --text-color: #00acdc;
+
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
```
roman-mirzoian, line 10: This is the default value, you don't need to specify it.
```suggestion
```

src/style.css
```diff
@@ -1,3 +1,69 @@
+* {
```
roman-mirzoian, line 1: Don't use * selector (it impacts performance), better use ``html`` or ``body`` tag.
image

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding-inline: 50px;
+}
+
+/*
+.header_logo {
+
+} */
```
roman-mirzoian, line 29: Do not leave unnecessary comments.
```suggestion
```

src/style.css
```diff
@@ -1,3 +1,69 @@
...
+
+.nav__list {
+  display: flex;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
```
roman-mirzoian, line 35: It is better to set these properties to the entire page in the ``body`` or ``html`` tag.
To avoid duplicating code.

roman-mirzoian (changes requested):
Well done, but pay attention to this comments.

dinaminahi (approved):
Well done 👍 

Just run `npm run deploy` so the [link](https://maricodec.github.io/layout_moyo-header/) will show your current changes

PR #4900, opened 2024-04-29
add solution

YaroslavMatviienko:
[DEMO LINK](https://YaroslavMatviienko.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://YaroslavMatviienko.github.io/layout_moyo-header/report/html_report/)

❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.

- [+ ] Header height is set in 1 place (for the links)
- [+] Content is vertically centered (for any header height)
- [+] CSS is used to show all letters in Uppercase (don't type them in HTML)
- [+] Logo is an image wrapped with a link
- [+] **CSS Variable** is used for a blue color
- [+] Pseudo-element is used for a blue line below the active link
- [+] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,75 @@
-body {
+:root {
+  --blue: #0070f3;
+}
+
+* {
   margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: Roboto, sans-serif, bold;
+}
+
+header {
+  width: 100%;
+  background-color: #fff;
+  display: flex;
+  justify-content: space-between;
+  flex-direction: column;
+}
+
+img {
+  display: flex;
+  align-items: flex-start;
+  padding: 32px;
+  margin-right: auto;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,75 @@
...
+* {
   margin: 0;
+  padding: 0;
+}
+
+body {
+  font-family: Roboto, sans-serif, bold;
+}
+
+header {
```
roman-mirzoian, line 14: Don't use tag names for styling (except html and body)
Fix it in all code.

image

roman-mirzoian (changes requested):
Good work, but there is still a lot of work to do.
Deployment does not work and tests fail. This needs to be fixed.

image
image

PR #4904, opened 2024-04-29
Develop

OleksiiKnihin:
- [DEMO LINK](https://OleksiiKnihin.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://OleksiiKnihin.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,104 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
+        <a
+          class="logo-header"
+          href="#"
+        >
+          <img
+            src="./images/logo.png"
+            alt="Moyo logo"
+          />
+        </a>
+        <nav class="navigation">
+          <ul class="navigation-list">
+            <li class="navigation-item">
+              <a
+                class="navigation-link is-active"
+                href="#"
+              >
+                Apple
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,68 @@
+/* CSS variables */
```
roman-mirzoian, line 1:
```suggestion
```

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  padding: 0 50px;
+  justify-content: space-between;
+  align-items: center;
+  max-width: 1200px;
+  min-width: 780px;
+}
+
+.navigation-list {
+  display: flex;
   margin: 0;
```
roman-mirzoian, line 27: Don't duplicate code.

src/index.html
```diff
@@ -11,12 +11,105 @@
...
+    />
     <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
```
roman-mirzoian, line 33: Excess wrapping. you can set styles for the header.

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+}
+
+.container {
+  display: flex;
+  margin: 0 auto;
+  padding: 0 50px;
+  justify-content: space-between;
+  align-items: center;
+  max-width: 1200px;
+  min-width: 780px;
```
roman-mirzoian, line 22: Not hardcoded or using static dimensions.
It is better to set more flexible sizes, for example (max-width: 100%)

roman-mirzoian (changes requested):
Nice job, but needs a bit of work.

src/style.css
OleksiiKnihin: I didn’t understand what this was about. If I remove this margin, then show the standard from the browser from ul. No matter how I tried it, I can’t do without it))
Other comments have been corrected!

roman-mirzoian (changes requested):
Well done the code looks good, but you forgot to update the deploy demo.
The current demo does not match your code.
https://gyazo.com/c5ff805d1c5c86c7d1fa854e724d19d5

roman-mirzoian (approved):
Well done!

PR #4905, opened 2024-04-29
add Moyo header solution

FaiHamid:
- [DEMO LINK](https://FaiHamid.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://FaiHamid.github.io/layout_moyo-header/report/html_report/)

[x] Code follows all the [Code Style Rules ❗️](https://github.com/FaiHamid/layout_moyo-header/blob/master/checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,103 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -11,12 +11,104 @@
...
+      <a
+        href="https://www.moyo.ua/ua/"
+        class="logo"
+      >
+        <img
+          src="images/logo.png"
+          alt="logo_header"
+        />
+      </a>
+
```
dinaminahi, line 42:
```suggestion
```

src/style.css
```diff
@@ -1,3 +1,79 @@
...
+
+.is-active::after {
+  content: '';
+  position: absolute;
+  top: 56px;
+  display: block;
+  background-color: var(--hover-color);
+  border-radius: 8px;
+  height: 4px;
+  width: 37px;
```
dinaminahi, line 78: To not add specific px width, you can set `position: relative` for `.is-active` element and add` width: 100% `for `after` element

Also better to add `bottom: 0` instead of setting top

dinaminahi (approved):
Well done ⭐

PR #4906, opened 2024-04-29
add moyo header

StsKushnir:
- [DEMO LINK](https://StsKushnir.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://StsKushnir.github.io/layout_moyo-header/report/html_report/)

[x] Code follows all the Code Style Rules

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,102 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a
+        href="https://www.moyo.ua/ua/"
+        class="logo__link"
+      >
+        <img
+          src="./images/logo.png"
+          alt="logo_site"
+        />
+      </a>
+      <nav class="header__menu">
+        <ul class="header__list">
+          <li class="header__item">
+            <a
+              href="https://www.moyo.ua/ua/portal_brand/apple_1/"
+              class="header__link is-active"
+            >
+              Apple
+            </a>
+          </li>
+          <li class="header__item">
+            <a
+              href="https://www.moyo.ua/ua/telecommunication/smart/samsung/"
+              class="header__link"
+            >
+              Samsung
+            </a>
+          </li>
+          <li class="header__item">
+            <a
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,101 @@
...
+
+p,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  margin: 0;
+}
```
dinaminahi, line 28: Do you have these elements on your page? If no, remove these styles as they are useless

src/style.css
```diff
@@ -1,3 +1,101 @@
...
+
+a {
+  text-decoration: none;
+}
```
dinaminahi, line 32: Let`s use classes instead of tags as selectors

src/style.css
```diff
@@ -1,3 +1,101 @@
...
+.header__item:nth-child(-n + 7) {
+  margin-right: 20px;
+}
+
+.header__link {
+  text-transform: uppercase;
+  text-decoration: none;
+  color: #000;
+  height: 100%;
+  height: 60px;
```
dinaminahi, line 76: Let's keep one height as it completely overlaps the upper height
```suggestion
  height: 60px;
```

src/style.css
```diff
@@ -1,3 +1,101 @@
...
+  line-height: 60px;
```
dinaminahi, line 77: If you want to center vertically text with it, the better way will be to use flex
```suggestion
display: flex;
align-items: center;
```

src/style.css
```diff
@@ -1,3 +1,101 @@
...
+
+.is-active::after {
+  position: absolute;
+  content: '';
+  display: block;
+  background-color: var(--select-color);
+  height: 4px;
+  width: 37px;
+  border-radius: 8px;
+  top: 56px;
```
dinaminahi, line 96: To not add specific px width, you can set position: relative for .is-active element and addwidth: 100%for after element

Also better to add bottom: 0 instead of setting top

src/style.css
```diff
@@ -1,3 +1,101 @@
...
+  font-size: 12px;
+
+  --select-color: #00acdc;
+}
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
```
dinaminahi, line 12: And let's place * selector at top of the file

dinaminahi (changes requested):
Good job! 👍 

Let's do few fixes to make it even better 😉

src/index.html
```diff
@@ -11,12 +11,103 @@
...
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a
+        href="https://www.moyo.ua/ua/"
+        class="logo__link"
+      >
+        <img
+          src="./images/logo.png"
+          alt="logo_site"
```
roman-mirzoian, line 39: For logos, use the format: "[Organization name] logo." 
Example:
image

src/style.css
```diff
@@ -1,3 +1,88 @@
...
```
roman-mirzoian, line 5: Don't use * selector (it impacts performance)
It is better to use the ``body`` or ``html`` tag.

image

roman-mirzoian (approved):
Well done, but pay attention to this comments.

PR #4913, opened 2024-04-30
add task solution

priemssonia:
[DEMO LINK](https://priemssonia.github.io/layout_moyo-header/)
 [TEST REPORT LINK](https://priemssonia.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,74 @@
 body {
+  font-family: Roboto, 'Helvetica Neue', sans-serif, bold;
+  font-size: 12px;
+  font-weight: 500;
   margin: 0;
+  padding: 0;
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  box-sizing: border-box;
+  background-size: 100% auto;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,86 @@
+* {
+  box-sizing: border-box;
+}
+
```
dinaminahi, line 4: Better not to use * selector, you can set this on those elements you need
image

src/style.css
```diff
@@ -1,3 +1,86 @@
...
+  font-size: 12px;
+  font-weight: 500;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  box-sizing: border-box;
+  padding-left: 50px;
+  height: 60px;
```
dinaminahi, line 18: Let's set the height to the link itself, not the header

image

dinaminahi (changes requested):
Somehow I can't see your site, and can't check the tests, the links don't work
Have you run deploy?

image

dinaminahi (approved):
Well done ⭐

PR #4948, opened 2024-05-13
Develop

MateuszCieplak:
- [DEMO LINK](https://MateuszCieplak.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://MateuszCieplak.github.io/layout_moyo-header/report/html_report/)

- [X] Header height is set in 1 place (for the links)
- [X] Pseudo-element is used for a blue line below the active link

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,6 +17,6 @@
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <h1>Moyo header!!!</h1>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -15,8 +15,90 @@
       rel="stylesheet"
       href="./style.css"
     />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap"
+    />
   </head>
   <body>
-    <h1>Moyo header!!!</h1>
+    <header class="container">
+      <a href="index.html">
+        <img
+          src="./images/logo.png"
+          alt="logo"
+          class="logo"
+        />
+      </a>
+      <nav class="nav">
+        <ul class="nav_list">
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link is_active"
+            >
+              apple
+            </a>
+          </li>
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link"
+            >
+              samsung
+            </a>
+          </li>
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link"
+            >
+              smartphones
+            </a>
+          </li>
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link blue"
+              data-qa="hover"
+            >
+              laptops & computers
+            </a>
+          </li>
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link"
+            >
+              gadgets
+            </a>
+          </li>
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link"
+            >
+              tablets
+            </a>
+          </li>
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link"
+            >
+              photos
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -15,8 +15,89 @@
...
+      <nav class="nav">
+        <ul class="nav_list">
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link is-active"
+            >
+              Apple
+            </a>
+          </li>
```
yevhenii-pyl, line 41: We need to add some 'breathing-space' here. 
image

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  display: flex;
+  align-items: center;
+}
+
+.logo {
+  display: flex;
+  align-items: center;
+}
+
+/* .nav {
```
yevhenii-pyl, line 27: Let's clean it up

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  flex-direction: row;
+  justify-content: stretch;
+} */
+
+.nav_list {
+  display: flex;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  margin-right: -20px;
```
yevhenii-pyl, line 39: Nooo way :) 

Consider using something like this for nav item:

```
...:last-child {
  margin-right: 0;
}
```

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+}
+
+.nav_item {
+  display: inline-block;
```
yevhenii-pyl, line 43: Not sure that you need it.

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+  color: rgba(0, 0, 0, 1);
+  margin: 0;
+  height: 60px;
+  line-height: 60px;
+  position: relative;
+}
+
+.is-active {
+  text-decoration: none;
+  color: rgba(0, 172, 220, 1);
```
yevhenii-pyl, line 60: Why don't we create a variables for colors? 

:root {
  --col-text: ...: ;
  --col-ative: ....;
}

src/index.html
```diff
@@ -15,8 +15,89 @@
...
       href="./style.css"
     />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap"
+      rel="stylesheet"
+    />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="container">
```
yevhenii-pyl, line 24: Let's give it a 'header' class. 
Or, if you want to stick to that '-container' conception, let's call it 'header-container'.

src/style.css
```diff
@@ -1,3 +1,82 @@
+html {
```
yevhenii-pyl, line 1: Let's avoid setting that rule at the -html level. It will do as a rule for header, for example.

yevhenii-pyl (changes requested):
Good job! But let's make it even better with just a few adjustments 💪

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+
+.nav_list {
+  display: flex;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+}
+
+.nav_item {
+  margin: 0 10px;
```
yevhenii-pyl, line 38: I asked you to use `:last-child ` pseudo-class to order to reset margin-right to 0 for the last element, but I've never recommended to set margin left & right for the element. It's actually a mistake here. 

Since it's a flex container, ideally, you should use 'gap' property to position elements. However, in this task margin is required instead of gap, so we have to use margins. 
But let's not make it worse: by using margin left & right you making it harder to check the spacing between elements and has to use two pseudo-classes (for the fists and last child) instead of just one. 

So let's use margin-right: 20 in combination with last-child pseudo-class. No need in adding extra complexity here.

yevhenii-pyl (changes requested):
Just two more steps, you are really close to it, but let's make it 100% right 💪

yevhenii-pyl (approved):
You got it! 🥇

PR #4965, opened 2024-05-20
add task solution

aedram:
- [DEMO LINK](https://aedram.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://aedram.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1,3 +1,63 @@
-body {
-  margin: 0;
+@import 'https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap';
+
+html {
+  --link-active-color: #00acdc;
```
gitlherme, line 4: A common practice is to define the variables in `:root` selector, because you can reference it globally. That is not a error, but maybe you can change if you want to.

gitlherme (approved):
Hello, @aedram!

Great job! Congratulations 🔥

Just a point about the code. It is not a problem, just an advice to the next projects, okay?

PR #4992, opened 2024-05-30
All tests done

Kimmladshii:
[x] - Header height is set in 1 place (for the links)
[x] - Content is vertically centered (for any header height)
[x] - CSS is used to show all letters in Uppercase (don't type them in HTML)
[x] - Logo is an image wrapped with a link
[x] - CSS Variable is used for a blue color
[x] - Pseudo-element is used for a blue line below the active link
[x] - Code follows all the [Code Style Rules ❗️](https://github.com/Kimmladshii/layout_moyo-header/blob/master/checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,105 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="container">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -11,12 +11,106 @@
...
+          <ul class="nav__list">
+            <li class="nav__item">
+              <a
+                href="#"
+                class="nav__link is-active"
+              >
+                apple
+              </a>
+            </li>
+            <li class="nav__item">
```
BogdanMaliuta, line 54: split sibling elements using empty line

src/style.css
```diff
@@ -1,3 +1,73 @@
...
+ul,
+li {
   margin: 0;
+  padding: 0;
+
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+
+  --color: #00acdc;
```
BogdanMaliuta, line 11: are you realy need this rules for body,ul,li?

BogdanMaliuta (approved):
Good job.Dont forget to split sibling elements by empty  line, also take a look on styles, when you resetting your css

src/style.css
Kimmladshii: I think this one is a better option compared to zeroing out padding / margin for ul and li to avoid unnecessary code. Otherwise without zeroing them the test will fail.

src/index.html
Kimmladshii: unfortunately that's how prettier worked and broke my usual indented code. I will be glad to get help with its customization.

PR #5007, opened 2024-06-04
Develop

kathetem:
- [DEMO LINK](https://kathetem.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://kathetem.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,7 +1,8 @@
 body {
   margin: 0;
   padding: 0;
-  box-sizing: border-box;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
 }
 
 .header {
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix the all failed tests 
image

Remove the default color for this link, the blue color must be only for hover effects 
image

Remove the padding left for the ul tag 
image

Remove the left margin for the first link 
image

Remove the right margin for the last link 
image

src/style.css
```diff
@@ -1,3 +1,86 @@
 body {
   margin: 0;
+  padding: 0;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
 }
+
+.header {
+  height: 60px;
```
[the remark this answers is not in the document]
kathetem: I have a problem with setting height to links, it is not working...

src/style.css
```diff
@@ -1,3 +1,91 @@
...
+
+.nav_item:not(:last-child) {
+  margin-right: 20px;
+}
+
+.nav_link {
+  display: block;
+  color: #000;
+  position: relative;
+  font-weight: bold;
```
vadiimvooo, line 48: here you have to use font-weight: 500

vadiimvooo (changes requested):
Good job, but you have to pass all tests. Feel free to ask for help in fe_chat.
1. This margins should be removed
image
2. Just add margins only for 1 direction instead of 2
image

src/style.css
```diff
@@ -1,3 +1,91 @@
...
+.is-active::after {
+  display: block;
+  content: ' ';
+  position: absolute;
+  left: 0;
+  right: 0;
+  width: 100%;
+  height: 4px;
+  bottom: 0;
+  background-color: #00acdc;
```
anastasiiavorobiova, line 78:
```suggestion
  background-color: var(--is-active-color);
```

src/style.css
```diff
@@ -1,3 +1,91 @@
...
+
+.nav_list {
+  list-style: none;
+  display: flex;
+  color: #000;
+  padding-left: 0;
+}
+
+a:hover {
+  color: #00acdc;
```
anastasiiavorobiova, line 90:
```suggestion
  color: var(--is-active-color);
```

anastasiiavorobiova (approved):
Great work!

PR #5015, opened 2024-06-06
add task solution

joselangoni:
- [DEMO LINK](https://joselangoni.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://joselangoni.github.io/layout_moyo-header/report/html_report/)

- [x] Code follows all the [Code Style Rules ❗️

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,79 @@
+@import 'https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap';
+
+html {
+  font-family: Roboto, Arial, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+}
+
+:root {
+  --font-color: #000;
+  --hover-color: #00acdc;
+}
+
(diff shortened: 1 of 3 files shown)


alysonvilela:
Hello @joselangoni, seems like your code is a bit different from what is expectect on the test runner.
From what I checked, could be simple like ajust the .nav_item margin from **right** to **left**.

Consider always use the test report to validate!

src/style.css
```diff
@@ -1,3 +1,79 @@
...
+.nav_list {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  list-style: none;
+}
+
+.nav_item {
+  text-decoration: none;
+  margin-right: 20px;
```
alysonvilela, line 48: Try usign margin-left instead

alysonvilela (changes requested):

src/style.css
joselangoni: Hi @alysonvilela now it passed in the tests.
I made the changes you told me and also removed the margin-left from the first-child.
Thanks!

gitlherme (approved):
Hello, @joselangoni!

Very good job! 🎉

PR #5029, opened 2024-06-11
add task solution

Yurii-Yaremko:
- [DEMO LINK](https://Yurii-Yaremko.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Yurii-Yaremko.github.io/layout_moyo-header/report/html_report/)

- [x]  Header height is set in 1 place (for the links)
- [x]  Content is vertically centered (for any header height)
- [x]  CSS is used to show all letters in Uppercase (don't type them in HTML)
- [x]  Logo is an image wrapped with a link
- [x]  CSS Variable is used for a blue color
- [x]  Pseudo-element is used for a blue line below the active link
- [x]  Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,8 +15,75 @@
       rel="stylesheet"
       href="./style.css"
     />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+    />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header is-active1">
+      <a
+        href="#"
+        class="header-link"
+      >
+        <img
+          src="images/logo.png"
+          alt="Logo"
+          class="logo"
+        />
+      </a>
+      <nav class="nav">
+        <a
+          href="#section-1"
+          class="nav__link nav__link1 is-active"
+          data-qa="hover"
+        >
+          APPLE
+        </a>
+        <a
+          href="#section-2"
+          class="nav__link nav__link2"
+        >
+          SAMSUNG
+        </a>
+        <a
+          href="#section-3"
+          class="nav__link nav__link3"
+        >
+          SMARTPHONE
+        </a>
+        <a
+          href="#section-4"
+          class="nav__link nav__link4"
+          data-qa="hover"
+        >
+          LAPTOPS & COMPUTERS
+        </a>
+        <a
+          href="#section-5"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -15,8 +15,75 @@
...
+      >
+        <img
+          src="images/logo.png"
+          alt="Logo"
+          class="logo"
+        />
+      </a>
+      <nav class="nav">
+        <a
+          href="#section-1"
```
DanilWeda, line 37: You can use name of section as href

src/index.html
```diff
@@ -15,8 +15,75 @@
...
+          class="nav__link nav__link1 is-active"
```
DanilWeda, line 38: Also you don't need number of link. you can use `nav__link` and `is-active` for all requiremets which be in this task.

src/index.html
```diff
@@ -15,8 +15,75 @@
...
+          data-qa="hover"
+        >
+          APPLE
```
DanilWeda, line 41: You can use class property (text-transform) and paste text in lowercase in html

src/style.css
```diff
@@ -1,3 +1,84 @@
...
+  color: #007bff;
+}
+
+.is-active1:hover {
+  border: 2px solid rgba(0, 172, 220, 1);
+}
+
+.nav__link[data-qa='hover']:hover::after {
+  content: '';
+  position: absolute;
```
DanilWeda, line 58: You don't need position absolute in this task, try use transform property

src/style.css
```diff
@@ -1,3 +1,84 @@
...
+  color: #007bff;
+}
+
+.nav__link4 {
+  color: rgba(0, 172, 220, 1);
+}
+
+.nav__link1 {
+  color: rgba(0, 172, 220, 1);
+}
```
DanilWeda, line 72: You can use pseudo selectors for first/last element

src/style.css
```diff
@@ -1,3 +1,84 @@
...
+
+.is-active::after {
+  content: '';
+  position: absolute;
```
DanilWeda, line 76: You don't need absolute ( see comment below)

DanilWeda (changes requested):
Good. 
Check some comments for improve your solution! 
Thanks you and have a nice day!

PR #5112, opened 2024-07-04
add task solution

Mykyta-snacj:
- [DEMO LINK](https://Mykyta-snacj.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Mykyta-snacj.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,8 +15,107 @@
       rel="stylesheet"
       href="./style.css"
     />
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500&display=swap"
+      rel="stylesheet"
+    />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="header__logo">
+        <a href="#header">
+          <img
+            src="images/logo.png"
+            alt="logo"
+            class="header__logo__img"
+          />
+        </a>
+      </div>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li>
+            <a
+              href="#apple"
+              class="nav__link is-active"
+            >
+              Apple
+            </a>
+          </li>
+
+          <li>
+            <a
+              href="#samsung"
+              class="nav__link"
+            >
+              Samsung
+            </a>
+          </li>
+
+          <li>
+            <a
+              href="#smartphones"
+              class="nav__link"
+            >
+              Smartphones
+            </a>
+          </li>
+
+          <li>
+            <a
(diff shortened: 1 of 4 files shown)


BogdanMaliuta (changes requested):
Can`t see demo and tests, fix it please, also remove regions from your css 
image

src/index.html
```diff
@@ -15,8 +15,107 @@
...
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <div class="header__logo">
+        <a href="#header">
+          <img
+            src="images/logo.png"
+            alt="logo"
+            class="header__logo__img"
```
Zibi95, line 38: According to BEM, an element cannot contain another element directly. Only blocks have elements. Therefore, `header__logo-img` is the correct class name, where `header` is the block and `logo-img `is the element.
```suggestion
            class="header__logo-img"
```

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+
+.nav__list {
+  display: flex;
+  padding: 0;
+  margin: 0;
+
+  list-style: none;
+}
+
+.nav__list > li:first-child a {
```
Zibi95, line 30: Simplify selectors to adhere to BEM principles and avoid increasing specificity unnecessarily. We can simplify this by using the fact that nav__link have also is-active class.

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+
+.header {
+  box-sizing: border-box;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  min-width: 1024px;
+  max-width: 1200px;
+  margin: 0 auto;
```
Zibi95, line 15: > the header should stretch the full page width (don't use horizontal margin).

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+}
+
+.header__logo__img {
```
Zibi95, line 18: Utilize SCSS nesting with & symbol as it can make the stylesheet more concise and maintainable.

src/style.css
```diff
@@ -1,3 +1,68 @@
 body {
   margin: 0;
+
+  --selected-color: #00acdc;
```
Zibi95, line 4: It is a common best practice in modern CSS to define global variables in `:root` to ensure that they are universally accessible and clearly signify their global nature.

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  text-decoration: none;
+  color: #000;
+}
+
+.nav__link:hover {
```
Zibi95, line 54: Utilzie scss nesting.

Zibi95 (changes requested):
Good job! There few things to fix. Follow suggestions left in comments below.

PR #5120, opened 2024-07-09
Develop

W3zzie:
[DEMO LINK](https://W3zzie.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://W3zzie.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,7 +1,9 @@
 html {
-  font-family: Roboto, sans-serif;
-  font-weight: 500;
-  font-style: normal;
+  font-family: Roboto, Arial, Helvetica, sans-serif;
+}
+
+:root {
+  --active-collor: #00acdc;
 }
 
 body {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,86 @@
...
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+}
+
+ul :first-child a {
+  margin: 0 10px 0 0;
```
Zibi95, line 39: If you want to apply a margin on one side, you can use the `margin-left` or `margin-right` property.

src/style.css
```diff
@@ -1,3 +1,86 @@
...
+  margin: 0 10px;
+  color: #000;
+  text-decoration: none;
+  font-size: 12px;
+  font-weight: 500;
+  text-transform: uppercase;
+}
+
+.logo {
+  position: absolute;
```
Zibi95, line 61: This one doesn't feel right for me here in this case.

Zibi95 (approved):
Well done, it's almost perfect. You broke one rule from the style guide:

Don't use tag names for styling (except html and body)

I approve this task, but try to follow my suggestions for practice and use it as a future reference.

PR #5136, opened 2024-07-16
Develop

Irina0401:
## Checklist
- [x] Code follows all the Code Style Rules ❗️


lerastarynets (changes requested):
Pls deploy test reports as now they're unavailable via the link
Pls try to pass all tests, if your tests pass locally, but you encounter a problem in server tests, pls attach a screen of all local tests passing

SanyaBratashchuk (approved):

PR #5146, opened 2024-07-18
fix header height

israelmarek:
- [DEMO LINK](https://israelmarek.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://israelmarek.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,13 +10,98 @@
       http-equiv="X-UA-Compatible"
       content="ie=edge"
     />
-    <title>Moyo header</title>
+    <title>Moyo header w Logo</title>
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500&display=swap"
+      rel="stylesheet"
+    />
     <link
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,13 +10,98 @@
...
+        href="#"
+        class="logo"
+      >
+        <img
+          class="img"
+          src="../src/images/logo.png"
+          alt="Logo"
+        />
+      </a>
+      <nav>
```
pedro-ruas, line 35: You are not using BEM in any class in this code.
Here is an example of what nav should be named:
```
<nav class="header__navbar navbar">
    <ul class="navbar__links">
        <li class="navbar__option">
            <a class="navbar__link">
 ....
```

pedro-ruas (changes requested):
Need to add BEM classes to the code

src/index.html
israelmarek: Thanks, just fixed the code and sent a review request.

PR #5153, opened 2024-07-20
Develop

nex1994:
- [DEMO LINK](https://nex1994.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://nex1994.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,8 +15,28 @@
       rel="stylesheet"
       href="./style.css"
     />
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="me">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet">
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a href="index.html" class="logo__link"><img src="/src/images/logo.png" alt="moyo logo"></a>
+      <nav class="nav">
+        <a href="#apple" class="nav__link ">apple</a>
+        <a href="#samsung" class="nav__link">samsung</a>
+        <a href="#smartphones" class="nav__link">smartphones</a>
+        <a href="#laptop&computers" class="nav__link data-qa=hover">laptops&computers</a>
+        <a href="#gadgets" class="nav__link">gadgets</a>
+        <a href="#tablets" class="nav__link">tablets</a>
+        <a href="#photo" class="nav__link">photo</a>
+        <a href="#video" class="nav__link">video</a>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -15,27 +15,81 @@
       rel="stylesheet"
       href="./style.css"
     />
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -15,8 +15,82 @@
...
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a
+        href="index.html"
+        class="logo__link"
```
Zibi95, line 36: For proper BEM this should be `header__logo`. Elements should be nested inside parent (block)

src/index.html
```diff
@@ -15,8 +15,82 @@
...
+        <a
+          href="#tablets"
+          class="nav__link"
+        >
+          tablets
+        </a>
+        <a
+          href="#photo"
+          class="nav__link"
+        >
```
Zibi95, line 84: This looks like list of elements. For this purpose there are semantic tags as `<ul>` or `<ol>` and `<li>` tags. This is also requrement for this task.

src/style.css
```diff
@@ -1,3 +1,70 @@
 body {
   margin: 0;
+  font-family: Roboto, Arial, Helvetica, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  text-transform: uppercase;
+
+  --link-hover: #00acdc;
```
Zibi95, line 8: give color variables more meanigful names like simply calling it for example `--color-blue`

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+}
+
+.is-active {
+  color: var(--link-hover);
+  display: flex;
+  flex-direction: column;
+  place-content: center center;
+}
+
+.is-active p {
```
Zibi95, line 66: forget about using element type selectors. Excluding of course html and body element selectors.

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+  text-align: center;
+}
+
+.nav__link:hover {
+  color: var(--link-hover);
+}
+
+.nav__link:not(:first-child) {
+  margin-left: 20px;
+}
```
Zibi95, line 46: Try too use scss nesting for these selectors.

src/style.css
```diff
@@ -1,3 +1,70 @@
...
+  height: 40px;
+  width: 40px;
+}
+
+.nav {
+  display: flex;
+  justify-content: space-between;
+  margin-right: 50px;
+  height: 60px;
+  align-items: stretch;
```
Zibi95, line 29: I believe that stretch value is default one value for `align-items` so you don;t have to explicitly declare it.

Zibi95 (changes requested):
Hey. Good job. I left some comments there is one thing to do to meet all the requirements for this task.

src/index.html
```diff
@@ -15,8 +15,100 @@
...
+          src="/src/images/logo.png"
+          alt="moyo logo"
+        />
+      </a>
+      <nav class="nav">
+        <ul class="list">
+          <li class="list__element">
+            <a
+              href="#apple"
+              class="nav__link is-active"
```
Zibi95, line 48: You broke the rule of BEM here. You can't use elements of one block inside other block.

Zibi95 (approved):
Good job you met the requirements. I left some comments about mistakes that you made with BEM convention. But its not the part of this task to write good BEM code so just read them for future reference.

PR #5154, opened 2024-07-20
Moyo Header solution

azazeka:
- [DEMO LINK](https://azazeka.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://azazeka.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,63 @@
+:root {
+  --main-color: #00acdc;
+}
+
 body {
   margin: 0;
+  font-family: 'Roboto', sans-serif;
+}
+
+.header-container {
+  height: 60px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+.logo-container {
+  width: 40px;
+  height: 40px;
+}
+
+.list-item {
+  list-style-type: none;
+}
+
+.navigation-container .list-item {
+  margin-right: 20px;
+  display: inline;
+}
+
+.list-item {
+  text-transform: uppercase;
+  text-decoration: none;
+}
+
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
good job, but demo links don't work, try to fix that or feel free to ask for some help in the chat

src/style.css
```diff
@@ -1,3 +1,79 @@
...
+  font-size: 12px;
+  max-width: 1200px;
+}
+
+.header-container {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding-left: 50px;
```
etojeDenys, line 17:
```suggestion
  padding: 0 50px;
```

src/style.css
```diff
@@ -1,3 +1,79 @@
...
+
+.logo {
+  display: block;
+  width: 100%;
+  height: auto;
+}
+
+.navigation {
+  margin-right: 50px;
+}
```
etojeDenys, line 33: and then remove this property at all

src/style.css
```diff
@@ -1,3 +1,79 @@
...
+  list-style-type: none;
+  padding: 0;
+  margin: 0;
+}
+
+.list-item {
+  text-transform: uppercase;
+  margin-right: 20px;
+  font-weight: 500;
+  line-height: normal;
```
etojeDenys, line 47: it's default value

etojeDenys (approved):

src/style.css
azazeka: So what should I do with it? Remove it at all, or move to body?

PR #5214, opened 2024-08-06
add task solution

AnyaRatusnyak:
x Header height is set in 1 place (for the links)
x Content is vertically centered (for any header height)
x CSS is used to show all letters in Uppercase (don't type them in HTML)
x Logo is an image wrapped with a link
x CSS Variable is used for a blue color
x Pseudo-element is used for a blue line below the active link
x Code follows all the [Code Style Rules ❗️]


SarakhmanFTMentor (approved):
let's speed up and move to the next topics. Our first priority - to have your own project written :)

PR #5242, opened 2024-08-12
Develop Moyo

VolodymyrV7:
х Header height is set in 1 place (for the links)
х Content is vertically centered (for any header height)
х CSS is used to show all letters in Uppercase (don't type them in HTML)
х Logo is an image wrapped with a link
х CSS Variable is used for a blue color
х Pseudo-element is used for a blue line below the active link
х Code follows all the Code Style Rules ❗️

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,95 @@
-body {
+html {
+  --blue-color: #00acdc;
+}
+
+/* Скидання стандартних відступів браузера */
+* {
   margin: 0;
+  box-sizing: border-box;
+}
+
+/* Встановлюємо шрифт Roboto */
+body {
+  font-family: Roboto, sans-serif;
+}
+
+/* Стилі для заголовка */
+header {
+  width: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  background: #fff;
+  padding: 0 50px;
+}
+
+/* Логотип як посилання */
+.logo-link img {
+  width: 40px;
+  height: 40px;
+}
+
+/* Основні стилі для навігації */
+nav {
+  display: flex;
+  align-items: center;
+}
+
+/* Список навігаційних посилань */
+.style-link {
+  list-style: none;
+  display: flex;
+}
+
+.style-link li {
+  margin-right: 20px;
+}
+
+.style-link li:first-child {
+  margin-left: 0;
+}
+
+.style-link li:last-child {
+  margin-right: 0;
+}
+
+.style-link li:not(:last-child) {
+  margin-right: 20px;
+}
+
+/* Стилі для навігаційних посилань */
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,93 @@
...
+  box-sizing: border-box;
+}
+
+/* Встановлюємо шрифт Roboto */
+body {
+  font-family: Roboto, sans-serif;
+}
+
+/* Стилі для заголовка */
+header {
```
SanyaBratashchuk, line 18: use css classes for styling

src/style.css
```diff
@@ -1,3 +1,93 @@
...
+
+/* Стилі для навігаційних посилань при наведенні */
+.nav-link:hover {
+  color: var(--blue-color);
+}
+
+/* Стилі для четвертого посилання (Laptops & Computers) */
+.nav-link[data-qa="hover"]:hover {
+  color: var(--blue-color);
 }
```
SanyaBratashchuk, line 93:
```suggestion
```
Already applied on line 87

SanyaBratashchuk (changes requested):
demo link is not working. run `npm run deploy` once again

src/style.css
VolodymyrV7: fixed

src/style.css
VolodymyrV7: +

VolodymyrV7 (commented):
fixed

src/style.css
```diff
@@ -1,3 +1,94 @@
...
+  box-sizing: border-box;
+}
+
+/* Встановлюємо шрифт Roboto */
+body {
+  font-family: Roboto, sans-serif;
+}
+
+/* Стилі для заголовка */
+header {
```
etojeDenys, line 18: better to fix it
image

you need to give some class to the header tag and then use that class here

src/style.css
etojeDenys: image

src/style.css
```diff
@@ -1,3 +1,94 @@
...
+  width: 100%;
+  position: absolute;
+  background-color: var(--blue-color);
+  height: 4px;
+  border-radius: 8px;
+  left: 0;
+  bottom: 0;
+}
+
+/* Стилі для навігаційних посилань при наведенні */
```
etojeDenys, line 86: and pls remove all comments
Try to write obvious code that explains all we need without comments)

etojeDenys (changes requested):
feel free to ask for some help in the chat about that problem with push

src/style.css
VolodymyrV7: ok

etojeDenys (approved):
good job

PR #5258, opened 2024-08-13
add solution

KateBerd:
[DEMO LINK](https://KateBerd.github.io/layout_moyo-header/)
[TEST REPORT LINK](https://KateBerd.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,93 @@
+html {
+  margin: 0;
+  padding: 0;
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+}
+
 body {
   margin: 0;
+  padding: 0;
+}
+
+.header {
+  width: 100%;
+  margin: 0;
+  padding: 0;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.li {
+  height: 100%;
+}
+
+.nav {
+  display: flex;
+  height: 60px;
+  margin: 0;
+  padding: 0;
+  margin-right: 50px;
+}
+
+.nav_link {
+  position: relative;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1,3 +1,93 @@
...
+  margin: 0;
+}
+
+.logo_img {
+  display: block;
+  width: 40px;
+  height: 40px;
+  left: 50px;
+  margin: 10px 50px;
+}
```
VolodymyrKirichenko, line 61: it's redundant because you're using `flex` with the `space-between` property in your header, as well as zero padding.
Try using horizontal padding in your header

```suggestion
.logo_img {
  width: 40px;
  height: 40px;
}
```

src/style.css
```diff
@@ -1,3 +1,93 @@
...
+
 body {
   margin: 0;
+  padding: 0;
+}
+
+.header {
+  width: 100%;
+  margin: 0;
```
VolodymyrKirichenko, line 15: why u need `margin: 0` in this place?

src/style.css
```diff
@@ -1,3 +1,93 @@
...
+  height: 100%;
+}
+
+.nav {
+  display: flex;
+  height: 60px;
+  margin: 0;
+  padding: 0;
+  margin-right: 50px;
+}
```
VolodymyrKirichenko, line 32: do u need it? why?
```suggestion
.nav {
  height: 60px;
  margin-right: 50px;
}
```

VolodymyrKirichenko (changes requested):
Good job, but let's make your code better!)

PR #5302, opened 2024-08-20
changes

Joao-Campos06:
- [DEMO LINK](https://Joao-Campos06.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Joao-Campos06.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  left: 333px;
+  margin: 0;
+  opacity: 0;
+ }
+
+ :hover {
+  color: blue;
+ }
+
+::after is-active {
```
pedro-ruas, line 69: I believe this isn't right and what you were looking for is:
```
.is-active::after {
    position: absolute;
}
```

pedro-ruas (changes requested):
João, I believe you got what needs to be done, your code looks overall well done, but I cannot test it to check if styling met the requirements.
Seems that you forgot top run the script `npm run deploy` so the DEMO and TEST links work, fix this so I can test it again, please 😉

pedro-ruas (changes requested):
João, please use the figma designs as a guide. Your header is not occupying the whole screen width, the height of it differs from the design, the spacings also.
Try to make it look as close as possible to the designs by solving these issues

src/index.html
```diff
@@ -10,13 +10,115 @@
...
+          <li class="nav__item5">
+            <a
+              href="#tablets"
+              class="nav__link"
+            >
+              Tablets
+            </a>
+          </li>
+
+          <li class="nav__item6">
```
jv-aquino, line 102: they are mostly the same item so they should have the same class

jv-aquino (changes requested):
you didn't need to use position absolute for every object - use display flex + gap instead! this design will be very fixed and crash after the screen goes a little smaller

Only use the absolute positioning for the after
By the way, I suggest you to use a hover animation here
image

the project is approved though - just correct it later please and move on - because the corrections were really late, we're so sorry

PR #5358, opened 2024-09-14
add task solution

AndersonAPinto:
- [DEMO LINK](https://AndersonAPinto.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://AndersonAPinto.github.io/layout_moyo-header/report/html_report/)

❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.


pedro-ruas (changes requested):
Anderson, your implementation has failed every test due to different styling. Double check it before approval, and ask for help in the chat if needed.
Seems that your links don't have the full height of the header, and they should. There is color mismatching in active item and the after is not well positioned

AndersonAPinto (commented):
execute changes.

AndersonAPinto (commented):
Update solutions

src/index.html
```diff
@@ -17,6 +17,37 @@
     />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <a
+        href="#"
+        class="logoImagem"
```
pedro-ruas, line 23: This class does not follow BEM convention

src/index.html
```diff
@@ -17,6 +17,37 @@
...
+      >
+        <img
+          src="./images/logo.png"
+          alt="logo Moyo"
+        />
+      </a>
+      <nav class="navigation">
+        <ul class="listUn">
```
pedro-ruas, line 31: Neither does this one

src/index.html
```diff
@@ -17,6 +17,37 @@
...
+          <li class="listLinks">
```
pedro-ruas, line 32: And this

src/style.css
```diff
@@ -1,3 +1,78 @@
...
+
+.navigation .listLinks:not(:last-child) {
+  margin-right: 20px;
+}
+
+.navigation .listLinks:first-child {
+  color: var(--color-ref);
+}
+
+.navigation a {
```
pedro-ruas, line 50: Instead of using specification here, you could add the styling to a class unique to the link. Would be a better approach

pedro-ruas (changes requested):
Anderson, despite the fact you have finished the task and it is working, you did not follow every item on the checklist. You should use BEM naming conventions in this task, and basically all tasks to be honest, and I don't see it here.
Also, when you update your code to use BEM, you should also update the styling to avoid using unnecessary specification in styles, when a single class would fix it

pedro-ruas (approved):
Now we are talking! Perfect job!

PR #5368, opened 2024-09-16
add task solution

bodik-s4:
- [DEMO LINK](https://bodik-s4.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://bodik-s4.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,71 @@
+:root {
+  --active-color: #00acdc;
+}
+
 body {
   margin: 0;
+}
+
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  text-transform: uppercase;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,73 @@
...
+  color: #000;
+}
+
+.nav__item {
+  cursor: default;
+  justify-content: center;
+}
+
+.is-active {
+  margin-left: 0;
```
etojeDenys, line 48:
```suggestion
```
it's bad practice to reset the margin in this class, because we can make the third link active, but we don't need in the third link margin-left:0

etojeDenys (changes requested):
feel free to ask for some help in the chat

lerastarynets (approved):
You've got this, keep pushing forward!

PR #5428, opened 2024-09-27
Develop

GuilhermeSant:
- [DEMO LINK](https://GuilhermeSant.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://GuilhermeSant.github.io/layout_moyo-header/report/html_report/)

- [x ] Header height is set in 1 place (for the links)
- [ x] Content is vertically centered (for any header height)
- [ x] CSS is used to show all letters in Uppercase (don't type them in HTML)
- [ x] Logo is an image wrapped with a link
- [ x] **CSS Variable** is used for a blue color
- [ x] Pseudo-element is used for a blue line below the active link
- [ x] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,8 +15,39 @@
       rel="stylesheet"
       href="./style.css"
     />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css?family=Roboto"
+    />
   </head>
   <body>
-    <h1>Moyo header</h1>
+    <header class="header">
+      <nav class="container">
+        <a
+          href="https://th.bing.com/th/id/OIP.axjC06-qxFXElNPCbM1MPAAAAA?rs=1&pid=ImgDetMain"
+          target="_blank"
+        >
+          <img
+            src="./images/logo.png"
+            alt="logo-moyo"
+          />
+        </a>
+        <div class="nav-bar">
+          <div class="nav-link is-active">Apple</div>
+          <div class="nav-link is-active">Samsung</div>
+          <div class="nav-link is-active">smartphones</div>
+          <div
+            class="nav-link is-active"
+            data-qa="hover"
+          >
+            laptops & computers
+          </div>
+          <div class="nav-link is-active">gadgets</div>
+          <div class="nav-link is-active">Tablets</div>
+          <div class="nav-link is-active">Photo</div>
+          <div class="nav-link is-active">Video</div>
+        </div>
+      </nav>
+    </header>
   </body>
 </html>
--- a/readme.md
+++ b/readme.md
@@ -27,8 +27,8 @@ The page should match the design Pixel Perfect: all the sizes, colors and distan
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_moyo-header/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_moyo-header/report/html_report/)
+- [DEMO LINK](https://GuilhermeSant.github.io/layout_moyo-header/)
+- [TEST REPORT LINK](https://GuilhermeSant.github.io/layout_moyo-header/report/html_report/)
 
 ❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.
 
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,60 @@
+html {
+  font-family: Roboto, sans-serif;
+}
+
+:root {
+  --link: #00acdc;
+}
+
 body {
   margin: 0;
+  background-color: rgb(216, 216, 216);
+}
+
+.container {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background-color: white;
+  width: 100%;
+}
+
+.container img {
+  margin: 10px 419px 0 20px;
+}
+
+.nav-bar {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+}
+
+.nav-link {
+  position: relative;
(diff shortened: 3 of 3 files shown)


pedro-ruas (changes requested):
Guilherme, your code has failed all tests, and it visually differs from the wanted result. Here are some points that may help you achieve the wanted result:

- The class `is-active` is not meant to be used in every link, only in the first.
- Hovering effects for that class are actually for the `nav__link` class
- `is-active` should make the color blue at all times
- Your BEM naming is wrong, if the Block is called `nav-bar`, then the elements should be `nav-bar__link`, not `nav-link`
- Double check The height for the header in the design
- Check the `bottom` attribute for `is-active::after`, the position is wrong. Also some other stylings like the height is also wrong, check the designs for it

src/index.html
```diff
@@ -13,10 +13,40 @@
...
+        <img
+          class="header__logo"
+          src="./images/logo.png"
+          alt="logo-moyo"
+        />
+      </a>
+      <nav class="header__nav">
+        <div class="header__nav__link header__nav__link--is-active">Apple</div>
+        <div class="header__nav__link">Samsung</div>
+        <div class="header__nav__link">smartphones</div>
```
pedro-ruas, line 38: Same idea is valid for all other links, ok? `class="header__nav-link"`

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  width: 40px;
+}
+
+.header__nav {
+  display: flex;
+  flex-direction: row;
+  padding-right: 50px;
+}
+
+.header__nav__link {
```
pedro-ruas, line 25: You will have to rename this as explained in the previous comments

src/index.html
```diff
@@ -13,10 +13,40 @@
...
+        target="_blank"
+      >
+        <img
+          class="header__logo"
+          src="./images/logo.png"
+          alt="logo-moyo"
+        />
+      </a>
+      <nav class="header__nav">
+        <div class="header__nav__link header__nav__link--is-active">Apple</div>
```
pedro-ruas, line 36: All `header__nav-link` should use the tag `<a>` instead of `<div>`

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  color: #000;
+  text-transform: uppercase;
+  font-size: 12px;
+  font-weight: 500;
```
pedro-ruas, line 29: I believe this should be higher. By the way the design and tests look, I would say it's either 600 or 700, not 500

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  line-height: 14.06px;
+  height: 15px;
```
pedro-ruas, line 31: height should be 100% of the parent, not fixed to 15px

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  margin-right: 20px;
+  padding: 22.5px 0;
```
pedro-ruas, line 36: Remove this padding, use `display: flex` with `align-items: center`

Usually, "half" or "broken numbers" in design are actually not fixed, but came from the centering or another type of alignment

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+}
+
+.header__nav__link--is-active {
```
pedro-ruas, line 39: Same styling should be added to the hover state of links:
```
.is-active, .header__nav-link:hover {
  ...
 }
```

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  content: '';
+  display: block;
+  height: 4px;
+  background-color: var(--link);
+  width: 100%;
+  margin-top: 18.5px;
+  border-radius: 8px;
+}
+
+.header__nav .header__nav__link:nth-child(4) {
```
pedro-ruas, line 53: Remove this styling. The 4th link is a reference for how a link should look when hovered, not that it should be fixed in this styling.

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  color: var(--link);
+}
+
+.header__nav__link--is-active::after {
+  content: '';
+  display: block;
+  height: 4px;
+  background-color: var(--link);
+  width: 100%;
+  margin-top: 18.5px;
```
pedro-ruas, line 49: Again, half values are unusual.
To implement this after correctly, what you are looking for is to use `position: absolute` and `bottom: 0` or something similar.
And for this to work, `header__nav-link` should have `position: relative`

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  margin: 10px;
+  padding-right: auto;
+  height: 40px;
+  width: 40px;
+}
+
+.header__nav {
+  display: flex;
+  flex-direction: row;
+  padding-right: 50px;
```
pedro-ruas, line 22: Why this right padding? The padding is being controlled by `.header`

src/styles/blocks/header.css
```diff
@@ -0,0 +1,55 @@
...
+  justify-content: space-between;
+  background-color: white;
+  width: 100%;
+  height: 60px;
+  padding: 0 50px;
+}
+
+.header__logo {
+  margin: 10px;
+  padding-right: auto;
```
pedro-ruas, line 14: Why this padding here?

pedro-ruas (changes requested):
It is looking way closer than before, but there is still much changes to do

PR #5461, opened 2024-10-03
implemented task

JK-devv:
- [DEMO LINK](https://JK-devv.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://JK-devv.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -16,7 +16,45 @@
       href="./style.css"
     />
   </head>
+  <link
+    rel="preconnect"
+    href="https://fonts.googleapis.com"
+  />
+  <link
+    rel="preconnect"
+    href="https://fonts.gstatic.com"
+  />
+  <link
+    href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+    rel="stylesheet"
+  />
   <body>
-    <h1>Moyo header</h1>
+    <header>
+      <a class="image-style">
+        <img
+          src="/src/images/logo.png"
+          alt="Moyo logo"
+        />
+      </a>
+      <nav>
+        <ul>
+          <li><a class="menu-item is-active">APPLE</a></li>
+          <li><a class="menu-item">SAMSUNG</a></li>
+          <li><a class="menu-item">SMARTPHONES</a></li>
+          <li>
+            <a
+              class="menu-item"
+              data-qa="hover"
+            >
+              LAPTOPS & COMPUTERS
+            </a>
+          </li>
+          <li><a class="menu-item">GADGETS</a></li>
+          <li><a class="menu-item">TABLETS</a></li>
+          <li><a class="menu-item">PHOTO</a></li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -16,7 +16,45 @@
...
+  <link
+    rel="preconnect"
+    href="https://fonts.googleapis.com"
+  />
+  <link
+    rel="preconnect"
+    href="https://fonts.gstatic.com"
+  />
+  <link
+    href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
```
BudnikOleksii, line 28: Import only font weights that you actually use

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+  align-items: center;
+  font-family: Roboto, sans-serif;
+  padding-left: 50px;
+  padding-right: 50px;
+  margin: 0;
+}
+
+ul {
+  margin: 0;
+  padding: 0;
```
BudnikOleksii, line 19: Don't use tag names for styling (except html and body)
image

src/style.css
```diff
@@ -1,3 +1,71 @@
...
+
+a.is-active::after {
+  content: '';
+  position: absolute;
+  height: 4px;
+  width: 37px;
+  bottom: 0;
+  left: 50%;
+  transform: translateX(-50%);
+  background-color: #00acdc;
```
BudnikOleksii, line 55: CSS Variable is used for a blue color
image

BudnikOleksii (changes requested):
Good work 👍 
Don't forget about the checklist

src/style.css
BudnikOleksii: > Don't use tag names for styling (except html and body) image

Not fixed

BudnikOleksii (changes requested):
Good progress, but you still use tags instead of classes for styles. Fix in all places

src/style.css
```diff
@@ -1,3 +1,76 @@
...
+}
+
+li {
+  display: flex;
+  height: 60px;
+  position: relative;
+  margin-right: 20px;
+}
+
+li:last-child {
```
BudnikOleksii, line 36: You still style tags, it's a bad practice and an idea. The same thing with the links below.
It looks ok when you have such a small app like this, but it's not. For tags you can only reset or normalize styles so they will be the same in all browsers. But when you set specific styles to them, you lose flexibility and have problems with unexpected behavior when you add another list that should have different styles.

BudnikOleksii (changes requested):
Please, check the comments and the [checklist](https://github.com/JK-devv/layout_moyo-header/blob/develop/checklist.md). If you have any questions regarding why it's a bad idea to do like this don't hesitate to ask

src/style.css
JK-devv: Hi
Im sorry for taking your time
There was problem with my commits, as you can see in index html was styles that not exist in styles.css file 
I missed `styles.scc` in my commits

BudnikOleksii (approved):
Good job 👍

PR #5582, opened 2024-10-21
add solution, but need to fix 2 tests...

RHerheman:
- [DEMO LINK](https://RHerheman.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://RHerheman.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,68 @@
+html {
+  font-family: Roboto, Arial, sans-serif;
+  font-size: 12px;
+  text-transform: uppercase;
+}
+
+:root {
+  --blue: #00acdc;
+}
+
 body {
   margin: 0;
+}
+
+.header {
+  width: 100%;
+  display: flex;
+  justify-content: space-between;
+}
+
+.logo {
+  position: absolute;
+  margin: 0 auto;
+  top: 10px;
+  left: 50px;
+}
+
+.nav {
+  padding-right: 50px;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,68 @@
...
+.logo {
+  position: absolute;
+  margin: 0 auto;
+  top: 10px;
+  left: 50px;
+}
+
+.nav {
+  padding-right: 50px;
+}
```
etojeDenys, line 30:
```suggestion

```
we can remove it

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  display: flex;
+  justify-content: space-between;
+}
+
+.logo {
+  position: absolute;
+  margin: 0 auto;
+  top: 10px;
+  left: 50px;
+}
```
etojeDenys, line 26:
```suggestion
  margin-top: 10px;
}
```

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+
 body {
   margin: 0;
 }
+
+.header {
+  width: 100%;
+  display: flex;
+  justify-content: space-between;
+}
```
etojeDenys, line 19: and just add here padding: 0 50px

etojeDenys (changes requested):
remove the default padding from the ul
image

PR #5589, opened 2024-10-24
moyo  header

jhonatanstefani:
[DEMO LINK](https://jhonatanstefani.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://jhonatanstefani.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -19,7 +19,7 @@
 
   </head>
   <body>
-    <h1>Moyo header</h1>
+
 
     <header class="header">
      <a href="#imagem">
@@ -31,11 +31,11 @@
           <li><a href="#item1" class="is-active">Iphone</a></li>
           <li><a href="#item2">samsung</a></li>
           <li><a href="#item3">smartphone</a></li>
-          <li><a href="#item4">laptops & computers</a></li>
+          <li><a href="#item4"  data-qa="hover" class="laptops">laptops & computers</a></li>
           <li><a href="#item5">gadgets</a></li>
           <li><a href="#item6">tablets</a></li>
           <li><a href="#item7">photo</a></li>
-          <li><a href="#item8">video</a></li>
+          <li><a href="#item8" class="video">video</a></li>
         </ul>
       </nav>
     </header>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -15,8 +15,56 @@
...
+            </a>
+          </li>
+
+          <li><a href="#item2">samsung</a></li>
+          <li><a href="#item3">smartphone</a></li>
+          <li>
+            <a
+              href="#item4"
+              data-qa="hover"
+              class="laptops"
```
pedro-ruas, line 50: Class `laptops` ? This shouldn't be here.
All links should have a single class called `nav__link`

src/index.html
```diff
@@ -15,8 +15,56 @@
...
+          class="image"
+        />
+      </a>
+
+      <nav class="nav">
+        <ul>
+          <li>
+            <a
+              href="#item1"
+              class="is-active"
```
pedro-ruas, line 38: This should be `nav__link is-active`

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+}
+
+a {
+  margin: 10px;
+  text-decoration: none;
+  color: black;
+  font-weight: bold;
+}
+
+ul {
```
pedro-ruas, line 48: Instead of styling all `ul`, you should add a class to the `ul` you want to style, and add the styling to that class.
The same is valid for all tags

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  display: flex;
+  margin-left: 333px;
+  margin-top: 23px;
+}
+
+li {
+  justify-content: left;
+}
+
+nav a:hover {
```
pedro-ruas, line 59: Should use the class as mentioned for the `ul`

pedro-ruas (changes requested):
Jhonatan, there are some issues in your code:
- You did not use BEM here,
- You added most styling using the tag as reference instead of using a class,
- You also forgot to run `npm run deploy`, so DEMO link and TEST REPORT link are not working

PR #5611, opened 2024-10-28
Develop

pat5513:
- [DEMO LINK](https://pat5513.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://pat5513.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,11 +12,46 @@
     />
     <title>Moyo header</title>
     <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    />
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+      rel="stylesheet"
+    />
+    <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
+
   <body>
     <h1>Moyo header</h1>
+    <header class="header">
+      <div class="logo_container">
+        <img
+          src="./images/logo.png"
+          alt="logo"
+          class="logo"
+        />
+      </div>
+      <nav class="nav">
+        <ul>
+          <li><a href="">Apple</a></li>
+          <li><a href="">Samsung</a></li>
+          <li><a href="">Smartphones</a></li>
+          <li><a href="">Laptops & computers</a></li>
+          <li><a href="">Gadgets</a></li>
+          <li><a href="">Tablets</a></li>
+          <li><a href="">Photo</a></li>
+          <li><a href="">Video</a></li>
+        </ul>
+      </nav>
+    </header>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -32,35 +32,90 @@
   <body>
     <h1>Moyo header</h1>
     <header class="header">
-      <img
-        src="./images/logo.png"
-        alt="logo"
-        class="logo"
-      />
-      <nav class="nav">
-        <ul class="nav__links">
-          <li>
+      <a
+        href="index.html"
+        class="header__logo"
+      >
+        <img
+          src="images/logo.png"
+          alt="moyo logo"
+        />
+      </a>
+
+      <nav class="navigation">
+        <ul class="nav_list">
+          <li class="nav_item">
             <a
-              class="is-active"
-              href="#"
+              href="index.html"
+              class="is-active nav_link"
             >
-              Apple
+              apple
             </a>
           </li>
-          <li><a href="#">Samsung</a></li>
-          <li><a href="#">Smartphones</a></li>
-          <li>
+
+          <li class="nav_item">
             <a
-              href="#"
+              href="index.html"
+              class="nav_link"
+            >
+              samsung
+            </a>
+          </li>
+
+          <li class="nav_item">
+            <a
+              href="index.html"
+              class="nav_link"
+            >
+              smartphones
+            </a>
+          </li>
+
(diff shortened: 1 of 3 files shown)


natalia-klonowska (changes requested):
test's are failing because of this extra spaces between elements 
image
to fix it don't use `display: inline-block` for li (also [don't use tag names for styling ](https://github.com/mate-academy/layout_moyo-header/blob/master/checklist.md#css)) 
instead use `list-style: none;` and `display: flex;` (for nav_links)

natalia-klonowska:
image
these tests are failing because you didn't follow this instruction:
image

src/index.html
```diff
@@ -11,12 +11,113 @@
...
+      rel="stylesheet"
+    />
     <link
       rel="stylesheet"
       href="./style.css"
     />
   </head>
+
   <body>
     <h1>Moyo header</h1>
```
natalia-klonowska, line 33: you don't need this title ;)

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+}
+
+.header {
+  display: flex;
+  height: 60px;
```
natalia-klonowska, line 14: [Requirements:](https://github.com/mate-academy/layout_moyo-header/blob/master/readme.md#requirements) "the height should be set for nav links (**not the header**), take it from the design"

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  justify-content: space-between;
+  padding: 0 50px;
+}
+
+.header__logo {
+  align-self: center;
+  display: flex;
+  margin-left: 0;
```
natalia-klonowska, line 22: don't need to reset margin-left

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  align-items: center;
+}
+
+.nav_list {
+  display: flex;
+  list-style: none;
+  justify-content: space-between;
+  height: 100%;
+  align-items: center;
+  padding-inline-start: 0;
```
natalia-klonowska, line 36: you don't need all these styles but you should reset default margins for ul 
image

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+}
+
+.nav_item {
+  display: flex;
+  align-items: center;
+  height: 100%;
```
natalia-klonowska, line 42: don't need align-items and it will be enough to add height only for link

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  margin-left: 0;
+  position: relative;
+}
+
+.is-active:visited {
+  color: var(--link-on);
+}
+
+.is-active::after,
+.nav_link:active::after {
```
natalia-klonowska, line 86: pseudo-element should be added only for link with is-active class so you can delete selector `.nav_link:active::after`

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  color: var(--link-on);
+  text-decoration: none;
+  font-size: 12px;
+  text-transform: uppercase;
+  align-items: center;
+  display: flex;
+  height: 100%;
+  margin-left: 0;
+  position: relative;
+}
```
natalia-klonowska, line 79: you don't need all of these styles. remember that element has also nav_link class so you don't need to repeat styles that are added there

src/style.css
```diff
@@ -1,3 +1,95 @@
...
+  padding-inline-start: 0;
+}
+
+.nav_item {
+  display: flex;
+  align-items: center;
+  height: 100%;
+}
+
+.nav_link:link {
```
natalia-klonowska, line 45: you can just style nav__link without :link pseudo-class. also you don't need to style every state just hover ;)

natalia-klonowska (changes requested):
Almost done 🚀

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+}
+
+.nav_link:active {
+  color: var(--link-on);
+  position: relative;
+}
+
+.is-active {
+  color: var(--link-on);
+  margin-left: 0;
```
Zibi95, line 64: reset of margin with is-active is bad. You should use `:first-child` pseudo class fro this

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+  color: var(--link-on);
+}
+
+.nav_link:active {
+  color: var(--link-on);
+  position: relative;
+}
+
+.is-active {
+  color: var(--link-on);
```
Zibi95, line 63: This color is not applied to the active link because `.nav_link:visited` has higher specificity. You have to increase specificity of this selector to apply that active color.

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+  align-self: center;
+  display: flex;
+}
+
+.navigation {
+  display: flex;
+  align-items: center;
+}
+
+.nav_list {
```
Zibi95, line 28: ul tag is also applying some left padding. Thats why test's are failing.

src/style.css
```diff
@@ -1,3 +1,77 @@
...
+  margin-left: 0;
+  position: relative;
+}
+
+.is-active::after {
+  content: '';
+  display: block;
+  border-bottom: 4px solid var(--link-on);
+  position: absolute;
+  border-radius: 2px;
```
Zibi95, line 73: I think that border radius on that underline should be higher. In figma i believe its 8px

Zibi95 (changes requested):
Good job! You’re almost done. There are just a few small things to fix.

src/style.css
Zibi95: It should be
```
.nav_link.is_active {
}
```

To be more precise

Zibi95 (changes requested):
One small thing and we are good to go.

Zibi95 (approved):

PR #5674, opened 2024-11-12
solution of the challenge Moyo Header

MannyShussan:


MannyShussan:
Resolution of a challenge

PR #5787, opened 2024-12-07
add task solution

Yevhen2802:
- [DEMO LINK](https://Yevhen2802.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Yevhen2802.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,80 @@
-body {
+* {
   margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  list-style: none;
+}
+
+a {
+  text-decoration: none;
+  color: inherit;
+}
+
+html {
+  font-size: 16px;
+  font-family: Roboto, sans-serif;
+  font-style: normal;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,80 @@
...
+  display: flex;
+  align-items: center;
+  height: 60px;
+  margin: auto;
+  text-transform: uppercase;
+  font-size: 12px;
+  line-height: 16px;
+  font-weight: 500;
+}
+
```
etojeDenys, line 63: try to avoid styling the tags like that, better to add some class for li (nav__item) and `a` (nav__link) and style it using these classes

etojeDenys (changes requested):

src/style.css
Yevhen2802: I fixed it

anastasiiavorobiova (approved):

PR #6011, opened 2025-01-26
add task solution

ihorkrasnozhon:
- [DEMO LINK](https://ihorkrasnozhon.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://ihorkrasnozhon.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,68 @@
 body {
   margin: 0;
+  padding: 0;
+  background-color: antiquewhite;
+  font-family: Roboto, sans-serif;
+}
+
+header {
+  background-color: white;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 30px;
+}
+
+.logo {
+  display: flex;
+  justify-content: flex-start;
+  top: 10px;
+  left: 50px;
+  height: 40px;
+  width: 40px;
+}
+
+nav {
+  display: flex;
+  flex-direction: row;
+}
+
+a {
+  margin: 0 10px;
+  font-size: 12px;
+  font-weight: 500;
+  text-decoration: none;
+  color: black;
+  height: 60px;
+  text-transform: uppercase;
+
+  --active-color: #00acdc;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 30px;
+}
+
+.logo {
+  display: flex;
+  justify-content: flex-start;
+  top: 10px;
+  left: 50px;
```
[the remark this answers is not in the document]
Zibi95: You don't need those properties

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  height: 40px;
+  width: 40px;
+}
+
+nav {
+  display: flex;
+  flex-direction: row;
```
Zibi95, line 27: It's by default row direction

src/style.css
```diff
@@ -1,3 +1,68 @@
 body {
   margin: 0;
+  padding: 0;
+  background-color: antiquewhite;
```
Zibi95, line 4: This color is not needed

src/style.css
```diff
@@ -1,3 +1,68 @@
...
+  font-family: Roboto, sans-serif;
+}
+
+header {
+  background-color: white;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 30px;
```
Zibi95, line 13: You are missing 50 px of padding on left and right

Zibi95 (changes requested):
Hey. You are missing many things. Tests aren't passed. Fix what AI has requested.
- Don't use tag selectors
- keep text consistent (GADGeTS ??) 
- for active color (#00acdc) you should use CSS variable
- `a:checked` is not valid selector for anchor tag
- You should use those tags (<header>, , <nav>, <ul>, <li> and <a>.). Currently you are missing list tags in code
- missing margin between element of the list

PR #6211, opened 2025-03-10
Develop

aket0r:
- [DEMO LINK](https://aket0r.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://aket0r.github.io/layout_moyo-header/report/html_report/)


aket0r:
I used the code solution from another user because I could not fully understand where to get the properties for the values ​​of width, height, etc. (there was not enough data in figma). After analyzing the code of the finished solution, I understood how to approach the task.

PR #6233, opened 2025-03-14
feat: add page header layout and styles

s1rserg:
- [DEMO LINK](https://s1rserg.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://s1rserg.github.io/layout_moyo-header/report/html_report/)

- [x] The Google Fonts Configuration follows requirements.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -27,17 +27,17 @@ The page should match the design Pixel Perfect: all the sizes, colors and distan
 
 ❗️ Replace `<your_account>` with your GitHub username and copy the links to the `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_moyo-header/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_moyo-header/report/html_report/)
+- [DEMO LINK](https://s1rserg.github.io/layout_moyo-header/)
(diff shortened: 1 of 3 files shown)


s1rserg:
@mateacademy-ai-mentor

In src/style.css, the requirement to use CSS for uppercase transformation is not met. Please use text-transform: uppercase in your CSS instead of directly typing in uppercase in the HTML.
It is met, check more properly

In src/index.html, the HTML formatting needs correction. Use 2 spaces for indentation, and ensure that attributes are correctly formatted, each on a new line if there are multiple attributes or long values. Remember about correct indentation between parent and child elements, and add empty lines between multiline sibling blocks of HTML.
Id doesn't need correction. I used formatter and linter so it's definitely good.

In src/style.css, the CSS variable for blue color is named --brand-color instead of --blue. Please rename it to --blue.
There is no specific requirement for this but ok. I will change it

In src/style.css, avoid using the body selector for styling.
It was here by default and it is needed for resetting browser default margins which is a requirement

In src/index.html, make sure that the data-qa="hover" attribute is properly formatted.
It is properly formatted as i said before

Make sure you have specified fallback fonts in font-family property.
I added a few.

Ensure that the Google Fonts configuration follows the requirements for Roboto font: ONLY roman style, medium (500) weight and normal width.
I am sure it is as required.

PR #6874, opened 2025-08-19
Test changes

maifee02:


Anton-Kuchmasov (changes requested):
Please add a soluion to the task and then - re-request the review

PR #7033, opened 2025-09-16
Add solution

Iryna8989:
- [x ] Header height is set in 1 place (for the links)
- [x ] Content is vertically centered (for any header height)
- [x ] CSS is used to show all letters in Uppercase (don't type them in HTML)
- [x ] Logo is an image wrapped with a link
- [x ] **CSS Variable** is used for a blue color
- [x ] Pseudo-element is used for a blue line below the active link
- [x ] Code follows all the [Code Style Rules ❗️](./checklist.md)
- [x ] The Google Fonts Configuration follows requirements.


src/index.html
```diff
@@ -4,19 +4,62 @@
...
+
+      <nav>
+        <ul>
+          <li>
+            <a
+              href="#"
+              class="is-active"
+            >
+              Apple
+            </a>
```
FaiHamid, line 45: Each element had to be assigned its own class using the BEM methodology.

src/style.css
```diff
@@ -1,3 +1,72 @@
...
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+}
+
+img {
+  display: block;
+}
+
+nav ul {
```
FaiHamid, line 31: The same thing here, styles should have been added to classes, not tags.

src/style.css
```diff
@@ -1,3 +1,72 @@
...
+  --nav-link-color: #000;
+  --nav-link-hover-color: #00acdc;
+  --nav-height: 60px;
+}
+
+body {
+  font-family: Roboto, sans-serif;
+  font-weight: 500;
+  font-style: normal;
+}
```
FaiHamid, line 17: Usually, font styles are added to the html tag, because that is where elements inherit their font from.

FaiHamid (changes requested):
Good job!
Let`s change s few details:

1.Add Demo LINK and TEST REPORT LINK to the pull request description.

PR #7234, opened 2025-11-14
first

kkolodii:


etojeDenys (changes requested):
please provide demo and test links, and then request the review again

PR #7387, opened 2025-12-26
add task solution

maksymivivan2000-dotcom:
I did everything as shown in the photo, but I don’t understand how to fix the errors from BackstopJS. I hope to get an answer about what’s wrong and how to fix it


maksymivivan2000-dotcom (commented):
How is it?

PR #7428, opened 2026-01-11
add solution

Igorls34:
- [DEMO LINK](https://igorls34.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://igorls34.github.io/layout_moyo-header/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,98 @@
+@import 'https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap';
+
+html,
 body {
   margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+/* make box-sizing inherit for all elements so padding doesn't add to 100% width */
+*,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
+
+:root {
+  --accent-color: #00acdc;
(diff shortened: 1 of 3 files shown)


pedro-ruas (changes requested):
Ok, your project is almost perfect. It passes the tests, but there is  huge problem:
The header overflows the page width.
It should be at maximum the full page width, the page should not scroll due to the header.
This will cause trouble in next projects

src/style.css
```diff
@@ -1,3 +1,98 @@
+@import 'https://fonts.googleapis.com/css2?family=Roboto:wght@500&display=swap';
```
raulriato, line 1: You're already importing the font in the `index.html`. You don't need to import it here

src/style.css
```diff
@@ -1,3 +1,98 @@
...
+
+html,
 body {
   margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
```
raulriato, line 9: These styles should not be written here. You don't need specific browser styles for this project.

raulriato (changes requested):
You still have some problemas in the code, and in the `DEMO LINK`  your header still overflows the page. Fix it and run `npm run deploy` so the `DEMO LINK` updates with the code.

DanielCamposDevX (approved):

PR #7568, opened 2026-02-26
add task solution

Ivan-Kyslenko:
- [DEMO LINK](https://Ivan-Kyslenko.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://Ivan-Kyslenko.github.io/layout_moyo-header/report/html_report/)


src/style.css
```diff
@@ -1,3 +1,83 @@
...
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 0 50px;
+}
+
+.header__logo img {
```
Anton-Kuchmasov, line 21: In general, there's no need to use `img` additional class with `header__logo` - you can just use one of them to add `display: block`

src/style.css
```diff
@@ -1,3 +1,83 @@
...
+  padding: 0;
+}
+
+.header__link:hover {
+  color: var(--blue);
+}
+
+.header__link[data-qa='hover']:hover {
+  color: var(--blue);
+}
```
Anton-Kuchmasov, line 67: what's purpose of adding extra selector here?

`.header__link:hover` will work in this case too

Anton-Kuchmasov (approved):
Good job!

Please read comments below

PR #7680, opened 2026-03-25
alteração layout_moyo-header

Ranierij:


Ranierij (commented):
feitas alterações

Ranierij (commented):
novas correções

Ranierij (commented):
.

IanLuan (changes requested):
DEMO LINK is not working. Please run npm run deploy e update the links on README.

readme.md
```diff
@@ -27,8 +28,8 @@ The page should match the design Pixel Perfect: all the sizes, colors and distan
 
 ❗️ Replace `<your_account>` with your GitHub username and copy the links to the `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_moyo-header/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_moyo-header/report/html_report/)
+- [DEMO LINK](https://github.com/Ranierij/layout_moyo-header)
```
raulriato: This is wrong. You should only replace `<your_account>` with your username instead of changing the structure of the URL.

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+
+.header {
+  width: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  font-weight: 500;
+  background-color: var(--white);
+  padding: 0 50px;
+  height: 60px;
```
raulriato: You should not set the height in the `header`. Read the requirements of the task, the `height` should be set in the `nav__link`

src/style.css
```diff
@@ -1,3 +1,82 @@
...
+}
+
+.logo {
+  display: flex;
+  align-items: center;
+  margin-right: 424px;
+}
+
+.nav-list {
```
raulriato: You're not following the `BEM` correctly. `list` is an `Element`  inside the `nav Block`, so it should be `nav__list`. The same goes to all your elements in the project, you should fix every element.

raulriato (changes requested):
Your code works but does not fill all the requirements. Make sure the tests pass after  fixing the issues and rerun the `npm run deploy` before re-requesting review.

pedro-ruas (changes requested):
Your request is missing the DEMO and TEST REPORT links.
I also see some tests have failed.
Please ad the links and run `npm run deploy`

PR #7715, opened 2026-04-05
Develop

GeorgeFront:
https://GeorgeFront.github.io/layout_moyo-header/

https://GeorgeFront.github.io/layout_moyo-header/report/html_report/


brespect (changes requested):
Well done, but you need to pass all tests before requesting the review, there is some problem with alignment I think

GeorgeFront:
> Well done, but you need to pass all tests before requesting the review, there is some problem with alignment I think

To be honest, no idea what is wrong. I asked about it in mentor chat yesterday and sent my code to the mentor. He tried it on his computer and everything was fine on his side. He said that it's kind of local issue, so should be fine. But when I test my program it still fails at 1 test, though I've made everything due to the layout.

PR #7753, opened 2026-04-18
Develop

VitinzX1:
- [DEMO LINK](https://VitinzX1.github.io/layout_moyo-header/)
- [TEST REPORT LINK](https://VitinzX1.github.io/layout_moyo-header/report/html_report/)


src/index.html
```diff
@@ -15,8 +16,105 @@
...
   </head>
+  <!-- #endregion -->
+
   <body>
-    <h1>Moyo header</h1>
+    <!-- #region HEAD - SITE -->
+    <header class="head_site">
+      <a
+        href="#"
+        class="Logo"
```
raulriato, line 39: It's not good practice to use `class` names starting with Uppercase.

src/index.html
```diff
@@ -15,8 +16,105 @@
...
+      >
+        <img
+          src="/src/images/logo.png"
+          alt="MOYO LOGO"
+          class="Logo__img"
+        />
+      </a>
+
+      <nav class="nav__head">
```
raulriato, line 48: You are not following `BEM` correctly. Since the `Elements` inside this `Block` are `nav__list` `nav__item`, etc. The name of this `Block` should be `nav`.

src/index.html
```diff
@@ -15,8 +16,105 @@
...
+      href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100..900;1,100..900&display=swap"
+      rel="stylesheet"
+    />
   </head>
+  <!-- #endregion -->
+
   <body>
-    <h1>Moyo header</h1>
+    <!-- #region HEAD - SITE -->
+    <header class="head_site">
```
raulriato, line 36: You should not use `underscore` in the names of your `class`. you can either go with `head-site` (but remember the `Elements` inside it will have to be `head-site__[ELEMENT-NAME]`) or simply `header`.

raulriato (changes requested):
Your code works, It's all good, but it's really important to follow some good practices in your code. Just fix these and you'll be good to go.

PR #7760, opened 2026-04-20
moyo header is done

WhiteAshDc:


WhiteAshDc (commented):
Чи є які помилки в завданні?

brespect (changes requested):
Almost done, but let's paste your DEMO LINK into PR description

PR #7846, opened 2026-05-22
Finish moyo header

obikhov:


obikhov:
> _ No description provided. _

- [x] Header stretches full page width
- [x] Logo is a link with image
- [x] Navigation uses semantic tags
- [x] Active link has class is-active
- [x] 4th link has data-qa="hover"
- [x] Hover color works
- [x] Active line uses ::after
- [x] No flex gap used
- [x] Nav links have height, no padding

PR #8048, opened 2026-07-20
add solution

toshka7313:
- [x] The Google Fonts Configuration follows requirements.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -10,8 +10,8 @@ html {
 }
 
 body {
-  margin: 0;
   padding: 0;
+  margin: 0;
   font-weight: 500;
   box-sizing: border-box;
 }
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1,3 +1,74 @@
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  line-height: 100%;
+  padding: 0;
+  margin: 0;
+  box-sizing: border-box;
+
+  --blue: #00acdc;
```
Moroz-Dmytro, line 9: Variable name should describe where this color will be used or some general name, It should not contain color name

src/style.css
```diff
@@ -1,3 +1,74 @@
...
+  text-transform: uppercase;
+}
+
+.nav__link:hover {
+  color: var(--blue);
+}
+
+.nav__link.is-active {
+  color: var(--blue);
+}
```
Moroz-Dmytro, line 65:
```suggestion
.nav__link:hover,
.is-active {
  color: var(--blue);
}
```

Moroz-Dmytro (approved):

